PR #13, opened 2026-03-03
fix: Replace Allauth Element Tags with Standard Form Rendering

HexagonStorms:
## Summary
- Replace broken allauth `{% element %}` tags with standard Django form rendering in login/signup templates
- Add `ALLAUTH_TRUSTED_PROXY_COUNT` env var for proxy IP resolution behind nginx

## Test plan
- [x] `pytest` — all tests pass
- [x] `ruff check .` / `ruff format --check .` — clean
- [x] `mypy` — clean

### DIFF
--- a/plfog/settings.py
+++ b/plfog/settings.py
@@ -143,7 +143,6 @@ LOGOUT_REDIRECT_URL = "/"
 SOCIALACCOUNT_ADAPTER = "plfog.adapters.AutoAdminSocialAccountAdapter"
 ACCOUNT_ADAPTER = "plfog.adapters.AdminRedirectAccountAdapter"
 SOCIALACCOUNT_LOGIN_ON_GET = True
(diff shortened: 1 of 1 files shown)


plfog/settings.py
```diff
@@ -147,6 +143,7 @@
 SOCIALACCOUNT_ADAPTER = "plfog.adapters.AutoAdminSocialAccountAdapter"
 ACCOUNT_ADAPTER = "plfog.adapters.AdminRedirectAccountAdapter"
 SOCIALACCOUNT_LOGIN_ON_GET = True
+ALLAUTH_TRUSTED_PROXY_COUNT = int(os.environ.get("ALLAUTH_TRUSTED_PROXY_COUNT", "0"))
```
markng: I don't think we need this, and it could accidentally introduce a security hole based on the .env. Was there a use case for it?

markng:
```suggestion
```

PR #57, opened 2026-03-28
fix: Email branding + auto-create User for Airtable members

HexagonStorms:
## Summary

- **Auto-create User on login**: Members synced from Airtable have Member records but no Django User accounts. When they enter their email on the login page, a User is now auto-created and linked so they receive a login code immediately instead of an "unknown account" email.
- **Brand allauth email templates**: Override `base_message.txt`, `unknown_account`, and `account_already_exists` templates to use Past Lives branding instead of "example.com"
- **Fix default Site domain**: `post_migrate` signal updates Django's default Site from `example.com` to `pastlives.plaza.codes`
- **Prevent empty-email Members from Airtable sync**: Skip creating Member records with no email (caused 91 ghost duplicates in production)
- **Production data cleaned**: Deleted 91 empty-email Members, nulled 3 orphaned guild_lead references

## Test plan

- [x] 721 tests pass, 100% coverage
- [x] Ruff + mypy clean
- [x] Production empty-email Members deleted
- [ ] Verify affected users (jenellegiordano@gmail.com, caitvonderwin@gmail.com) can now log in after deploy
- [ ] Verify branded emails render correctly (unknown account, account already exists)

### DIFF
--- a/airtable_sync/management/commands/airtable_pull.py
+++ b/airtable_sync/management/commands/airtable_pull.py
@@ -102,6 +102,11 @@ class Command(BaseCommand):
             results["updated"] += 1
             return
 
+        if not email:
+            logger.info("Skipping Airtable record %s (no email): %s", record_id, django_kwargs.get("full_legal_name"))
+            results["skipped"] += 1
+            return
+
         if not dry_run and default_plan:
             django_kwargs["airtable_record_id"] = record_id
             django_kwargs["membership_plan"] = default_plan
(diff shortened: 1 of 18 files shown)


PastLivesReviewBot (changes requested):
One BLOCK issue: missing tests for the new management command. Two WARNs noted inline. Fix the BLOCK and this is good to merge.

PastLivesReviewBot (approved):
BLOCK resolved — tests added for fix_unverified_emails command (3 specs: verify, zero-count, dry-run). 724 tests pass, 100% coverage. Prior WARNs still apply but don't block. Ship it.

PR #59, opened 2026-03-30
Admin member management redesign (v1.2.0)

HexagonStorms:
## Summary

- **MemberEmail model** — members can have multiple email aliases, all work for login
- **Unified admin Members page** — "Users" toggle filter shows who has logged in (with tooltip), "All Members" shows everyone
- **Create login immediately** — checkbox on the add form creates both User + Member in one step
- **Email aliases inline** — add/remove emails right from the member edit form
- **Login flow checks aliases** — logging in with any alias email links to the correct member
- **Search by alias** — admin search finds members by any of their email addresses
- **Hidden User/EmailAddress admin** — everything managed through Members now

## Test plan

- [ ] Admin Members list loads with Users/All toggle filter
- [ ] Tooltip on "Users" filter says "Users are members who have logged into this web app"
- [ ] Add member without "Create login" — no User created
- [ ] Add member with "Create login" checked — User + Member created and linked
- [ ] Edit member — email aliases inline visible, can add/remove
- [ ] Login with alias email — correct member linked
- [ ] Search by alias email in admin — member found
- [ ] User admin page no longer in sidebar
- [ ] 746 tests pass, 100% coverage


PastLivesReviewBot (changes requested):
One deploy-blocking issue: missing migration for the help_text added to MemberEmail.member FK. Two minor WARNs noted inline.

PastLivesReviewBot (approved):
All prior findings addressed. Migration 0018 covers the missing help_text, adapters test now uses a targeted mock instead of bare except, and TypedDicts clean up the int() casts. No BLOCK or WARN issues remain — LGTM.

PR #64, opened 2026-04-07
feat: direct API keys mode for StripeAccount

HexagonStorms:
## Summary

Adds a second auth mode to `StripeAccount` (`direct_keys`) alongside the existing Stripe Connect OAuth flow. Guilds can now connect their own Stripe account by pasting API keys directly in the admin dashboard — no registered Connect platform required. This unblocks consumable billing (clay, materials, etc.) where 100% of revenue belongs to the guild that owns the items.

- Direct-keys charges create hosted Stripe Checkout sessions on the guild's own account using their own secret key. No `transfer_data`, no `application_fee_amount`, no platform fee — money settles directly in the guild's balance.
- Secret key + webhook signing secret are encrypted at rest with a new `EncryptedCharField` (Fernet). Plaintext is never stored in the DB column.
- Existing OAuth Connect mode is unchanged. Both modes coexist on `StripeAccount.auth_mode` so we can migrate guilds to Connect later for membership/space-lease billing where PL needs a platform fee.
- New per-guild webhook endpoint at `/billing/webhooks/stripe/guild/<guild_id>/` verifies signatures with each guild's own `whsec_…`.
- Admin payments dashboard now has a two-tab "Connect Stripe Account" modal — "Paste API Keys" (default) and "Connect via OAuth" — with a "Test connection" button that hits `accounts.retrieve('self')` against the pasted key before saving.

Plan: \`/home/josh/.claude/plans/snoopy-churning-hanrahan.md\`

## Test plan

- [x] 311 billing tests pass (\`pytest tests/billing/\`)
- [x] Full suite passes (1082 tests)
- [x] \`ruff check\` + \`ruff format --check\` clean
- [x] Migration applies cleanly
- [ ] Generate \`STRIPE_FIELD_ENCRYPTION_KEY\` and set on local + Hetzner + Render
- [ ] End-to-end: paste a real test-mode \`sk_test_\` from a sandbox Stripe account, verify "Test connection" returns the right \`acct_…\`, save, then run \`bill_tabs\` and confirm a Checkout session URL gets created on the guild's account
- [ ] Wire up webhook in the guild's Stripe dashboard pointing at \`/billing/webhooks/stripe/guild/<guild_id>/\`, paste the \`whsec_…\` back into the modal, trigger \`checkout.session.completed\` via Stripe CLI and confirm the TabCharge flips to SUCCEEDED
- [ ] Regression: existing OAuth Connect path still works with a valid \`STRIPE_CONNECT_CLIENT_ID\`

## Notes

- **Losing \`STRIPE_FIELD_ENCRYPTION_KEY\` bricks all stored direct-mode credentials.** Document and back up.
- Saved cards / payment methods are out of scope for direct-keys MVP — Checkout sessions only. A per-guild \`GuildPaymentMethod\` model is the planned follow-up if we want stored cards in this mode.
- Refunds/disputes for direct-keys charges happen in each guild's own Stripe dashboard, not the unified admin.


PastLivesReviewBot (approved):
Looks solid. Architecture is clean (oauth/direct_keys mode coexists cleanly), encryption is correct, webhook routing is per-guild, tests cover the new branches at 100%. A few non-blocking suggestions left as inline comments for follow-up.

PR #67, opened 2026-04-11
1.4.1: hide contributor counts + admin email aliases

HexagonStorms:
## Summary

1.4.1 release. Bundles two pieces of work:

1. **Privacy fix:** the funding results section no longer shows how many members contributed to each snapshot — that detail is admin-only now.
2. **Admin email aliases:** dedicated admin page at `/admin/members/<pk>/aliases/` for full CRUD on member email aliases (add, remove, set-primary, toggle-verified). Entry point is a "Manage email aliases →" link on the member change page.
3. **Bug fix found along the way:** the post_save signal on User was silently reverting any primary email change made through allauth's `set_as_primary()` (called by both self-service and the new admin endpoints). Fixed by gating the signal on `created=True`.

## Why the admin email aliases feature exists

A real use case surfaced after 1.4.0 shipped: adding `writersguild@pastlives.space` to a member's account so they can sign in as that address. The shared mailbox case breaks the self-service flow at `/accounts/email/` because the member can't easily receive a verification code from a guild-owned inbox. This PR gives admins a one-click path to add such aliases (auto-verified on creation).

Members will accumulate shared/aliased email addresses over time — guild mailboxes get reassigned, members get role-based addresses. Full CRUD makes sense because admins also need to clean up: remove a stale alias, promote a new one to primary.

## The signal bug we fixed

While building the remove + set-primary endpoints, tests kept failing because primary changes were being silently reverted. Root cause in `membership/signals.py`: the `post_save` signal on `User` was calling `MemberEmail.objects.migrate_to_user(user)` on **every** save, which force-re-promoted `Member._pre_signup_email` to primary. allauth's `set_as_primary()` calls `user.save()` internally, which fired the signal, which reverted whatever primary had just been set.

**This means 1.4.0's own self-service primary-change at `/accounts/email/` was silently broken** for anyone who tried to set a non-signup-email as primary. Nobody noticed because most members keep their original signup email.

The fix gates the signal on `created=True`. Every branch of the signal is only meaningful on first user save. Two regression specs prove primary changes now persist across `user.save()` and through the full `set_as_primary()` round-trip.

## Safety rules implemented (admin email aliases)

1. **Cannot remove the only email** — refuses with error flash (would lock the user out)
2. **Last verified email** — removal proceeds but flashes a loud warning
3. **Removing the primary** — auto-promotes the lowest-pk verified row to primary
4. **Cannot set primary on unverified** — refuses with error flash
5. **Duplicate on self / on other user** — caught by form validation with friendly errors
6. **Cross-user URL crafting** — `get_object_or_404(EmailAddress, pk=email_pk, user=member.user)` returns 404 if you try to operate on another member's email by hand-crafted URL
7. **Unlinked members** — GET redirects to the member change page with a hint pointing at the staged emails inline

## Test coverage

- **45 new specs** in `tests/plfog/member_aliases_spec.py` — full CRUD matrix + form validation + template rendering + admin link + end-to-end login flow
- **2 new regression specs** in `tests/membership/user_link_email_migration_spec.py` for the signal fix
- **3 new regression specs** in `tests/hub/guild_voting_spec.py` for the contributor-count privacy hotfix
- **790 total tests pass** (full project), no regressions

## Spec & plan documents

- `docs/superpowers/specs/2026-04-11-admin-email-aliases-design.md`
- `docs/superpowers/plans/2026-04-11-admin-email-aliases.md`

## Test plan

- [x] `pytest tests/plfog/member_aliases_spec.py` — 45 passed
- [x] `pytest tests/hub/guild_voting_spec.py` — privacy regression specs pass
- [x] `pytest tests/membership/user_link_email_migration_spec.py` — signal fix regression specs pass
- [x] Full project pytest — 790 passed
- [x] `ruff check .` / `ruff format --check .` — clean
- [x] `mypy plfog/ core/ membership/ hub/` — clean
- [x] `manage.py check` — clean
- [x] End-to-end: admin adds `writersguild@pastlives.space` to a linked member → fresh client requests login code at that address → submits code → session authenticated as the member's user
- [ ] Manual smoke on staging once merged

### DIFF
--- a/docs/superpowers/plans/2026-04-11-admin-email-aliases.md
+++ b/docs/superpowers/plans/2026-04-11-admin-email-aliases.md
@@ -1,0 +1,1835 @@
+# Admin-Managed Email Aliases Implementation Plan
+
+> **For agentic workers:** REQUIRED SUB-SKILL: Use superpowers:subagent-driven-development (recommended) or superpowers:executing-plans to implement this plan task-by-task. Steps use checkbox (`- [ ]`) syntax for tracking.
+
+**Goal:** Build a dedicated admin page at `/admin/members/<pk>/aliases/` that lets staff add, remove, set-primary, and toggle-verified on member email aliases, with full safety rules and BDD test coverage.
+
+**Architecture:** Thin Django views in `plfog/admin_views.py` (mirroring the Snapshot Analyzer pattern), one form in `membership/forms.py`, one Unfold-styled template, and a `MemberAdmin` readonly link field as the entry point. No inline formsets, no `change_form.html` override.
+
+**Tech Stack:** Django 6, django-allauth (`EmailAddress` model + `set_as_primary()`), Unfold admin, pytest-describe + factory-boy, `@staff_member_required` + `@require_POST`.
+
+**Reference spec:** `docs/superpowers/specs/2026-04-11-admin-email-aliases-design.md` — read it first. Every rule in the "Safety rules" section of the spec must be honored.
+
+**PLFOG coding standards:** See `/Users/joshplaza/Code/hexagonstorms/plfog/CLAUDE.md`:
+- Fat models, skinny views
+- Validation lives in forms, never in views
+- Type annotations on every function (including `-> None`)
+- 100% branch coverage, no `@pytest.mark.skip` / `# pragma: no cover` / `# pragma: no mutate`
+- BDD specs in `*_spec.py` with `describe_*` / `context_*` / `it_*`
+- `dict[key]` not `dict.get(key, default)`
+- `ruff check .`, `ruff format .`, `mypy plfog/ core/ membership/ hub/` must all pass before every commit
+
+---
+
+## File Structure
+
+**Create:**
+- `templates/admin/membership/member/aliases.html` — the aliases page template
+- `tests/plfog/member_aliases_spec.py` — BDD specs
+
+**Modify:**
+- `membership/forms.py` — add `AddEmailAliasForm`
+- `plfog/admin_views.py` — add 5 view functions
+- `plfog/urls.py` — add 5 URL routes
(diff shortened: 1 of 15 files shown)


PastLivesReviewBot (approved):

**Scope covered:** three threads bundled into 1.4.1 — contributor-count privacy hotfix, admin email aliases CRUD, and a post_save signal bug fix that repaired 1.4.0's silently-broken self-service primary-change flow.

**Standards verified:**
- Fat models / skinny views — all 5 new admin views are thin, with correct `get_object_or_404(EmailAddress, pk=email_pk, user=member.user)` guards preventing cross-user IDOR on hand-crafted URLs
- Validation lives in `AddEmailAliasForm.clean_email` (case-insensitive self-duplicate + other-user duplicate), not in the view
- Type annotations on every new function including `-> None` on the signal
- CSRF token on every POST form in the aliases template; destructive Remove action wrapped in `onsubmit="return confirm(...)"` with `|escapejs` on the interpolated email
- No `|safe` / `mark_safe` usage that could leak unescaped user input
- 1220 tests pass, 100% branch coverage — no `@pytest.mark.skip`, `# pragma: no cover`, or `# pragma: no mutate`
- Signal fix gate on `created=True` is sound; two regression specs prove primary changes now persist across `user.save()` and through the full `set_as_primary()` round-trip
- Privacy hotfix has three positive-absence regression specs, one per member-facing template surface
- Version correctly bumped to 1.4.1 with member-friendly changelog

**Non-blocking observation (leaving for follow-up, not blocking merge):**
- `plfog/admin_views.py:362-364` — when `member_aliases_remove` removes the last verified email, both a `messages.warning` ("no verified emails left") and `messages.success` ("removed alias X") fire simultaneously. Both convey accurate information, but the simultaneous warning+success can be confusing. Consider `else: messages.success(...)` to suppress the success flash when the warning fires.

CI green: lint + test both pass on commit 97baaaf.

PR #73, opened 2026-04-15
1.5.4: product revenue splits + public guild editing + My Tab polish

HexagonStorms:
## Summary

- **Product revenue splits** — Every product now splits revenue across Admin + any combination of guilds (percentages sum to 100). Replaces the old single-recipient / admin-percent model. Existing products + past tab entries were wiped as part of this migration.
- **Public guild editing** — Guild admin pages are gone from the Django admin. Admins, guild officers, and a guild's lead now edit name/about and add/delete products from the public `/guilds/<id>/` page via modals. `Member.can_edit_guild()` gates both the UI and the views.
- **My Tab polish** — Hoverable `?` next to the page title explains how a tab differs from a cart. Next-charge timestamp moves to subtext under the heading instead of buried in the middle of the card.
- **Sidebar** — Guilds section expanded by default; preference persists in localStorage when collapsed.

## Test plan

- [ ] Log in as a regular member on `/guilds/<id>/` — no Edit / Add Product buttons visible
- [ ] Log in as admin — both Edit Guild Page (header) and Add Product (Products card) buttons show
- [ ] Set a Member as a guild's lead — they see the buttons on that guild only
- [ ] Add Product modal: open it, delete a row, hit Cancel, re-open — rows reset to default Admin 20% / Guild 80%
- [ ] Create a product with a 3-way split (e.g. Admin 20% / Guild A 40% / Guild B 40%) and add it to a tab — TabEntrySplit rows written
- [ ] Reports page (Payments → Reports) shows one row per recipient, not one row per entry
- [ ] My Tab page: `?` icon tooltip on hover/focus; next-charge line appears as subtext under the title
- [ ] Collapse the sidebar Guilds section, reload — it stays collapsed
- [ ] `pytest` — all 1237 tests pass
- [ ] `ruff check .`, `mypy plfog/ core/ membership/ hub/` clean


hub/views.py
```diff
@@ -220,10 +249,146 @@ def guild_detail(request: HttpRequest, pk: int) -> HttpResponse:
...
+    # as new; the actual link to ``product`` happens on save() below.
+    formset = build_product_split_formset(data=request.POST, instance=Product())
+
+    if form.is_valid() and formset.is_valid():
+        updated = form.save(commit=False)
+        updated.guild = guild  # always bind to the page's guild
+        updated.save()
+        updated.splits.all().delete()
+        formset.instance = updated
+        formset.save()
```
PastLivesReviewBot, line 366: **WARN — data-integrity race**

`splits.all().delete()` runs before `formset.save()` without `transaction.atomic()`. If the formset save fails mid-way (DB connection hiccup, constraint violation, etc.), the product is left with zero splits and no way to recover the previous ones. Form validation passes first so this is unlikely in practice, but wrapping the delete+save block in `with transaction.atomic():` would make it atomic and safer. Not blocking — follow-up.

billing/forms.py
```diff
@@ -124,12 +98,14 @@ def __init__(
...
-                queryset=Product.objects.filter(is_active=True).select_related("guild"),
+                queryset=Product.objects.select_related("guild"),
                 required=False,
                 empty_label="— Manual entry (no product) —",
                 label="Product",
             )
-        elif context == CONTEXT_MEMBER_GUILD_PAGE:
+        elif (
+            context == CONTEXT_MEMBER_GUILD_PAGE
+        ):  # pragma: no branch — VALID_CONTEXTS guard makes other branch unreachable
```
PastLivesReviewBot, line 108: **WARN — style**

The `# pragma: no branch` comment is accurate (the `VALID_CONTEXTS` guard at line 76 makes the implicit `else` unreachable), but an explicit `else: raise AssertionError(...)` would make the invariant visible in the code instead of relying on a pragma. Not blocking.

PastLivesReviewBot (approved):
Approved. Migrations, models, views, and tests all meet project standards. 100% coverage maintained. Two WARN-level notes inline — neither blocks this PR but worth addressing in a follow-up.

PR #74, opened 2026-04-15
1.5.6: Guild voting hotfixes, User Settings, sidebar polish, deploy fix

HexagonStorms:
## Summary

- **Guild voting point inflation fixed.** Live standings, snapshots, and the admin dashboard now only count `VotePreference`s from signed-up members (`member__user__isnull=False`) — the backfilled Airtable votes from members who never signed up were tripling the top guild's points.
- **Guild Voting page now has 3 tabs:** Current Standings · New Votes This Month · Last Month's Results.
- **Admin VotePreference change view** shows per-member voting history pulled from each snapshot's `raw_votes` — audit how a member's picks contributed to past totals.
- **New `/settings/` tabbed User Settings page** (Profile + Emails) replaces `/settings/profile/`, `/settings/emails/`, and the off-theme `/accounts/email/` page. Old URLs 302 into the new page; profile dropdown collapsed to **Settings + Log Out**.
- **Emails tab** splits Manage Email Addresses and Email Preferences into two separate `hub-card`s. "Re-send Verification" only shows when the selected email is unverified (Alpine-driven, seeded from the primary).
- **Sidebar footer:** Member Guide, Code of Conduct, and Feedback now live in the sidebar bottom using existing `.hub-sidebar__link` styling — no more floating FAB overlapping content.
- **Render deploy fix:** `billing.0010` now runs with `atomic = False` so the data wipe commits before the `ALTER TABLE`, resolving the Postgres "cannot ALTER TABLE … because it has pending trigger events" that bricked the last deploy.

## Test plan

- [x] `pytest` — 1308 passing, 100% coverage target held
- [x] `ruff check` / `ruff format` — clean
- [x] `mypy .` — clean
- [ ] `/guilds/voting/` — confirm Current / New / Last Month tabs render and point totals are in the 21-point ballpark again (not 60+)
- [ ] `/settings/` — Profile and Emails tabs both save; redirects from `/settings/profile/`, `/settings/emails/`, `/accounts/email/` land on the right tab
- [ ] Emails tab — "Re-send Verification" hidden for a verified primary, shows after selecting an unverified address
- [ ] Sidebar — Resources + Feedback render with proper padding, each on its own line, Feedback no longer floats
- [ ] Admin → Voting Preferences → click a row — Historical Votes table lists every snapshot containing that member
- [ ] After merge, verify Render deploy runs `billing.0010` cleanly


templates/hub/user_settings.html
```diff
@@ -0,0 +1,136 @@
+{% extends "hub/base.html" %}
+{% block title %}User Settings{% endblock %}
+
+{% block content %}
+<h1 class="hub-page-title">User Settings</h1>
+
+<div x-data="{ tab: '{{ active_tab|default:"profile" }}' }">
```
PastLivesReviewBot, line 7: **BLOCK — reflected XSS.** `active_tab` is set in the view from `request.GET.get("tab", "profile")` with no whitelist, then interpolated into Alpine's `x-data` JS expression here. Django's default HTML autoescape converts `'` to `&#x27;`, but browsers decode entities when reading attribute values, so Alpine evaluates the decoded JS. Proof of concept: `/settings/?tab=%27%2Balert(1)%2B%27` → attribute becomes `{ tab: ''+alert(1)+'' }` → `alert(1)` fires.

Fix in `hub/views.py` by whitelisting:
```python
tab_param = request.GET.get("tab", "profile")
active_tab = tab_param if tab_param in {"profile", "emails"} else "profile"
```
Or apply `|escapejs` here. Line 57's `primary_verified_json` is safe because the view hard-codes it to `"true"`/`"false"`.

plfog/urls.py
```diff
@@ -48,8 +51,25 @@
...
+
+class HubEmailView(EmailView):
+    """Override allauth's email management view to redirect into the hub's User Settings page.
+
+    POSTs (add, make primary, re-send, remove) still run through allauth's
+    EmailView logic; only the success_url and GET rendering change so the user
+    always lands on /settings/?tab=emails instead of the legacy themed page.
+    """
+
+    success_url = reverse_lazy("hub_user_settings")
```
PastLivesReviewBot: **WARN — UX.** `success_url = reverse_lazy("hub_user_settings")` lands users on the Profile tab after they POST an email action (add / make primary / remove) from the Emails tab. Since every allauth email action originates from the Emails tab, consider `success_url = "/settings/?tab=emails"` (plain string works since `reverse_lazy` is only needed when the pattern may not be resolvable at import time). Not a correctness bug — just jarring tab-hop.

membership/admin.py
```diff
@@ -274,9 +274,104 @@ def queryset(self, request: HttpRequest, queryset: QuerySet[VotePreference]) ->
...
+            "<td style='padding:6px 10px'>{}</td>"
+            "</tr>",
+            rows,
+        )
+        return mark_safe(  # noqa: S308
+            f"<table style='border-collapse:collapse;width:100%'><thead>{header}</thead><tbody>{body}</tbody></table>"
+        )
+
+
+def _count_member_snapshots(member_id: int) -> int:
```
PastLivesReviewBot: **WARN — N+1 on admin list view.** `snapshots_participated` calls this helper once per VotePreference row on the changelist, and each call does `FundingSnapshot.objects.only("raw_votes").iterator()` — one full table scan per row. At 200 members × 12 monthly snapshots that's 200 scans per page load, each reading every snapshot's JSON. Consider pre-computing once per request (e.g., `get_queryset` builds a `{member_id: count}` dict from a single snapshot iteration, then `snapshots_participated` reads from it) or dropping the column from `list_display` and keeping the history only on the change view.

PastLivesReviewBot (changes requested):
Found one reflected-XSS BLOCK plus two WARN items. Inline comments below.

tests/membership/admin_spec.py
```diff
@@ -305,6 +311,94 @@ def it_filters_non_paying_only(admin_client):
...
+        for i in range(10):
+            FundingSnapshot.objects.create(
+                cycle_label=f"Cycle {i}",
+                contributor_count=5,
+                funding_pool=Decimal("1000.00"),
+                raw_votes=[{"member_id": 1, "guild_1st_name": "x"}],
+                results={},
+            )
+
+        resp = admin_client.get("/admin/membership/votepreference/")
```
PastLivesReviewBot, line 395: **WARN — weak regression guard.** The docstring says this is the N+1 regression guard, but nothing here actually bounds queries: `django_assert_num_queries` is accepted as a fixture but never invoked as a context manager, and the trailing assertion `b"voter" in resp.content.lower() or resp.status_code == 200` is a tautology because `status_code == 200` is already asserted on the line above. As written, this test would still pass if `snapshots_participated` were re-added and it started doing N+1 scans again.

Consider wrapping the request in `with django_assert_num_queries(MAX):` where `MAX` is something like 15 — enough slack for admin's own queries, tight enough to fail if someone reintroduces a per-row scan. The current approach works as a smoke test but doesn't fulfill the claim in the docstring.

PastLivesReviewBot (approved):
All three items from the prior review are fixed and have regression tests. One minor WARN inline. Approving.

PR #75, opened 2026-04-15
hotfix(1.5.7): guild standings cross-join bug

HexagonStorms:
## Summary
- Current Vote Standings was displaying inflated point totals (e.g. Leatherwork + Woodworking at 60 pts when the real totals were 17 and 18). Root cause: the three `Count()` annotations on `first_choice_votes`, `second_choice_votes`, and `third_choice_votes` cross-joined in a single query, so each count got multiplied by the other two. A guild with 1/2/3 votes at first/second/third rank came out as 6/6/6.
- Adding `distinct=True` on each `Count` makes Django count distinct VotePreference rows per rank, restoring the correct totals.
- Fixed in all three sites: `hub/views.py::_compute_live_standings`, `hub/views.py::_compute_new_votes_since`, and `plfog/dashboard.py::dashboard_callback`.
- Removed the misleading 1.5.6 changelog line that claimed standings were fixed — the signed-up-member filter in that release was correct but didn't touch this bug.

## Test plan
- [x] New regression test in `tests/hub/guild_voting_spec.py` reproduces the exact 60-pt scenario from production data and now passes.
- [x] Full suite: 1314 tests passing at 100% coverage.
- [x] Verified against the live Render database — standings now render as Jewelry 21 / Ceramics 18 / Wood 18 / Glass 17 / Leather 17 / …

### DIFF
--- a/tests/hub/guild_voting_spec.py
+++ b/tests/hub/guild_voting_spec.py
@@ -534,6 +534,32 @@ def describe_compute_live_standings():
         assert "Backfill Only" not in names
         assert names == {"Signed Up Pick", "Signed Up 2nd", "Signed Up 3rd"}
 
+    def it_does_not_cross_multiply_counts_when_one_guild_appears_at_multiple_ranks():
+        """Regression: three Count annotations cross-joined reverse FKs and multiplied.
+
+        With 1 first-place, 2 second-place, and 3 third-place votes for the same
+        guild, the raw JOIN produced 1*2*3 = 6 rows; each un-distinct Count
+        returned 6, yielding 6*5 + 6*3 + 6*2 = 60 instead of 1*5 + 2*3 + 3*2 = 17.
+        """
+        target = GuildFactory(name="Popular Across Ranks")
+        # Filler guilds so each VotePreference has three distinct slots.
+        fillers = [GuildFactory(name=f"Filler {i}") for i in range(6)]
+
+        # 1 first-place vote for target
+        VotePreferenceFactory(guild_1st=target, guild_2nd=fillers[0], guild_3rd=fillers[1])
+        # 2 second-place votes for target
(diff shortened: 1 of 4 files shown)


tests/hub/guild_voting_spec.py
```diff
@@ -534,6 +534,32 @@ def it_excludes_votes_from_members_without_linked_user():
...
+        VotePreferenceFactory(guild_1st=fillers[1], guild_2nd=target, guild_3rd=fillers[2])
+        # 3 third-place votes for target
+        VotePreferenceFactory(guild_1st=fillers[2], guild_2nd=fillers[3], guild_3rd=target)
+        VotePreferenceFactory(guild_1st=fillers[3], guild_2nd=fillers[4], guild_3rd=target)
+        VotePreferenceFactory(guild_1st=fillers[4], guild_2nd=fillers[5], guild_3rd=target)
+
+        result = _compute_live_standings()
+
+        by_name = {r["guild_name"]: r for r in result}
+        assert by_name["Popular Across Ranks"]["total_points"] == 1 * 5 + 2 * 3 + 3 * 2
```
PastLivesReviewBot, line 561: WARN: the same cross-join fix is applied to `_compute_new_votes_since` and `plfog/dashboard.py::dashboard_callback`, but this regression test only covers `_compute_live_standings`. Since all three sites share the bug shape, adding a parallel case for at least one of the other two would prevent a future refactor from silently regressing them. Not blocking — the fix is correct and 100% coverage is maintained — but worth a follow-up.

PastLivesReviewBot (approved):
Root-caused and fixed cleanly. The `distinct=True` explanation in the comments is accurate, the regression test reproduces the exact 60-pt scenario from prod data, and the 1.5.6 changelog correction is the right call. Verified the fix matches observed prod standings (Jewelry 21 / Ceramics 18 / Wood 18 / Glass 17 / Leather 17). Approving.

PR #76, opened 2026-04-15
feat(calendar): classes color, pagination, loading indicator, UX polish (1.6.1)

HexagonStorms:
## Summary

- **Classes color**: Classes events now have a separate configurable color in `SiteConfiguration` (admin fieldset), independent of General calendar
- **Event filtering**: Week and month event lists now show only events within the current navigated range (was showing 90-day horizon)
- **Month pagination**: Month view paginated at 10 events per page with Prev/Next arrows
- **Classes sync button**: Syncing 700+ classes records moved from auto-run on every admin save to a manual "Sync Classes Now" button
- **Admin UX**: General Calendar and Classes sections separated with descriptive fieldsets; color picker and last-fetched timestamp share a row
- **Register button**: Classes event titles non-clickable; "Register →" button appears next to the "Classes" source label in Tuscan Yellow
- **Export button**: Renamed to "Export Calendar" with a note clarifying it exports all community events (not just registered classes)
- **Tooltip**: `?` icon next to page title explains the 5-minute auto-refresh
- **Loading indicator**: Animated Tuscan Yellow bar at the top of the events area during week/month/page navigation

## Test plan

- [ ] Navigate between weeks — loading bar appears and disappears after content loads
- [ ] Navigate between months — same
- [ ] Paginate month events — same
- [ ] Classes events show their own color dot on the calendar grid (not General color)
- [ ] Week view event list shows only events in that week range
- [ ] Month view event list shows only events in that month, paginated
- [ ] Admin: save `SiteConfiguration` — only general calendar syncs (no classes sync)
- [ ] Admin: click "Sync Classes Now" — syncs and shows success message
- [ ] Classes event in list has non-clickable title and "Register →" button
- [ ] Export Calendar dropdown note is present
- [ ] `?` tooltip appears on hover next to "Community Calendar" title


PastLivesReviewBot (dismissed):
Reviewed the full diff. All clear — approving.

**Checked:**
- Type annotations: all new functions fully annotated
- help_text: all new model fields covered
- Migrations: purely additive AddField ops, no RunPython, no direct model imports
- N+1: CalendarEvent queries use select_related("guild")
- Exception handling in refresh_stale_sources: intentional — documented to prevent one bad calendar from crashing the page
- Security: no XSS vectors, auth unchanged
- Architecture: business logic stays in models/service, views are thin
- Tests: BDD specs added for new fields and view behavior

**Minor WARN (non-blocking):**
- The week and month event list blocks in calendar_content.html are nearly identical — worth extracting to a shared include if this template grows further.
- _get_calendar_context in hub/views.py could use a docstring noting the week_offset/month_offset/event_page parameter semantics.

core/admin.py
```diff
@@ -2,18 +2,74 @@
```
PastLivesReviewBot, line 2: **WARN:** The `# type: ignore[attr-defined]` suppressor on `sync_classes_button.short_description` is the correct fix here (Django adds it dynamically), but a brief inline note like `# Django sets this at class instantiation` would help future readers understand why mypy can't see it.

hub/views.py
```diff
@@ -3,9 +3,11 @@
 from __future__ import annotations
```
PastLivesReviewBot, line 3: **WARN:** `default_filters_json` is built with `json.dumps(...)` directly in the view and then embedded in the Alpine `x-data` attribute in the template. Works correctly as-is; just worth noting that if filter serialization grows more complex this could move to a dedicated template tag.

PastLivesReviewBot (approved):
Full diff reviewed — no BLOCK-level issues found. Approving with two minor WARNs as inline comments.

**Verified:**
- All migrations are additive-only `AddField` ops; no `RunPython`, no direct model imports
- All new model fields have `help_text`
- `TextChoices` used for `CalendarEvent.Source`
- All new functions have complete type annotations
- `select_related("guild")` on every `CalendarEvent` queryset
- No business logic in views — calendar context built in `_get_calendar_context`, event filtering is data shaping not business logic
- Test coverage added for new fields and view behaviour; factory-boy used throughout
- Duplicate event item template extracted to `calendar_event_item.html` partial — DRY

PR #82, opened 2026-04-18
feat(hub): Viewing-as role preview popover (1.6.3)

HexagonStorms:
## Summary

Adds a minimal "Viewing as" popover to the hub topbar so admins (and guild officers, once we gate more UI on their role) can preview the hub the way a member of a different role would see it — without logging out.

- **Session-backed:** toggles are stored in `session["view_as_hidden_roles"]`, never persisted to the DB.
- **Fail-safe:** the popover can only *hide* roles the user actually holds — you can't grant yourself admin by poking the endpoint.
- **Zero schema changes:** roles derive entirely from the existing `member.fog_role` field.

### What you see

- Admins get a new "Viewing as" button in the topbar next to "Admin View".
- Clicking it opens a small popover with a checkbox per role the user holds (Member, Guild Officer, Admin).
- Unchecking a role reloads the page and the corresponding UI disappears for the current session. So far the only gated hub UI is the "Admin View" button — this PR installs the scaffolding; per-role gating of other affordances can land in follow-up PRs as they're needed.

### How it works

- `hub/view_as.py` — `compute_actual_roles(user)`, `ViewAs` wrapper, `ViewAsMiddleware` that attaches `request.view_as` to every request.
- `POST /view-as/toggle/` — JSON endpoint `{"role": "...", "hidden": bool}` → writes to session. Rejects unknown roles (400) and roles-you-don't-hold (403).
- `templates/hub/base.html` — topbar popover + Alpine component; the existing "Admin View" button is re-gated on `request.view_as.is_admin` instead of `user.is_staff` so hiding admin removes the button.
- 26 BDD specs in `tests/hub/view_as_spec.py` covering role derivation, session read/write, middleware, the toggle endpoint, and template rendering.

Plan: `docs/superpowers/plans/2026-04-17-viewing-as.md`

## Test plan

- [ ] Log in as an admin → "Viewing as" button appears in topbar.
- [ ] Open popover → three checkboxes (Member, Guild Officer, Admin), all checked.
- [ ] Uncheck "Admin" → page reloads, "Admin View" button disappears.
- [ ] Re-check "Admin" → "Admin View" button comes back.
- [ ] Log in as a plain member → no "Viewing as" button.
- [ ] `curl -X POST /view-as/toggle/` with `role=wizard` → 400.
- [ ] `curl -X POST /view-as/toggle/` as a plain member with `role=admin` → 403.
- [ ] Log out and back in → hidden roles reset (session-only, by design).

### DIFF
--- a/hub/view_as.py
+++ b/hub/view_as.py
@@ -44,7 +44,7 @@ def compute_actual_roles(user: "AbstractBaseUser | AnonymousUser | None") -> fro
     from membership.models import Member as MemberModel
 
     try:
-        member = user.member
+        member = user.member  # type: ignore[union-attr]
     except MemberModel.DoesNotExist:
         member = None
 
@@ -124,5 +124,5 @@ class ViewAsMiddleware:
         self.get_response = get_response
 
     def __call__(self, request: "HttpRequest"):
-        request.view_as = ViewAs.for_request(request)
+        request.view_as = ViewAs.for_request(request)  # type: ignore[attr-defined]
         return self.get_response(request)
--- a/hub/view_as.py
+++ b/hub/view_as.py
@@ -1,0 +1,113 @@
+"""Session-backed "Viewing as" role preview for the hub.
+
+A Member's *actual* roles are derived from ``member.fog_role``. Any subset
+can be *hidden* for the current session via the topbar popover — hidden role
+names are stored in ``request.session["view_as_hidden_roles"]``.
+
+*Effective* roles = actual − hidden. Every hub-side UI gate and convenience
+check should read ``request.view_as`` (attached by ``ViewAsMiddleware``) so
+the session override is respected uniformly.
+"""
+
+from __future__ import annotations
+
+from typing import TYPE_CHECKING
+
+if TYPE_CHECKING:
+    from django.contrib.auth.models import AbstractBaseUser, AnonymousUser
+    from django.http import HttpRequest
+
+ROLE_MEMBER = "member"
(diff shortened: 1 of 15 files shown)


hub/view_as.py
```diff
@@ -0,0 +1,158 @@
...
+    """Attach ``request.view_as`` to every request.
+
+    Must run after ``AuthenticationMiddleware`` (needs ``request.user``)
+    and ``SessionMiddleware`` (needs ``request.session``).
+    """
+
+    def __init__(self, get_response) -> None:  # noqa: ANN001
+        self.get_response = get_response
+
+    def __call__(self, request: "HttpRequest"):
```
PastLivesReviewBot, line 156: **WARN:** Missing return type annotation on `__call__`. Per CLAUDE.md §6 ("All functions typed including `-> None`"), this should have an explicit return type — e.g. `-> HttpResponseBase` or the actual response type returned by `get_response`. Not blocking since the rest of the module is fully annotated.

PastLivesReviewBot (approved):
Clean, well-scoped feature with strong BDD coverage. Middleware wiring is correct (runs after session + auth), session override is fail-safe (`has_actual` guards against privilege escalation), and CSRF/auth on `view_as_set` is properly enforced. Approving with a couple of minor warns — nothing blocking.

Heads-up that the PR description still describes the earlier popover-with-checkboxes design (`POST /view-as/toggle/` with `{role, hidden}`); the final implementation is a single-select dropdown at `POST /view-as/set/` with `{role}`. Worth updating the description so future readers aren't misled.

PR #83, opened 2026-04-21
feat(classes): Plan 1 — foundation & admin tabs (v1.7.0-alpha.1)

HexagonStorms:
## Summary

First of three PRs implementing the Class Management System. This one lands the backend foundation + the admin UI — admins can build out a full class catalog end-to-end from `/classes/admin/`, but there's no public-facing registration flow yet (that's Plan 2) and no instructor self-service (that's Plan 3).

**Spec:** `docs/superpowers/specs/2026-04-20-classes-design.md`
**Plan:** `docs/superpowers/plans/2026-04-20-classes-01-foundation.md`

### What's in this PR

- **New `classes/` Django app** with 10 models: Category, Instructor, ClassOffering, ClassSession, Registration (+ member-by-verified-email auto-linking + unique self_serve_token), Waiver, RegistrationReminder, DiscountCode (+ apply_to/is_currently_valid), ClassSettings singleton (+ seeded default waiver text via data migration)
- **New `instructor` role** added to `hub/view_as.py` — parallel to the member hierarchy, derived from the existence of an Instructor row. A user can be a member, an instructor, or both. The view-as dropdown correctly excludes instructor (it's a capability, not a tier).
- **Admin Classes UI at `/classes/admin/`** with six tabs: Classes (list + CRUD + approve/archive/duplicate), Categories (CRUD), Instructors (list + invite flow that creates User + verified EmailAddress + Instructor), Registrations (list + detail with waiver viewer + cancel), Discount Codes (CRUD), Settings (waiver text, member discount %, reminder timing, MailChimp, GA, `enabled_publicly` toggle)
- **Sidebar link** to `/classes/admin/` at the top of the hub sidebar when viewing as admin
- **Auto-slug** — class title and category name auto-fill the URL slug as the user types (stops when the user edits the slug manually)
- **Context processor** `hub.context_processors.hub_sidebar` puts `{guilds, user_initials}` on every template extending `hub/base.html` — fixes an empty Guilds dropdown bug that appeared on any non-hub view
- **Settings guard** — `plfog/settings.py` now refuses to silently fall back to on-disk sqlite for local dev; pytest still gets the sqlite escape hatch

### Not in this PR (future work)

- Public `/classes/` portal and registration flow — **Plan 2**
- Stripe Checkout Session helper + webhook — **Plan 2**
- Passwordless self-serve (`/classes/my/<token>/`) + email lookup — **Plan 2**
- Confirmation + instructor-alert emails + MailChimp + GA tag — **Plan 2**
- Instructor dashboard (`/classes/instructor/`) + draft→pending self-submit — **Plan 3**
- Upcoming-class reminder scheduler — **Plan 3**
- Calendar integration (merge local Class/Session into community calendar) — **Plan 3**
- `enabled_publicly=True` launch flip + final v1.7.0 version bump — **Plan 3**

### Version

Bumped to `1.7.0-alpha.1`. The final `1.7.0` stable version lands in Plan 3's PR.

## Test plan

- [ ] `pytest classes/ tests/hub/view_as_spec.py` — 110 specs pass locally
- [ ] `ruff check classes/` — clean
- [ ] `python manage.py migrate` — 8 classes migrations apply cleanly (0001_initial through 0008_seed_settings)
- [ ] Sidebar: log in as admin → "Classes" link at top of sidebar → click through all 6 tabs
- [ ] Classes tab: create a Category, invite an Instructor, create a Class (auto-slug fires), submit for review, approve, archive, duplicate
- [ ] Settings tab: defaults populated from seed migration; save a change → persists
- [ ] Role: invite an Instructor → that User can log in via email-code and gets `view_as.is_instructor == True` (dashboard not built yet, but role check works)
- [ ] Django admin (`/admin/`): all 10 classes models appear via auto-admin
- [ ] Non-admin members hitting `/classes/admin/*` → 403


classes/views.py
```diff
@@ -0,0 +1,288 @@
...
+        request,
+        "classes/admin/class_detail.html",
+        {"active_tab": "classes", "offering": offering},
+    )
+
+
+@admin_required
+def admin_class_approve(request: HttpRequest, pk: int) -> HttpResponse:
+    offering = get_object_or_404(ClassOffering, pk=pk)
+    offering.approve(request.user)
```
PastLivesReviewBot: **BLOCK — CSRF via GET.** This view mutates state (publishes the class) on any HTTP method. A logged-in admin hitting a crafted `` or a link-prefetch would silently publish. Django's CSRF middleware only guards unsafe methods, so GET bypasses it entirely.

Guard on method like `admin_category_delete` does:

```python
@admin_required
def admin_class_approve(request: HttpRequest, pk: int) -> HttpResponse:
    offering = get_object_or_404(ClassOffering, pk=pk)
    if request.method == "POST":
        offering.approve(request.user)
        messages.success(request, f"{offering.title} is published.")
    return redirect("classes:admin_class_detail", pk=offering.pk)
```

And add a spec like `it_ignores_get_on_approve_and_redirects` mirroring the one already in `admin_categories_spec.py`.

classes/views.py
```diff
@@ -0,0 +1,288 @@
...
+    messages.success(request, f"{offering.title} is published.")
+    return redirect("classes:admin_class_detail", pk=offering.pk)
+
+
+@admin_required
+def admin_class_archive(request: HttpRequest, pk: int) -> HttpResponse:
+    offering = get_object_or_404(ClassOffering, pk=pk)
+    offering.archive()
```
PastLivesReviewBot: **BLOCK — same CSRF-via-GET issue as `admin_class_approve`.** Please gate `offering.archive()` behind `request.method == "POST"` and add a spec confirming GET is a no-op.

classes/views.py
```diff
@@ -0,0 +1,288 @@
...
+@admin_required
+def admin_class_duplicate(request: HttpRequest, pk: int) -> HttpResponse:
+    src = get_object_or_404(ClassOffering, pk=pk)
+    src.pk = None
+    src.slug = f"{src.slug}-copy"
+    src.status = ClassOffering.Status.DRAFT
+    src.published_at = None
+    src.approved_by = None
+    src.title = f"{src.title} (copy)"
+    src.save()
```
PastLivesReviewBot: **BLOCK — two problems here.**

1. CSRF via GET: `src.save()` runs on any method. Must gate on `request.method == "POST"`.
2. Fat-models/skinny-views violation: clearing `pk`/`published_at`/`approved_by`, slug-renaming, and title-renaming are business-logic state changes that belong on the model. Move this into a `ClassOffering.duplicate() -> ClassOffering` method and have the view just call it:

```python
# models.py
def duplicate(self) -> "ClassOffering":
    self.pk = None
    self.slug = f"{self.slug}-copy"
    self.title = f"{self.title} (copy)"
    self.status = self.Status.DRAFT
    self.published_at = None
    self.approved_by = None
    self.save()
    return self

# views.py
if request.method == "POST":
    copy = get_object_or_404(ClassOffering, pk=pk).duplicate()
    messages.success(request, "Class duplicated.")
    return redirect("classes:admin_class_edit", pk=copy.pk)
return redirect("classes:admin_class_detail", pk=pk)
```

Side note: `{slug}-copy` will collide on the second duplicate (slug is unique). Worth handling with an incrementing suffix inside the new model method.

classes/models.py
```diff
@@ -0,0 +1,441 @@
...
+    )
+    first_name = models.CharField(max_length=100, help_text="Registrant first name.")
+    last_name = models.CharField(max_length=100, help_text="Registrant last name.")
+    pronouns = models.CharField(max_length=50, blank=True, help_text="Optional pronouns.")
+    email = models.EmailField(help_text="Registrant email — drives member linking + self-serve link.")
+    phone = models.CharField(max_length=20, blank=True, help_text="Optional phone.")
+    address_line1 = models.CharField(max_length=255, blank=True)
+    address_city = models.CharField(max_length=100, blank=True)
+    address_state = models.CharField(max_length=50, blank=True)
+    address_zip = models.CharField(max_length=20, blank=True)
```
PastLivesReviewBot: **BLOCK — missing `help_text`.** CLAUDE.md section 3 requires `help_text` on every model field. The four address fields (lines 315–318) don't have it. Even something minimal like `help_text="Street address (optional)."` / `"City (optional)."` / etc. satisfies the rule and gives the admin form meaningful hover hints.

templates/classes/admin/classes_list.html
```diff
@@ -0,0 +1,41 @@
...
+    <tbody>
+    {% for c in classes %}
+        <tr style="border-bottom:1px solid var(--hub-border);">
+            <td style="padding:0.5rem 0.75rem;">
+                <a href="{% url 'classes:admin_class_detail' pk=c.pk %}">{{ c.title }}</a>
+            </td>
+            <td style="padding:0.5rem 0.75rem;">{{ c.instructor.display_name }}</td>
+            <td style="padding:0.5rem 0.75rem;">{{ c.category.name }}</td>
+            <td style="padding:0.5rem 0.75rem;">{{ c.get_status_display }}</td>
+            <td style="padding:0.5rem 0.75rem;">{{ c.registrations.count }}/{{ c.capacity }}</td>
```
PastLivesReviewBot: **BLOCK — N+1.** `{{ c.registrations.count }}` fires a separate `SELECT COUNT(*)` per row. Section 9 ("Avoid N+1") explicitly calls this out as a must-fix.

Annotate in the view instead:

```python
from django.db.models import Count

classes = (
    ClassOffering.objects
    .select_related("instructor", "category")
    .annotate(registration_count=Count("registrations"))
    .order_by("-created_at")
)
```

Then render `{{ c.registration_count }}/{{ c.capacity }}`.

templates/classes/admin/instructors.html
```diff
@@ -0,0 +1,30 @@
...
+            <th style="text-align:left; padding:0.5rem 0.75rem; font-weight:600;">Classes</th>
+            <th style="text-align:left; padding:0.5rem 0.75rem; font-weight:600;">Active</th>
+        </tr>
+    </thead>
+    <tbody>
+    {% for i in instructors %}
+        <tr style="border-bottom:1px solid var(--hub-border);">
+            <td style="padding:0.5rem 0.75rem;">{{ i.display_name }}</td>
+            <td style="padding:0.5rem 0.75rem;">{{ i.user.email }}</td>
+            <td style="padding:0.5rem 0.75rem;">{{ i.classes.count }}</td>
```
PastLivesReviewBot: **BLOCK — same N+1 pattern** as the classes list. `{{ i.classes.count }}` triggers one `COUNT(*)` per instructor. Annotate in `admin_instructors` view:

```python
instructors = (
    Instructor.objects
    .select_related("user")
    .annotate(class_count=Count("classes"))
)
```

Then render `{{ i.class_count }}`.

classes/migrations/0008_seed_settings.py
```diff
@@ -0,0 +1,26 @@
+"""Seed ClassSettings singleton with default waiver text."""
+
+from django.db import migrations
+
+
+def forward(apps, schema_editor):
+    from classes.models import DEFAULT_LIABILITY_TEXT, DEFAULT_MODEL_RELEASE_TEXT
```
PastLivesReviewBot: **WARN.** Migration imports `DEFAULT_LIABILITY_TEXT` / `DEFAULT_MODEL_RELEASE_TEXT` from `classes.models` at forward-run time. Migrations should be frozen snapshots — if either constant is renamed or deleted in a future refactor, replaying this migration on a fresh DB will raise `ImportError`. Safer to inline the default text directly in the migration file. Not blocking since the constants are stable, but worth fixing.

classes/models.py
```diff
@@ -0,0 +1,441 @@
...
+                user__emailaddress__verified=True,
+            )
+            .distinct()
+            .first()
+        )
+        if match is not None:
+            self.member = match
+            super().save(update_fields=["member"])
+
+    def cancel(self, reason: str = "") -> None:
```
PastLivesReviewBot, line 401: **WARN.** `reason` is accepted but never persisted. Either add a `cancellation_reason` `TextField` to the model and store it, or drop the parameter. As written, admins filling in the "Reason (internal)" input on the detail page think it's being recorded — it isn't.

classes/views.py
```diff
@@ -0,0 +1,288 @@
...
+from django.contrib import messages
+from django.contrib.auth.decorators import login_required
+from django.http import HttpRequest, HttpResponse, HttpResponseForbidden
+from django.shortcuts import get_object_or_404, redirect, render
+
+from classes.forms import CategoryForm, ClassOfferingForm, ClassSettingsForm, DiscountCodeForm, InstructorInviteForm
+from classes.models import Category, ClassOffering, ClassSettings, DiscountCode, Instructor, Registration
+
+if TYPE_CHECKING:
+    pass
```
PastLivesReviewBot: **WARN.** Empty `if TYPE_CHECKING: pass` block is dead code — remove it (and the `TYPE_CHECKING` import if no other uses).

PastLivesReviewBot (changes requested):
Really nice foundation PR — the model layer, factories, BDD specs, and admin UI all hang together cleanly. Requesting changes on a few correctness issues before approving:

**BLOCKs**
1. Three state-mutating views (`admin_class_approve`, `admin_class_archive`, `admin_class_duplicate`) run on GET. This bypasses CSRF protection — a logged-in admin following an `` or prefetched link could unknowingly publish/archive/duplicate a class. They should gate on `request.method == "POST"` like `admin_category_delete` and `admin_registration_cancel` already do, and the corresponding specs should assert GET is a no-op.
2. N+1 in the classes list + instructors list — each row fires a fresh `COUNT(*)` via `c.registrations.count` and `i.classes.count`. This is the exact pattern CLAUDE.md section 9 calls out. Use `annotate(...=Count("registrations"))` / `annotate(...=Count("classes"))` on the queryset.
3. `Registration`'s four address fields are missing `help_text`, violating the "help_text on every field" rule from section 3.
4. `admin_class_duplicate` does field-mutation business logic in the view (clearing pk, slug-renaming, title-renaming, clearing timestamps/approver). Per fat-models/skinny-views this should be a `ClassOffering.duplicate()` model method that returns the new instance.

**WARNs** (won't hold up approval once the above are fixed)
- `classes/migrations/0008_seed_settings.py` imports `DEFAULT_LIABILITY_TEXT` / `DEFAULT_MODEL_RELEASE_TEXT` from `classes.models`. If those constants ever get renamed/removed, this migration becomes unreplayable. Safer to inline the default text directly in the migration file (migrations are supposed to be frozen snapshots).
- `Registration.cancel(reason="")` accepts `reason` and throws it away — either store it on the model or drop the parameter.
- Empty `if TYPE_CHECKING: pass` block in `classes/views.py` — dead code, can just delete.

No security, auth, data-integrity, or migration-proxy issues beyond the CSRF-via-GET listed above. The role-capability split for `instructor`, the singleton settings pattern, and the member auto-link by verified email all look great.

PastLivesReviewBot (approved):
All BLOCKs from the previous review are resolved:

- ✅ `admin_class_approve/archive/duplicate` now gate on `request.method == "POST"` and the three `it_ignores_get_on_*_and_redirects` specs assert GET leaves state untouched
- ✅ N+1s killed — `admin_classes` annotates `registration_count`, `admin_instructors` annotates `class_count`, templates render the annotated fields
- ✅ `ClassOffering.duplicate()` is now a model method with a collision-safe slug loop; `it_gives_second_copy_a_unique_slug` covers the `-copy-2` fallback
- ✅ All four `Registration` address fields plus the auto-timestamp fields have `help_text`

All WARNs are also resolved:

- ✅ Migration 0008 inlines the waiver text — no more live import from `classes.models`
- ✅ `Registration.cancellation_reason` field persists the reason from `cancel(reason=...)`, spec updated
- ✅ Empty `TYPE_CHECKING: pass` block removed from `classes/views.py`

Migration 0009 is pure schema (AddField + AlterField) with correct dependency on 0008; no RunPython to worry about. Version intentionally kept at 1.7.0.1 per authoring call. Ship it.

PR #84, opened 2026-04-21
fix(deploy): dummy DATABASE_URL for collectstatic build (1.7.0.2)

HexagonStorms:
## Summary
- Render deploys were failing in the Docker build at `RUN python manage.py collectstatic --noinput` because `settings.py` hard-fails when `DATABASE_URL` is unset.
- `collectstatic` never touches the database — pass a throwaway `sqlite:///tmp/build.db` just for that build step. Runtime still uses Render's real `DATABASE_URL`.
- Bumps version to 1.7.0.2 with a member-friendly changelog entry.

## Test plan
- [ ] Merge and confirm Render build completes past the `collectstatic` step
- [ ] Confirm app boots on Render with the real `DATABASE_URL` from env
- [ ] Confirm Discord release announcement posts on merge

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -15,7 +15,7 @@ RUN pip install --no-cache-dir -r requirements.txt
 
 COPY . .
 
-RUN python manage.py collectstatic --noinput
+RUN DATABASE_URL="sqlite:///tmp/build.db" python manage.py collectstatic --noinput
 
 EXPOSE ${PORT:-10000}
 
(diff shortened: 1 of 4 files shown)


PastLivesReviewBot (approved):
Clean hotfix. Dockerfile change is correctly scoped to the build step (collectstatic never touches the DB, and dj_database_url.parse is a string parse). The classes refactor unblocks the pre-push mypy check by colocating ClassOfferingQuerySet with its model and parametrizing it as QuerySet["ClassOffering"] — matches the pattern used in membership/billing. Tests pass, mypy clean, version + changelog bumped.

PR #85, opened 2026-04-22
feat: image uploads on Cloudflare R2 (1.7.1)

HexagonStorms:
## Summary
- Switches the default Django storage backend to **Cloudflare R2** (S3-compatible) so images uploaded by users actually persist on Render.com — Render's web-service filesystem is wiped on every deploy, which would silently delete every uploaded image.
- Adds `Member.profile_photo` (shown next to a member's name in `/members/`) and `Guild.banner_image` (shown above the title on `/guilds/<pk>/`).
- Adds a 5 MB upload-size validator (`core.validators.validate_image_size`) wired into every existing and new `ImageField` (members, guilds, classes, spaces).
- Existing class images (`Category.hero_image`, `Instructor.photo`, `ClassOffering.image`) automatically benefit from the storage swap — no other code changes needed.
- Storage gracefully falls back to the local filesystem when the `R2_*` env vars aren't set, so tests and unconfigured dev machines keep working.

## Required env vars (Render + Hetzner)
Before merging, set these on Render and Hetzner so the new backend actually engages:
- `R2_ACCOUNT_ID` = `fefad3c0242f171f0bdb68d505689fb4`
- `R2_BUCKET_NAME` = `plfog`
- `R2_ACCESS_KEY_ID` = (from Cloudflare → R2 → API Tokens, shown once at creation)
- `R2_SECRET_ACCESS_KEY` = (from Cloudflare → R2 → API Tokens, shown once at creation)
- `R2_PUBLIC_URL` = (enable "Public Development URL" on the `plfog` bucket and paste it here, e.g. `https://pub-xxxxxxxxxxxx.r2.dev`)

> ℹ️ The `cfat_xxxxx` token alone isn't enough — boto3/django-storages need the Access Key ID + Secret Access Key pair that Cloudflare displays alongside the token. Without all five vars set, the app keeps using local storage (so this is safe to merge before the vars are populated, but uploads won't persist on Render until they are).

## Test plan
- [x] Full pytest suite passes (1524 tests, 99.75% coverage)
- [x] `python manage.py check` clean
- [x] `ruff check` + `ruff format` clean
- [ ] After merge + env vars set: upload a profile photo on `/settings/profile/`, see it in `/members/`
- [ ] After merge + env vars set: upload a banner via the guild edit modal, see it on `/guilds/<pk>/`
- [ ] After merge + env vars set: upload a class hero image in admin, redeploy, verify it still loads


core/files.py
```diff
@@ -0,0 +1,28 @@
...
+
+    Safe to call on creates (returns early) and when the field is unchanged.
+    """
+    if not instance.pk:
+        return
+    model = type(instance)
+    try:
+        old_instance = model._default_manager.only(field_name).get(pk=instance.pk)
+    except model.DoesNotExist:  # type: ignore[attr-defined]
+        return
```
PastLivesReviewBot, line 23: **WARN — swallowed `DoesNotExist`.** CLAUDE.md §4 prefers re-raising `DoesNotExist` as a domain exception rather than silently returning. In a cleanup helper this is arguably the right semantics ("there's nothing to clean up"), and the new `it_returns_when_instance_vanished_between_query_and_save` test documents the intent, but a short inline comment explaining why the silent-return is deliberate would save the next reader from wondering whether this is a hidden bug. Non-blocking.

hub/views.py
```diff
@@ -618,6 +623,34 @@ def user_settings(request: HttpRequest) -> HttpResponse:
...
+@login_required
+@require_POST
+def profile_photo_delete(request: HttpRequest) -> HttpResponse:
+    """Clear the logged-in member's profile photo and redirect back to settings."""
+    member = _get_member(request)
+    if member is None:
+        messages.error(request, "Your account is not linked to a membership.")
+        return redirect("hub_user_settings")
+    if member.profile_photo:
+        member.profile_photo.delete(save=True)
```
PastLivesReviewBot, line 635: **WARN — light business logic in the view.** The single-line `member.profile_photo.delete(save=True)` (and the matching `guild.banner_image.delete(save=True)` below) technically live on the wrong side of the "fat models, skinny views" line — model methods like `member.clear_profile_photo()` / `guild.clear_banner()` would read better and make the domain operation easier to test in isolation. One-liners though, so not worth blocking on. Non-blocking.

PastLivesReviewBot (approved):
Approved. Nice R2 migration — the env-driven storage switch with a clean local-filesystem fallback is the right shape, the reusable `image_field.html` component keeps the hub markup consistent, and the orphan-cleanup helper closes the real gap that would have left dead blobs in R2.

Review pass on the test additions:
- `core/validators.py` now at 100% coverage (size limit, exactly-at-limit, over-limit, missing `.size`)
- `core/files.py` now at 100% coverage (create, unchanged, replaced, cleared, vanished-mid-save)
- `profile_photo_delete` and `guild_banner_delete` both have full auth/method/happy-path/noop coverage, including the 403 branch for `guild_banner_delete`
- Migration `0028` cleanly folds in the `Space.photo` help_text so there's no extra churn migration

Migrations are schema-only (AddField/AlterField), no `RunPython`, no reverse-function concerns. All `reverse()` / URL wiring resolves. No N+1 risks — the new image fields are local columns, not FKs.

Two tiny WARN items left inline — neither is blocking.

**Reminder for deploy:** set `R2_ACCOUNT_ID`, `R2_BUCKET_NAME`, `R2_ACCESS_KEY_ID`, `R2_SECRET_ACCESS_KEY`, and `R2_PUBLIC_URL` on Render and Hetzner before merging, per the PR description — without them the app silently stays on local filesystem storage.

PR #87, opened 2026-04-25
Unify hub: Classes, Calendar, admin pages all live in one dashboard (v1.7.4)

HexagonStorms:
## Summary
- Collapses the dual member/admin layout into one hub. Admins now see one sidebar with everything they need, alphabetized: Classes & Workshops, Community Calendar, Manage Classes & Workshops, Manage Members, Member Directory, My Tab, Payments, Reports, Site Settings, Voting Dashboard.
- Classes public portal at `/classes/` is unconditionally public again — the `enabled_publicly` admin gate is removed (was a fresh-deploy footgun).
- Built hub-native versions of Voting Dashboard, Manage Members, Member Edit, and Site Settings so admin work no longer kicks the user out of the FOG layout into Unfold chrome.
- Active sidebar item gets a gold left-border + tinted background so the current page is unmistakable.
- Login lands every user on Community Calendar.
- "Viewing as" preview dropdown stays so admins can preview what each role sees.

## Scope of changes
**New hub admin pages** (templates under `templates/hub/admin/`):
- `voting_dashboard.html` (`/manage/voting/`) — reuses `plfog.dashboard.dashboard_callback`
- `members.html` (`/manage/members/`) — paginated, status-filtered list
- `member_edit.html` (`/manage/members/<pk>/edit/`) — inline `ModelForm`, dark-theme inputs, resilient to members without linked Users
- `site_settings.html` (`/manage/site-settings/`) — `<input type="color">` for color fields, full input theming

**Permissions**:
- New `hub.view_as.fog_admin_required` decorator — replaces `@staff_member_required` on all 9 billing admin views, lets FOG admins (not just Django staff) access Payments + Reports
- `compute_actual_roles` now grants superusers full hierarchy unconditionally

**Template rehoming** (everything renders inside `hub/base.html`):
- `templates/billing/admin_dashboard.html`, `templates/billing/admin_reports.html` — were Unfold admin templates
- `templates/hub/community_calendar.html` — was the public-site shell
- `templates/classes/base_public.html` — was the public-site shell; CSS rescoped from `body.classes-portal` → `.cp-page` wrapper so navigating between pages doesn't leak styles
- Added `extra_head`, `body_class`, `extra_body` blocks to `hub/base.html`

**Sidebar** (`templates/hub/base.html`):
- All admin links point at hub URLs, not `admin:` URLs
- `target="_blank"` removed from Classes / Calendar — everything stays in-app
- `<nav>` gets `hx-boost="false"` so each click is a full page load — fixes the bug where stale per-page CSS leaked across navigations
- Edit Member, Cancel, Back buttons all explicitly opt out of htmx-boost
- BETA pill moved into the brand area, replacing the old Admin/Member role label

**Adapter** (`plfog/adapters.py`):
- `get_login_redirect_url` lands every user on Community Calendar — no more separate staff-to-/admin/ branch

**Add Instructor flow** (`classes/`):
- Removed `InstructorInviteForm`, `admin_instructor_invite` view, `admin/instructors/new/` URL, `instructor_form.html` template, and `admin_instructors_spec.py`
- Renamed Promote Existing User → Add Instructor; URL is now `/classes/admin/instructors/add/`
- Dropdown labels now pull each user's `Member.full_legal_name` instead of email

**Instructor sidebar**:
- "Teaching" renamed to "Manage Classes & Workshops" (and the dashboard's own header)
- Settings tab remains admin-only (instructor dashboard never had it)

**Migration**:
- `classes/migrations/0013_remove_enabled_publicly.py` — drops `ClassSettings.enabled_publicly`

## Test plan
- [x] `pytest --tb=short`: **1599 passed**
- [x] Coverage: **98.24%** (clears 98% gate)
- [x] `ruff check .`: clean
- [x] `ruff format --check` (tracked files): clean
- [x] `mypy plfog/ core/ membership/ hub/`: clean
- [x] `python manage.py check`: clean
- [x] New `tests/hub/admin_views_spec.py` — 19 specs covering auth, render, save, validation for all four new admin views
- [x] Existing `tests/plfog/adapters_spec.py` updated for the new login redirect target
- [x] `tests/hub/view_as_spec.py` updated for the simplified `ViewAs` API (kept dropdown/session preview tests, dropped `instructor_label` assertions)


hub/views.py
```diff
@@ -997,3 +1008,179 @@ def view_as_set(request: HttpRequest) -> JsonResponse:
...
+                obj.status = Member.Status.FORMER
+            else:
+                obj.fog_role = picked_role
+            obj.save()
+            if picked_role == "instructor" and member.user and not Instructor.objects.filter(user=member.user).exists():
+                Instructor.objects.create(
+                    user=member.user,
+                    display_name=obj.display_name or obj.full_legal_name or member.user.email,
+                    slug=f"instructor-{member.pk}",
+                )
```
PastLivesReviewBot: **BLOCK — business logic in a view.** This block does cross-model state changes (mutates `fog_role`/`status` from a non-model `picked_role` value, then conditionally creates a separate `Instructor` row) — exactly the pattern CLAUDE.md calls out as belonging on the model:

> Single object operations | Model methods (`membership.renew()`, `member.deactivate()`)

Move this to a `Member.apply_admin_role(picked_role)` (or similar) method so:
- the role-token-to-state mapping has one home and is unit-testable in isolation,
- the `Instructor` creation idempotency check sits next to the field it touches,
- the view stays a thin parse-and-dispatch shell.

The `display_name` collision logic for the new `Instructor` (and the slug, see other comment) should live there too.

hub/views.py
```diff
@@ -997,3 +1008,179 @@ def view_as_set(request: HttpRequest) -> JsonResponse:
...
+    ]
+
+    def _initial_role() -> str:
+        if member.status != Member.Status.ACTIVE:
+            return "guest"
+        if Instructor.objects.filter(user=member.user).exists() and member.fog_role == Member.FogRole.MEMBER:
+            return "instructor"
+        return member.fog_role
+
+    class MemberEditForm(django_forms.ModelForm):
```
PastLivesReviewBot: Per `hub/CLAUDE.md` (the Forms section explicitly enumerates `VotePreferenceForm`, `ProfileSettingsForm`, etc.), forms in this app live in `hub/forms.py`. Defining `MemberEditForm` inside the view also recreates the class on every request and makes it untestable in isolation. Same applies to `SiteSettingsForm` below.

hub/views.py
```diff
@@ -997,3 +1008,179 @@ def view_as_set(request: HttpRequest) -> JsonResponse:
...
+                obj.fog_role = Member.FogRole.MEMBER
+                obj.status = Member.Status.FORMER
+            else:
+                obj.fog_role = picked_role
+            obj.save()
+            if picked_role == "instructor" and member.user and not Instructor.objects.filter(user=member.user).exists():
+                Instructor.objects.create(
+                    user=member.user,
+                    display_name=obj.display_name or obj.full_legal_name or member.user.email,
+                    slug=f"instructor-{member.pk}",
```
PastLivesReviewBot: `slug=f"instructor-{member.pk}"` will raise `IntegrityError` if a previous Instructor with this slug exists (e.g. someone was promoted, demoted, the row deleted, then re-promoted — or a manual seed used this slug). `PromoteUserToInstructorForm.save()` in `classes/forms.py` already implements the canonical slug-uniqueness loop — reuse that pattern when this moves to the model method.

hub/view_as.py
```diff
@@ -238,11 +215,33 @@ def for_request(cls, request: "HttpRequest") -> "ViewAs":
         return cls(actual=actual, picked=picked)
 
 
+def fog_admin_required(view_func):  # noqa: ANN001
```
PastLivesReviewBot: CLAUDE.md requires full type annotations on every function. The pre-existing `enabled_publicly_required` (now removed) typed this as `_ViewFunc = Callable[..., HttpResponse]` — please use the same pattern instead of `# noqa: ANN001`.

tests/hub/admin_views_spec.py
```diff
@@ -0,0 +1,274 @@
...
+        @fog_admin_required
+        def view(request):
+            return "ok"
+
+        request = rf.get("/")
+        request.user = AnonymousUser()
+        # No view_as attribute attached — simulates middleware not running.
+        response = view(request)
+        # @login_required wraps the result, so anonymous users get a redirect, not 403.
+        assert response.status_code in (302, 403)
```
PastLivesReviewBot: Test name says "returns 403" but the assertion accepts `(302, 403)`. Either rename to reflect what's actually being asserted (e.g. `it_blocks_anonymous_users`) or split into two tests — one for the missing-middleware → 403 path and one for the anonymous-redirect path.

classes/migrations/0013_remove_enabled_publicly.py
```diff
@@ -0,0 +1,17 @@
+# Generated by Django 6.0.3 on 2026-04-25 19:26
+
+from django.db import migrations
+
+
+class Migration(migrations.Migration):
+    dependencies = [
+        ("classes", "0011_remove_classsettings_google_analytics_measurement_id_and_more"),
+        ("classes", "0012_copy_mailchimp_ga_to_site_settings"),
```
PastLivesReviewBot: Minor: `0012` already depends on `0011`, so listing both here is redundant — `("classes", "0012_copy_mailchimp_ga_to_site_settings")` alone is sufficient.

PastLivesReviewBot (changes requested):
Solid restructuring overall — sidebar unification reads cleanly, the `enabled_publicly` removal is a sensible footgun fix, and 1599 tests with 98.24% coverage is great. Requesting changes on one architecture issue and noting a few smaller items inline.

**BLOCK:** State-mutation + cross-model orchestration logic in `admin_member_edit` belongs on the `Member` model — see inline comment.

**WARN (won't block):** Inline form classes recreated per-request, missing type annotation on `fog_admin_required`, misleading test name, redundant migration deps, slug-collision risk.

hub/views.py
```diff
@@ -997,3 +1016,100 @@ def view_as_set(request: HttpRequest) -> JsonResponse:
...
+@fog_admin_required
+def admin_member_edit(request: HttpRequest, pk: int) -> HttpResponse:
+    """Hub-native edit form for a single Member."""
+    member = get_object_or_404(Member, pk=pk)
+
+    if request.method == "POST":
+        form = MemberAdminEditForm(request.POST, instance=member)
+        if form.is_valid():
+            obj = form.save(commit=False)
+            obj.save()
```
PastLivesReviewBot, line 1086: Nit (non-blocking): the explicit `obj.save()` is redundant — `apply_admin_role()` calls `self.save()` at the end, so this saves the row twice on every successful POST. You can drop line 1086 and let `apply_admin_role` do the single save.

PastLivesReviewBot (approved):
Re-review of `be53dab` — all prior BLOCK + WARN items addressed cleanly:

- ✅ Role-dispatch + Instructor creation now lives on `Member.apply_admin_role()` / `Member._ensure_instructor_record()` with a slugify-uniqueness loop. View is genuinely thin.
- ✅ `MemberAdminEditForm` and `SiteSettingsForm` extracted to `hub/forms.py`; the form derives initial `role` from the instance.
- ✅ `fog_admin_required` typed as `Callable[..., HttpResponse]`, inner `wrapper` typed too.
- ✅ Fog-admin-required test split into `it_redirects_anonymous_users_to_login` (asserts 302) and `it_returns_403_for_authenticated_non_admin` (asserts 403).
- ✅ Migration 0013 deps reduced to just `0011` (which is the actual leaf since `0011` depends on `0012`).

Approving. One non-blocking nit inline.

PastLivesReviewBot (approved):
Re-review of `9f10822` — clean type-narrowing fix for the 3 mypy errors that broke CI lint on `be53dab`.

- `_ensure_instructor_record` now reads `self.user` into a local, early-returns when None, then uses the narrowed local everywhere (`user.pk`, `user.email`, `user`).
- Caller's redundant `self.user_id is not None` guard dropped — the helper is now self-guarding.
- Behavior unchanged; lint job on this SHA passed.

No new issues. Approving.

PR #88, opened 2026-04-26
Classes: public registration + Stripe Checkout, unified page, fixes (v1.7.5)

HexagonStorms:
## Summary

This PR rolls together three related improvements to the Classes & Workshops experience plus a new public-preview affordance.

### 1. Public class registration + Stripe Checkout
- Members and visitors can register for any published class directly on the site.
- Verified-member emails get the member discount auto-applied; discount codes stack on top.
- Free classes skip Stripe and confirm immediately; paid classes go through Stripe Checkout.
- Confirmed registrants get a confirmation email with a self-serve link to view or cancel.
- Logged-in members get the form pre-filled from their Member record (name, email, phone, pronouns).

### 2. Unified Classes & Workshops page
- `/classes/` is now the single sidebar entry; admins/instructors see a Manage button on the page.
- Bug fix: published classes now appear on the list immediately, even before sessions are scheduled (previously hidden until at least one upcoming session existed).

### 3. Public-facing chrome
- For anonymous visitors, the hub sidebar + topbar are hidden site-wide. They get a slim public topbar with the Past Lives wordmark, a Classes link, and a Log in CTA.
- Authenticated users get a **"View Publicly"** button on the Classes & Workshops list and the Community Calendar — clicking it loads the same URL with `?public=1`, rendering the public chrome.
- The public topbar swaps the Log in button for an "← Exit public view" pill while previewing.

### 4. Admin polish
- Classes admin header swaps "View public portal ↗" for an in-tab "← Classes & Workshops" back-nav.

## Architecture

- `RegistrationForm` (classes/forms.py) owns validation, price computation, and waiver creation.
- `register` view orchestrates: free path confirms inline; paid path creates a Stripe Checkout Session and rolls back the half-created Registration on Stripe failure.
- `classes.webhook_handlers.handle_checkout_session_completed` is registered into the existing `billing/views.py:_WEBHOOK_HANDLERS` dispatcher; gated by `metadata.kind == "class_registration"` so it can't collide with future Checkout uses. Idempotent on Stripe re-deliveries.
- `DiscountCode.use_count` is incremented atomically with `F("use_count") + 1`.
- Self-serve via `Registration.self_serve_token` (already on the model) — no auth required.
- "View Publicly" is page-local: `?public=1` makes `hub/base.html` render the public chrome. Doesn't touch the session or the "Viewing as" preference.

## Test plan
- [x] `pytest` — 1649 passed (33+ new tests across form/view/webhook/prefill)
- [x] `ruff check` clean
- [ ] `/classes/<slug>/register/` renders + submits as guest
- [ ] Logged-in member sees prefilled form
- [ ] Verified-member email auto-applies member discount
- [ ] Discount code applies on top of member discount
- [ ] Free class skips Stripe → confirmation email arrives immediately
- [ ] Paid class redirects to Stripe Checkout test card → webhook fires → registration goes CONFIRMED → email arrives
- [ ] Stripe Checkout Cancel button cleans up the pending registration
- [ ] `/classes/my/<token>/` shows the right registration + cancels via POST
- [ ] Sold-out class blocks new registration with a friendly error
- [ ] Admin `/classes/admin/` "← Classes & Workshops" link navigates in-tab
- [ ] As anon visitor, hub chrome is hidden site-wide; public topbar shows
- [ ] As logged-in user, "View Publicly" toggles to public chrome with Exit pill

### DIFF
--- a/classes/views.py
+++ b/classes/views.py
@@ -3,15 +3,19 @@
 from __future__ import annotations
 
 from functools import wraps
-from typing import Any, Callable
+from typing import TYPE_CHECKING, Any, Callable
 
 from django.contrib import messages
 from django.contrib.auth.decorators import login_required
+from django.contrib.auth.models import AbstractBaseUser, AnonymousUser
 from django.db.models import Count, Min, Q
 from django.http import HttpRequest, HttpResponse, HttpResponseForbidden
 from django.shortcuts import get_object_or_404, redirect, render
 from django.urls import reverse
 from django.utils import timezone
+
+if TYPE_CHECKING:
+    from membership.models import Member
 
 from classes.emails import send_registration_confirmation
 from classes.forms import (
@@ -159,7 +163,7 @@ def _client_ip(request: HttpRequest) -> str:
     return request.META.get("REMOTE_ADDR", "")
 
 
-def _member_for_email(email: str):
+def _member_for_email(email: str) -> "Member | None":
     """Verified Member matching this email, or None."""
     from membership.models import Member
 
@@ -173,7 +177,7 @@ def _member_for_email(email: str):
     )
 
 
-def _registration_initial_for_user(user) -> dict[str, str]:
+def _registration_initial_for_user(user: "AbstractBaseUser | AnonymousUser | None") -> dict[str, str]:
     """Pre-fill values pulled from the logged-in user's Member record."""
     if not user or not user.is_authenticated:
         return {}
--- a/classes/views.py
+++ b/classes/views.py
@@ -173,6 +173,24 @@ def _member_for_email(email: str):
     )
 
 
+def _registration_initial_for_user(user) -> dict[str, str]:
+    """Pre-fill values pulled from the logged-in user's Member record."""
+    if not user or not user.is_authenticated:
+        return {}
+    member = getattr(user, "member", None)
+    if member is None:
+        return {"email": user.email or ""}
+    name = (member.preferred_name or member.full_legal_name or user.get_full_name() or "").strip()
+    first_name, _, last_name = name.partition(" ")
+    return {
+        "first_name": first_name or user.first_name or "",
+        "last_name": last_name.strip() or user.last_name or "",
+        "email": member.primary_email or user.email or "",
+        "phone": member.phone or "",
+        "pronouns": member.pronouns or "",
+    }
+
+
 def register(request: HttpRequest, slug: str) -> HttpResponse:
     """Public registration form — collects info, signs waivers, kicks off Stripe Checkout.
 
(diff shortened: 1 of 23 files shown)


classes/views.py
```diff
@@ -146,6 +151,235 @@ def public_instructor(request: HttpRequest, slug: str) -> HttpResponse:
...
 
+def _client_ip(request: HttpRequest) -> str:
+    """Best-effort client IP, honoring X-Forwarded-For when behind a proxy."""
+    forwarded = request.META.get("HTTP_X_FORWARDED_FOR", "")
+    if forwarded:
+        return forwarded.split(",")[0].strip()
+    return request.META.get("REMOTE_ADDR", "")
+
+
+def _member_for_email(email: str):
```
PastLivesReviewBot: **BLOCK — missing return type annotation.** Project standard (CLAUDE.md §6) requires full type annotations on every function. This returns `Member | None`, so:

```python
def _member_for_email(email: str) -> "Member | None":
```

Use a string-quoted forward ref (or add `Member` to the existing `TYPE_CHECKING` block at the top of the file) so the lazy import inside the function body keeps working.

classes/views.py
```diff
@@ -146,6 +151,235 @@ def public_instructor(request: HttpRequest, slug: str) -> HttpResponse:
...
+        Member.objects.filter(
+            user__emailaddress__email__iexact=email,
+            user__emailaddress__verified=True,
+        )
+        .distinct()
+        .first()
+    )
+
+
+def _registration_initial_for_user(user) -> dict[str, str]:
```
PastLivesReviewBot: **BLOCK — `user` parameter is unannotated.** This is the only un-annotated parameter in the file. `request.user` is either an authenticated `User` or `AnonymousUser`. Suggested:

```python
from django.contrib.auth.models import AbstractBaseUser, AnonymousUser

def _registration_initial_for_user(user: "AbstractBaseUser | AnonymousUser | None") -> dict[str, str]:
```

If you'd rather keep imports light, `user: Any` is acceptable per the existing `_browsable_classes() -> Any` precedent — but it can't be left bare.

PastLivesReviewBot (changes requested):
Nice work overall — the registration flow is well-structured, the webhook handler is properly idempotent and gated on `metadata.kind == "class_registration"`, the Stripe failure path rolls back the half-created Registration, and the 33+ new tests give solid coverage across form/view/webhook paths. The unified Classes & Workshops sidebar entry plus the public-chrome toggle read cleanly too.

One small change needed before approval: two new private helpers in `classes/views.py` are missing type annotations, which violates the project's typing standard (CLAUDE.md §6: "All functions must have full type annotations, including `-> None`."). Every other function in this file has complete annotations, so these two stick out. See inline comments. Once those are fixed I'll re-review and approve.

PastLivesReviewBot (approved):
Re-reviewed at `1572c96` — both type-annotation BLOCKs are resolved:

- `_member_for_email(email: str) -> "Member | None"` ✅
- `_registration_initial_for_user(user: "AbstractBaseUser | AnonymousUser | None") -> dict[str, str]` ✅
- `Member` is now imported under a `TYPE_CHECKING` block, keeping the lazy runtime import inside the function body ✅

The rest of the PR remains in good shape: idempotent webhook handler gated on `metadata.kind == "class_registration"`, atomic discount-code increment via `F("use_count") + 1`, Stripe failure rolls back the half-created Registration, and 33+ new tests cover form/view/webhook paths. Approving.

PR #89, opened 2026-04-26
v1.7.6: Free class/workshop checkbox + Register page CSS hotfix

HexagonStorms:
## Summary

Two changes bundled into one release:

### 1. Free class/workshop option (feature)
- New "This is a free class / workshop" checkbox on both `ClassOfferingForm` (admin) and `InstructorClassOfferingForm` (instructor portal). When checked, `price_cents` and `member_discount_pct` are zeroed on save.
- Pre-checked when editing an existing class with `price_cents == 0`.
- Form `clean()` now requires a price unless the box is checked (clear validation error otherwise).
- `cents_as_price` template filter renders 0 as **"Free"** instead of "$0", so the class card, detail page, and registration summary all use the friendlier label automatically.
- The runtime registration flow already supported `price_cents=0` (no Stripe round-trip, instant confirmation email) — this just gives admins/instructors a clear UX for it.

### 2. Register page CSS hotfix + permanent prevention
The Register page lost its styling when navigated to via hx-boost from a class detail page. We've fixed this same class of bug repeatedly. Three layers of fix so it can't recur:

- **Real fix:** Vendor htmx's [head-support](https://htmx.org/extensions/head-support/) extension and load it on `hub/base.html` with `hx-ext="head-support"`. Boost now merges `<head>` contents on every navigation — retroactively fixes every existing page with per-page styles.
- **Pattern fix:** Move the Register page's inline `<style>` into `static/css/classes-register.css`, referenced via `<link>`. Works whether or not head-support is loaded.
- **Recurrence prevention:** New `scripts/check_no_inline_style_in_extra_head.py` lints templates for `<style>` blocks inside any `extra_head` block. Wired into CI (`.github/workflows/ci.yml`) and the local pre-push hook. A baseline grandfathers 5 pre-existing offenders so the lint is additive — but **any new template that puts inline `<style>` in `extra_head` will fail CI**.

## Test plan
- [x] `pytest classes/` — 183 passed (7 new specs for free-class form behavior)
- [x] `ruff check .` clean
- [x] `python3 scripts/check_no_inline_style_in_extra_head.py` exits 0 (baseline matches)
- [ ] Admin: create a class with the "free" box checked → no price required → "Free" shows on the public list
- [ ] Instructor: same flow via the Teaching dashboard
- [ ] Member registers for a free class → no Stripe redirect → confirmation email arrives
- [ ] Visit a class detail page logged-in → click Register → form is fully styled (no hard refresh needed)
- [ ] Same flow as anonymous visitor

## Why we keep hitting the CSS bug
Every time someone (human or AI) writes a new page and quickly drops a `<style>` block in `{% block extra_head %}`, hx-boost silently breaks the styling on intra-app navigation. There's no runtime error — the page just looks broken. Layer 1 (head-support) makes the symptom impossible at runtime; layer 3 (CI lint) makes the lazy pattern impossible at PR time.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -67,3 +67,6 @@ jobs:
 
       - name: Mypy
         run: mypy plfog/ core/ membership/ hub/
+
+      - name: No inline <style> in template extra_head blocks
+        run: python scripts/check_no_inline_style_in_extra_head.py
--- a/classes/forms.py
+++ b/classes/forms.py
@@ -24,7 +24,55 @@ if TYPE_CHECKING:
     from membership.models import Member
 
 
-class ClassOfferingForm(forms.ModelForm):
+class _FreeClassMixin:
+    """Adds an `is_free` checkbox that, when checked, forces price/discount to 0.
+
+    Source of truth remains `price_cents` on the model (0 = free). The checkbox
+    is a UX affordance so the instructor/admin doesn't have to know that "type 0
+    in cents" makes a class free — they just tick a box.
+    """
+
+    def add_is_free_field(self) -> None:
+        instance = getattr(self, "instance", None)
+        initial = bool(instance and instance.pk and instance.price_cents == 0)
+        self.fields["is_free"] = forms.BooleanField(  # type: ignore[attr-defined]
+            required=False,
+            initial=initial,
+            label="This is a free class / workshop",
+            help_text="Check this if there's no fee. Members will be able to register without entering payment info.",
+        )
+        # Price and discount aren't required when the class is free — the form's
+        # clean() enforces that price_cents is filled in for non-free classes.
+        self.fields["price_cents"].required = False  # type: ignore[attr-defined]
+        self.fields["member_discount_pct"].required = False  # type: ignore[attr-defined]
+        # Render the checkbox just above price so the visual flow is "Is this free?
+        # → if not, here's the price." Django keeps this order when iterating `form`.
+        ordered: list[str] = []
+        for name in self.fields:  # type: ignore[attr-defined]
+            if name == "price_cents":
+                ordered.append("is_free")
+            if name == "is_free":
+                continue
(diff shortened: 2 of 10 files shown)


PastLivesReviewBot (approved):
Approving — solid, well-scoped PR covering both pieces of v1.7.6 cleanly.

**What works well:**
- The htmx `head-support` extension is the right *root-cause* fix for the recurring boost-vs-head-CSS bug — it retroactively repairs every page that uses `extra_head`, not just the Register page.
- Defense in depth: vendoring the extension + extracting register CSS to a static file + the CI lint with a baseline — any one layer can fail and the bug still can't ship. The baseline approach (grandfather 5 existing offenders, block new ones) is the pragmatic call here; full migration would have ballooned scope.
- `_FreeClassMixin` cleanly factors the free-class wiring across both `ClassOfferingForm` and `InstructorClassOfferingForm` without touching the model. Keeping `price_cents=0` as the single source of truth is correct — no migration needed and the existing registration runtime (no Stripe round-trip when price is 0) just works.
- `cents_as_price` rendering 0 as "Free" is a small change with a friendly footprint — verified all 10 call sites handle it correctly (class card, instructor page, detail page, registration summary; the redundant `{% if price_cents == 0 %}Free{% else %}...{% endif %}` blocks in `detail.html`/`register.html` continue to work but could now be simplified — minor follow-up, not worth holding the PR for).
- 7 new BDD specs cover both forms (zero-on-save, paid-pricing-preserved, validation when not free, pre-check state) — full classes/ suite stays green at 183 passing.
- All new functions have full type annotations; ruff clean.

**Coverage check:**
- `_FreeClassMixin.apply_is_free_to_instance` is exercised on both branches (`is_free=on` and `is_free` absent) by the new specs.
- `clean_is_free_pricing` validation path is tested.
- Pre-check init for both free and paid existing instances is tested.

Ship it.

PR #90, opened 2026-04-26
Free class/workshop checkbox in the class form (v1.7.6)

HexagonStorms:
## Summary

The registration runtime has supported free classes since v1.7.5 (price_cents=0 → no Stripe round-trip, instant confirmation), but the admin/instructor UX didn't make this obvious — they had to know "type 0 in cents = free." This PR adds a clear affordance.

## What changed

- **\"This is a free class / workshop\" checkbox** on both `ClassOfferingForm` (admin) and `InstructorClassOfferingForm` (instructor portal). Renders just above the price field via `order_fields()`.
- When checked on save: `price_cents` and `member_discount_pct` are zeroed.
- When unchecked on save: `clean()` requires a price (clear error if missing).
- Pre-checked when editing an existing free class (`instance.price_cents == 0`).
- `cents_as_price` template filter now renders `0` as **\"Free\"** instead of `\"$0\"` — so the class card, detail page, and registration summary all show the friendlier label automatically.

## Architecture

- New `_FreeClassMixin` keeps the wiring DRY across the two forms (admin vs instructor).
- No model change — `price_cents=0` is the single source of truth.
- No migration.

## Test plan
- [x] 7 BDD specs in `classes/spec/forms/free_class_spec.py` cover:
    - Checking the box zeros price + discount on save (both forms)
    - Unchecking + leaving price blank fails validation
    - Pre-check state matches existing instance pricing
    - Paid pricing is preserved unchanged when unchecked
- [x] Full `pytest classes/` — 183 passed
- [x] `ruff check classes/` clean
- [ ] Admin: create a class with the box checked → confirms with no price field error → \"Free\" shows on the public list
- [ ] Instructor: same flow via the Teaching dashboard
- [ ] Member registers for a free class → no Stripe redirect → confirmation email arrives

### DIFF
--- a/classes/forms.py
+++ b/classes/forms.py
@@ -24,7 +24,55 @@ if TYPE_CHECKING:
     from membership.models import Member
 
 
-class ClassOfferingForm(forms.ModelForm):
+class _FreeClassMixin:
+    """Adds an `is_free` checkbox that, when checked, forces price/discount to 0.
+
+    Source of truth remains `price_cents` on the model (0 = free). The checkbox
+    is a UX affordance so the instructor/admin doesn't have to know that "type 0
+    in cents" makes a class free — they just tick a box.
+    """
+
+    def add_is_free_field(self) -> None:
+        instance = getattr(self, "instance", None)
+        initial = bool(instance and instance.pk and instance.price_cents == 0)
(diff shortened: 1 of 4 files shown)


HexagonStorms:
Combined into PR #89 — single PR per request.

PR #91, opened 2026-04-27
Hotfix v1.7.7: site layout broken by multi-line {# #} comment in hub/base.html

HexagonStorms:
## Summary
- v1.7.6 added a multi-line `{# ... #}` comment above the head-support `<script>` tag in `templates/hub/base.html`. Django's `{# #}` syntax only matches on a single line — the comment text was rendered into the page body, and the literal `<style>` substring opened a real style block that consumed the rest of the document.
- Symptom: every page on the site rendered at half-width, with a stray fragment of comment text near the top of the body, until the user did a hard refresh on a non-boosted route.
- Fix: switch to `{% comment %}{% endcomment %}`, the only Django syntax that legitimately spans multiple lines. Behavior of the head-support extension itself is unchanged.

## Test plan
- [x] `templates/hub/base.html` renders with no occurrence of `head-support: makes hx-boost` or `silently disappear on boosted` in the output HTML.
- [x] `htmx-ext-head-support.js` `<script>` tag is still present in the rendered head, and `hx-ext="head-support"` is still on `<body>`.
- [x] `scripts/check_no_inline_style_in_extra_head.py` exits 0.
- [ ] Manual: log into the hub, navigate around (Hub → Classes → a class → Register) and confirm full-width layout on every page with no stray text near the top.

### DIFF
--- a/plfog/version.py
+++ b/plfog/version.py
@@ -2,9 +2,17 @@
 
 from __future__ import annotations
 
-VERSION = "1.7.6"
+VERSION = "1.7.7"
 
 CHANGELOG: list[dict[str, str | list[str]]] = [
+    {
+        "version": "1.7.7",
+        "date": "2026-04-27",
+        "title": "Hotfix: site layout broken for everyone",
+        "changes": [
+            "Fixed a bug introduced in the last release where the entire site rendered at half-width with a stray line of text near the top — the page was unusable. Everything looks correct again.",
+        ],
+    },
     {
         "version": "1.7.6",
         "date": "2026-04-26",
--- a/plfog/version.py
+++ b/plfog/version.py
@@ -2,11 +2,11 @@
 
 from __future__ import annotations
 
-VERSION = "1.7.7"
+VERSION = "1.7.5.2"
 
 CHANGELOG: list[dict[str, str | list[str]]] = [
     {
-        "version": "1.7.7",
+        "version": "1.7.5.2",
         "date": "2026-04-27",
         "title": "Hotfix: site layout broken for everyone",
         "changes": [
--- a/templates/hub/base.html
+++ b/templates/hub/base.html
@@ -383,8 +383,10 @@
     {% include "components/toast.html" %}
     {% include "includes/changelog_modal.html" %}
     <script src="{% static 'js/htmx.min.js' %}"></script>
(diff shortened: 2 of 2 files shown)


plfog/version.py
```diff
@@ -2,9 +2,17 @@
 
 from __future__ import annotations
 
-VERSION = "1.7.6"
+VERSION = "1.7.5.2"
```
PastLivesReviewBot, line 5: WARN — version regression: `VERSION` is now `1.7.5.2`, which is semver-less-than the previous `1.7.6` (which is still listed in the CHANGELOG below this entry). The Discord workflow reads the top entry so the announcement will work, but if anything else compares versions as semver this will go backwards. Confirm this is the intent — it reads like 1.7.6 is being soft-yanked, even though this PR keeps the free-class checkbox feature from 1.7.6 in place. If 1.7.6 stays released, a `1.7.6.1` (or `1.7.7`) would be more conventional.

templates/hub/base.html
```diff
@@ -383,8 +383,10 @@
     {% include "components/toast.html" %}
     {% include "includes/changelog_modal.html" %}
     <script src="{% static 'js/htmx.min.js' %}"></script>
-    {# head-support: makes hx-boost merge <head> contents (per-page styles, link tags) — without it,
-       page-specific <link>/<style> in extra_head silently disappear on boosted navigations. #}
+    {% comment %}
+        head-support: makes hx-boost merge <head> contents (per-page styles, link tags) — without it,
+        page-specific <link>/<style> in extra_head silently disappear on boosted navigations.
```
PastLivesReviewBot, line 388: WARN (optional) — recurrence prevention: this is the second `extra_head` / hx-boost / Django-comment foot-gun in two releases. v1.7.6 added `scripts/check_no_inline_style_in_extra_head.py` for the prior class of bug. A sibling lint that scans templates for multi-line `{# ... #}` blocks (i.e. a `{#` whose matching `#}` is on a different line) would prevent this exact recurrence in CI without imposing on PR authors.

PastLivesReviewBot (approved):
Approving — the template fix is correct: `{% comment %}{% endcomment %}` is the only Django syntax that legitimately spans multiple lines, and the head-support extension still loads with `hx-ext="head-support"` on `<body>`. Two non-blocking WARNs left inline.

PR #92, opened 2026-05-01
v1.8.0: Calendar QOL + privacy-aware Profile redesign + mobile feedback FAB

HexagonStorms:
## Summary

A bundle of member-facing QOL improvements:

### Calendar
- **Rolling 4-week month view**: instead of showing the calendar month (which left an empty grid by month-end), the Month view now shows current week + 3 upcoming weeks. Prev/Next steps the window by 4 weeks.
- **Event titles on the grid**: each day cell shows up to 3 chips with truncated titles; cells are fixed-height with overflow-hidden so layout never breaks. Overflow shows `+N`.
- **Click-to-focus**: clicking a chip smooth-scrolls to the matching event card in the list and flashes a colored ring. If the event lives on a different pagination page, the list auto-flips first via `htmx.ajax`.
- **Live-now indicator**: today's column has a thin red line across the top; in-progress events get a red border + pulsing "Live now" pill on their card.

### Profile settings (redesign)
- New `Member.directory_visibility` JSONField + `is_public(field)` helper (defaults to public so existing members aren't accidentally hidden).
- `ProfileSettingsForm` now exposes 7 visibility booleans (`pronouns`, `phone`, `email`, `discord_handle`, `other_contact_info`, `about_me`, `profile_photo`).
- New two-column UI in `user_settings.html`: editable fields on the left (each with a `Public/Hidden` eye-pill), sticky live "What other members see" preview on the right, Listed/Hidden master segmented control at the top, and a "Never shown to other members" reassurance panel at the bottom (legal name, billing, emergency contact, internal notes).
- Member directory respects the per-field flags via a new `is_public:"key"` template filter.

### Mobile fixes
- New floating yellow Feedback bubble pinned to bottom-right on `≤768px` (the in-sidebar Feedback link is hidden on mobile to avoid duplication).
- Sidebar uses `100dvh` so it survives mobile URL-bar collapse, plus `env(safe-area-inset-bottom)` and `overscroll-behavior: contain`.

### Versioning
Bumped to **1.8.0** with member-friendly changelog entry.

## Test plan
- [x] `pytest tests/hub tests/membership` — 604 passed (full suite previously: 1442 passed)
- [x] Calendar `/calendar/` renders new chip + state-island markup (verified via `curl`)
- [ ] Manual: open `/calendar/`, verify Month view shows 4 weeks starting from this week
- [ ] Manual: click a chip in Week and Month view → event card flashes
- [ ] Manual: click a chip whose event is on page 2 → list auto-flips and event highlights
- [ ] Manual: open `/settings/?tab=profile`, toggle a few Public/Hidden pills → preview card updates live; type in About Me → preview updates as you type
- [ ] Manual: flip the master Listed/Hidden segmented control off → preview shows "Hidden" empty state
- [ ] Manual: another member loads `/members/` → fields you marked Hidden don't appear on your card
- [ ] Manual: load any page on a phone (or devtools mobile viewport) → yellow Feedback bubble visible bottom-right; sidebar opens fully without bottom cutoff


hub/forms.py
```diff
@@ -60,10 +62,22 @@ def __init__(self, *args, **kwargs) -> None:
...
+                initial=bool(current.get(field_name, True)),
+                label=f"Show {field_name.replace('_', ' ')} on my directory card",
+            )
+
     def save(self, commit: bool = True) -> Member:
         member = super().save(commit=False)
         if member.must_be_listed_in_directory:
             member.show_in_directory = True
+        member.directory_visibility = {
+            field_name: bool(self.cleaned_data.get(f"{self.VISIBILITY_PREFIX}{field_name}", True))
```
PastLivesReviewBot, line 78: Minor: `cleaned_data.get(..., True)` uses a silent default. Since the visibility fields are always added in `__init__` and `BooleanField(required=False)` resolves missing input to `False`, these keys will always be present. Per CLAUDE.md's fail-loudly rule, `cleaned_data[key]` would be more explicit. (Same nit applies to line 69 — though there the `True` default for missing keys is intentional default-public behavior, which is fine.)

membership/models.py
```diff
@@ -1081,3 +1108,13 @@ def source_key(self) -> str:
         if self.source == self.Source.GUILD and self.guild_id:
             return str(self.guild_id)
         return self.source
+
+    @property
+    def is_in_progress(self) -> bool:
```
PastLivesReviewBot, line 1113: New `is_in_progress` property has no dedicated `*_spec.py` test asserting the True/False outcomes under specific time conditions. CI 100% coverage passes because the property is exercised by calendar template rendering, but a behavior-asserting test (all_day → False; start <= now < end → True; before/after window → False) would be ideal for catching mutations.

PastLivesReviewBot (approved):
Approving — solid set of QOL changes. Migration is clean (additive JSONField, default=dict), types are annotated, fat-models pattern respected, and the directory_visibility default-on logic correctly avoids hiding existing members. Two minor non-blocking notes inline.

PR #93, opened 2026-05-06
Mailchimp signup + site-wide Google Analytics (v1.9.0)

HexagonStorms:
## Summary

- **Mailchimp integration goes live.** The `mailchimp_api_key` and `mailchimp_list_id` fields on Site Settings have promised "auto-subscribe new class registrants" since they shipped — that promise is now real. Three subscribe paths land on a single Mailchimp audience, segmented by tag (`class-registrant`, `newsletter`).
- **Google Analytics expands from the public Classes portal to the whole site.** GA4 fires on every member-facing and public page; the Django admin is excluded.
- **VERSION bumped to 1.9.0** with a member-friendly changelog entry.

### Mailchimp

- New `core/integrations/mailchimp.py` — minimal v3 REST client using `requests.put` against the upsert endpoint (`PUT /lists/{id}/members/{md5(email)}`) so re-subscribing is idempotent and unsubscribed users are not silently re-subscribed. Single opt-in (`status_if_new: "subscribed"`). 5-second timeout. Fails quietly: every public method returns `bool` and never raises.
- New `classes/services/mailchimp_subscribe.py` bridge — called from the free-class flow and the Stripe webhook handler right after the confirmation email. Mailchimp failures are logged but never block registration confirmation.
- New `Registration.wants_newsletter` boolean (migration `classes/0014`). Existing `subscribed_to_mailchimp` is now actually written when a subscribe succeeds.
- New opt-in checkbox on the public class registration form (default unchecked). Renders with the same `reg-check` styling as the waiver checkboxes.
- New standalone `/newsletter/` page — anyone (members or not) can subscribe with email + optional first/last name. Tagged `newsletter` in Mailchimp.
- Implementation plan doc at `docs/mailchimp-implementation-plan.md` — the blueprint this PR follows. Member auto-subscribe (Step 6 in the plan) is **not** in this PR; it's a follow-up.

### Google Analytics

- Extracted gtag block into `templates/includes/google_analytics.html` — single source.
- Included from `templates/base.html` (member hub, home, auth pages, billing, etc.) **and** `templates/classes/base_public.html` (public Classes portal, replacing the old inline duplicate).
- New `core.context_processors.google_analytics` exposes the measurement ID everywhere — but returns empty for any path under `/admin/` so internal admin traffic never reports to GA.
- Help text + admin description updated to say "site-wide, admin excluded."

## Test plan

- [x] `pytest --no-cov -q` — 1699 passed locally.
- [x] `python manage.py makemigrations --check --dry-run` clean.
- [x] `ruff format` + `ruff check` clean across touched files.
- [ ] **Staging Mailchimp smoke test:** set test API key + list ID in Site Settings → register for a free class with the opt-in box ticked → confirm the subscriber appears in the Mailchimp audience tagged `class-registrant`.
- [ ] **Staging Mailchimp paid path:** register for a paid class with opt-in ticked → complete Stripe Checkout → confirm subscribe lands.
- [ ] **Staging Mailchimp opt-out:** register without ticking the box → confirm no API call lands and `subscribed_to_mailchimp` stays False.
- [ ] **Staging Mailchimp newsletter page:** visit `/newsletter/` → subscribe → confirm subscriber tagged `newsletter`.
- [ ] **Staging Mailchimp failure mode:** set a bogus API key → register → confirm the registration still confirms and the confirmation email still sends.
- [ ] **Staging GA:** set a real GA4 ID → load `/`, `/guilds/voting/`, `/classes/`, `/accounts/login/` → confirm pageviews land in GA Realtime within ~30s; load `/admin/` → confirm no GA call fires.
- [ ] **Staging GA disabled:** clear the GA ID → confirm tag disappears on next reload.

## Open questions for product

The implementation plan doc (`docs/mailchimp-implementation-plan.md` §9) has a longer list, but the immediate ones:

1. The opt-in checkbox is **default unchecked** — confirm that's the right call.
2. Single Mailchimp audience with tags vs. one audience per source — this PR uses the tag approach.
3. Welcome automation in Mailchimp — does Lyle want one set up before this ships, so new subscribers get an immediate "thanks for signing up" message?
4. Member auto-subscribe (signal on `Member.post_save` ACTIVE → Mailchimp with tag `member`) is not in this PR. Schedule it as a follow-up?

## Out of scope (intentional)

- Transactional class emails (confirmation, reminder) stay in Anymail/Resend. We are not moving those to Mailchimp.
- No interest groups / preference centers — tags cover the current need.
- No auto-unsubscribe when a member is deactivated — explicit subscribers stay subscribed unless they unsubscribe themselves via the Mailchimp footer link.


PastLivesReviewBot (approved):
LGTM. Reviewed against the pl-bot-review-pr two-tier checklist; no Tier-1 BLOCK issues, no Tier-2 WARN issues worth inline comments.

PR #95, opened 2026-05-21
Mitigate login-page email abuse (v1.9.1)

HexagonStorms:
## Why

Over the last two days the login page (`/accounts/login/code/`) has been getting hammered with random email submissions, exhausting the Resend daily quota and blocking real members from getting their login codes. Each random email was triggering an "unknown account" notification email; the per-key/per-IP rate limits that allauth already ships with weren't holding because no shared cache backend was configured (every gunicorn worker had its own `LocMemCache`).

## The fix — five layers, ordered cheapest first

1. **Shared cache backend.** Configure Django's `DatabaseCache` (Postgres-backed) plus a migration that creates the cache table. Allauth's existing rate limits now actually hold across workers. Tests still use `LocMemCache` in-process.
2. **Stop emailing random addresses.** Set `ACCOUNT_EMAIL_UNKNOWN_ACCOUNTS = False`. This is the single biggest win — the spam vector. Form UX is unchanged because allauth always returns the "we sent you a code" message regardless of whether the address exists (the standard enumeration-prevention pattern).
3. **Tighter rate limits.** Override `ACCOUNT_RATE_LIMITS["request_login_code"]` from the default `20/m/ip,3/m/key` to `5/m/ip,3/h/key`. Five requests per minute per IP is more than enough for any legitimate user.
4. **Honeypot field.** Hidden `website` input on the login form. Real browsers leave it blank, bots tend to fill every field. If it comes back filled, the form raises the same validation error allauth uses for rate-limited submissions so the bot can't tell it's been honeypotted.
5. **Global circuit breaker.** `core.abuse_limits` keeps a rolling hourly + daily count of login emails sent. Hard caps (defaults: 100/hour, 500/day) keep one bad day from blowing the Resend quota even if every other layer fails. Configurable via `LOGIN_CODE_HOURLY_LIMIT` and `LOGIN_CODE_DAILY_LIMIT` env vars.

## What the attacker sees afterwards

- A bot dumping random emails: form pretends to succeed, zero outgoing emails (layer 2).
- A bot hammering the same fake email: blocked after 3/hour (layer 1 + 3, persisting across workers now).
- A naive bot filling every field: same "too many login attempts" response that real throttling would give (layer 4).
- Any distributed attack that somehow gets past all of the above: capped at the global hourly/daily limit (layer 5), with a logged error.

## What a real user sees afterwards

Identical to before. Submit email → see "we sent you a login code." If the address is on file, the code arrives. If not, no email goes out, but the user sees the same success message and a member-facing follow-up ("Don't know which email is on file?") is right there on the page.

## Deploy notes

- The cache-table migration (`core/0012_create_cache_table.py`) runs `createcachetable` via `RunPython` when the backend is `DatabaseCache`. It is a no-op in tests (which use `LocMemCache`). Render runs `python manage.py migrate --run-syncdb` on build, so this picks up automatically.
- No new env vars are required. `LOGIN_CODE_HOURLY_LIMIT` and `LOGIN_CODE_DAILY_LIMIT` are optional overrides; if you want to tune later, set them in Render's dashboard without a redeploy (gunicorn reloads on env changes? if not, restart).
- `RESEND_API_KEY` and the existing trusted-proxy setup are unchanged.

## Test plan

- [x] `pytest tests/core/abuse_limits_spec.py tests/plfog/adapters_spec.py` — 63 passing
- [x] Full suite — 1715 passing, coverage unchanged from baseline (97.97%)
- [x] `ruff check` + `ruff format --check` clean on all changed files
- [ ] After deploy: watch Render logs for `Login-code circuit breaker tripped` and `Login honeypot triggered` entries to confirm the safety nets are wired up; expected count is 0 under normal use
- [ ] After deploy: confirm Resend dashboard daily send count drops back to a baseline that matches real member activity
- [ ] Verify a real member can still request and receive a login code

## Follow-ups (not in this PR)

If the abuse continues despite all five layers (e.g., the attacker starts using real member emails to provoke 3/hour real sends), the next move is Cloudflare Turnstile on the login form. That would add an account, frontend JS, and a server-side verification step, so leaving it for a follow-up.


PastLivesReviewBot (approved):

PR #97, opened 2026-05-25
v2.1.0: Polished HTML emails for CMS

HexagonStorms:
## Summary
- Class confirmation and reminder emails now use styled HTML matching the login-code design (dark card, gold accents, branded footer) instead of plain text
- Billing receipt emails also upgraded to the same HTML design
- All self-serve links are now absolute URLs (`https://book.pastlives.space/classes/my/...`) so they're clickable in every email client
- Added `BOOK_BASE_URL` setting for building absolute links in outbound emails

## Test plan
- [x] All 2050 tests pass (99.21% coverage)
- [x] mypy clean
- [x] ruff clean
- [ ] Register for a free class on book.pastlives.space and verify the confirmation email renders as styled HTML with a clickable "Manage Registration" button
- [ ] Verify the plain-text fallback still reads correctly (test with a text-only email client or view source)
- [ ] Verify cancellation works from the self-serve link in the email


classes/views.py
```diff
@@ -577,19 +587,27 @@ def instructor_class_submit(request: HttpRequest, pk: int) -> HttpResponse:
...
     )
+    class_groups = []
+    for offering in offerings:
+        regs = (
+            Registration.objects.filter(class_offering=offering)
+            .select_related("member")
+            .prefetch_related("custom_answers__question")
+            .order_by("-registered_at")
+        )
+        class_groups.append({"offering": offering, "registrations": list(regs)})
```
PastLivesReviewBot, line 603: **WARN (non-blocking):** This loop issues one query per offering to fetch registrations. For instructors with many classes this becomes N+1-ish. A single query with `Prefetch('registrations', ...)` on the outer queryset would consolidate to 2 queries total. Fine for now given instructor class counts are typically small, but worth revisiting if the page gets slow.

PastLivesReviewBot (approved):
LGTM — clean feature PR. One minor note inline (WARN, non-blocking).

PR #100, opened 2026-05-26
v2.2.2: Image uploads work on the New Class form

HexagonStorms:
## Summary
- Hero image on `/classes/admin/new/` now shows the same styled drag-and-drop upload zone with instant preview as the edit form (was a bare file input)
- Gallery images on `/classes/admin/new/` now have a multi-file upload with preview thumbnails (was just text saying "Save the class first")
- Gallery files submit with the form and are saved as ClassImage rows on create
- Styles moved out of the conditional block so both create and edit forms share them

## Test plan
- [ ] Go to `/classes/admin/new/` — hero image should show styled upload zone, not a plain file input
- [ ] Drag or click to select a hero image — instant preview should appear
- [ ] Gallery section should show "+ Add images" upload zone instead of "Save the class first" text
- [ ] Select multiple gallery images — thumbnails should preview with filenames and remove buttons
- [ ] Remove a gallery preview image before saving — it should disappear
- [ ] Fill out form and Save — class should be created with hero + gallery images
- [ ] Go to `/classes/admin/4/edit/` — existing AJAX upload behavior should be unchanged
- [ ] Instructor create form at `/classes/instructor/new/` — same improvements should apply

### DIFF
--- a/classes/spec/views/admin_classes_spec.py
+++ b/classes/spec/views/admin_classes_spec.py
@@ -2,7 +2,15 @@
 
 from __future__ import annotations
 
+from io import BytesIO
+
+from django.core.files.uploadedfile import SimpleUploadedFile
 from django.urls import reverse
+
+
+def _image_file(name: str = "shot.png") -> SimpleUploadedFile:
+    buf = BytesIO(b"\x89PNG\r\n\x1a\n" + b"\x00" * 64)
+    return SimpleUploadedFile(name, buf.getvalue(), content_type="image/png")
 
 
 def describe_status_filter():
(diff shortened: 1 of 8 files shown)


classes/spec/views/admin_classes_spec.py
```diff
@@ -2,9 +2,17 @@
...
 from __future__ import annotations
 
+from io import BytesIO
+
+from django.core.files.uploadedfile import SimpleUploadedFile
 from django.urls import reverse
 
 
+def _image_file(name: str = "shot.png") -> SimpleUploadedFile:
+    buf = BytesIO(b"\x89PNG\r\n\x1a\n" + b"\x00" * 64)
```
PastLivesReviewBot, line 12: **WARN — duplicated helper.** `_image_file()` is now copy-pasted across three spec files (`class_offering_spec.py`, `admin_classes_spec.py`, `instructor_dashboard_spec.py`). Could live in `classes/spec/conftest.py` as a shared fixture. Not blocking — just a DRY opportunity for next time.

PastLivesReviewBot (approved):
All clear. Three issues from the first pass are resolved: gallery creation logic moved to `ClassOffering.add_gallery_images()`, `file.name` escaped via `textContent`, and specs added for both create views plus the model method. One minor WARN noted inline.

PR #103, opened 2026-06-08
Release: v2.4.0 — legacy CMS import + automatic daily sync refactor

HexagonStorms:
## v2.4.0 release branch

Integration branch for the **2.4.0** release. It lands two shipped pieces of work plus the design groundwork for the rest of 2.4.0.

### Legacy CMS import
- Periodic sync from `classes.pastlives.space` (Drupal JSON:API) into `ClassOffering` — hybrid pattern: always syncs title/description/price/dates, never overwrites admin-set slug, instructor, or uploaded image
- Image proxy view at `/_legacy-image/` with SSRF protection and 24 h cache; template falls back to it when no local image is present
- Legacy CMS tab in Site Settings: sync toggle, last-synced timestamp, Sync Now button, instructor match status table
- Members admin: class-count column linking to the filtered classes list; Classes admin: instructor dropdown filter
- `sync_legacy_cms` and `download_legacy_images` management commands for manual ops

### Automatic daily sync (new)
Replaces the old page-load sync trigger with a single scheduled refresh:
- `sync_all_sources()` does a full sweep — guild iCal feeds, general feeds, the Drupal classes feed, the legacy CMS, and local classes — with per-source error isolation
- New `sync_all_sources` management command; fails the run (non-zero exit) if any source errors so it surfaces in Render logs
- New `daily-sync` Render cron at `0 13 * * *` (~6 AM Portland) in `render.yaml`
- Removed `refresh_stale_sources()` and the Community Calendar HTMX trigger — pages now read `CalendarEvent` / `ClassOffering` straight from the DB; the sync runs once each morning
- Site Settings "Sync Now" buttons remain for on-demand pulls

> **Deploy note:** the new `daily-sync` cron needs the same env vars as the web service (`DATABASE_URL`, `DJANGO_SECRET_KEY`, `EMAIL_*`, `WEBPUSH_*`, `SENTRY_DSN`) set on it in the Render dashboard, exactly like `airtable-pull`.

## Groundwork for the rest of 2.4.0
Design spec and implementation plans for the remaining 2.4.0 features (notifications, redesigned guild pages, site-wide activity dashboard) are included under `docs/superpowers/`. Implementation to follow on this branch.

## Test plan
- [ ] Site Settings → Legacy CMS → Sync Now: classes appear; class cards show legacy images via the proxy
- [ ] Edit a class's slug + instructor, re-sync — neither is overwritten; remove a class from the CMS, re-sync — it's archived
- [ ] Run `python manage.py sync_all_sources` — calendar + legacy classes refresh into the DB; a broken feed fails the run loudly
- [ ] Confirm the Community Calendar renders from the DB with no page-load fetch
- [x] Full suite green: 2264 passed, 98.13% coverage, ruff clean


classes/import_service.py
```diff
@@ -0,0 +1,195 @@
...
+    next_url: str | None = LEGACY_CMS_API_URL
+    while next_url:
+        data = _fetch_json(next_url)
+
+        for item in data.get("data") or []:
+            node_id = _upsert_offering(item)
+            if node_id:
+                seen_ids.append(node_id)
+
+        next_url = ((data.get("links") or {}).get("next") or {}).get("href")
```
PastLivesReviewBot, line 184: **WARN (non-blocking):** This pagination loop follows `links.next.href` straight from the upstream API response with no host check, then `_fetch_json` opens it directly. The image proxy in `views_legacy_image.py` is careful to refuse redirects and pin the host against SSRF; this loop trusts whatever `next` URL the upstream returns. It's a first-party, server-side scheduled sync so the practical risk is low — but for defense-in-depth consistency with the proxy, consider asserting `next_url.startswith(LEGACY_CMS_BASE)` before fetching. Approving as-is.

PastLivesReviewBot (approved):
**PastLivesReviewBot — automated review ✅ Approved**

Reviewed the full 2.4.0 diff with focus on deploy-correctness. No blocking issues found.

**Migrations** — the high-risk `0031_instructor_to_member` data migration is well-built: uses `apps.get_model()` (no direct imports), touches only real DB columns (no historical-proxy method misuse), uses base `save(update_fields=...)`, runs `atomic = False` with a documented reason, and is deliberately + clearly documented as irreversible (raises `IrreversibleError` rather than a silent `RunPython.noop`). `makemigrations --check` reports no drift.

**Runtime** — `Instructor` model fully removed with no dangling references; all `.instructor.*` access now goes through `Member` (verified `display_name`/`primary_email` exist as properties). Relay URLs and the legacy-image proxy are wired correctly. `Member` is imported under `TYPE_CHECKING` for annotations.

**Security** — image proxy SSRF guard is solid (`startswith` with trailing slash defeats `…space.evil.com` / `…space@evil.com`; redirects refused). Cross-surface relay uses 30s HMAC-signed tokens, validates `book_host` against `PUBLIC_HOSTS`, and rejects `@`/`/` smuggling.

**Standards** — `help_text` on every field, `TextChoices`, meaningful `__str__`, `default=dict`, partial `UniqueConstraint`, full type annotations. Business logic lives in models (`submit_for_review`/`approve`/`decide`/`on_review_decision_recorded`) and validation in forms (`ClassReviewDecisionForm.clean`). `admin_activity` uses `select_related` to avoid N+1. Ruff clean. Version bumped 2.2.2 → 2.4.0 with changelog.

One non-blocking WARN left inline — it does not hold up the merge.

PR #104, opened 2026-06-09
v2.5.0 — Activity audit dashboard, booking-site fixes, instructor→teach rename

HexagonStorms:
## v2.5.0

Three workstreams. All green: **2327 tests, 98.28% coverage, ruff + mypy clean**.

### 1. Audit foundation — site-wide Activity feed + email logging
- New `TransactionalEmailLog` and `SiteActivity` models (migrations `0021`, `0022`).
- `core/email.send()` — a single logged choke point; **every** transactional email in billing/core/classes routes through it and is recorded whether it sends or fails.
- Instrumentation across auth (login/logout/signup), profile updates, guild votes, billing (charges, failures, tab entries), funding snapshots, class lifecycle, and invites.
- Staff dashboard at **`/manage/activity/`** — **Activity Feed** + **Email Log** tabs (filters + pagination). Linked from the hub admin sidebar (admin-only, above "Classes & Workshops") and the django-unfold admin sidebar.

### 2. Booking-site (book.pastlives.space) fixes
- **"Manage Classes & Workshops" / "Manage My Classes" buttons** now work from the public surface — they cross to the members host via a new `MEMBER_BASE_URL` setting. (The relative links 404'd on `book.*` after the v2.0.0 surface split — a regression.)
- The booking topbar's **FOG↗ / Member / Log out** chrome is now styled (those CSS classes were never defined), and the FOG↗/Member cluster shows only on the public surface.
- Removed the redundant **Browse Classes** hero button.

### 3. "Instructor role" → "teach" rename
- There is no separate instructor role — any active member can host a class. Renamed the self-service portal: `/classes/instructor/*` → `/classes/teach/*`, plus `teaching_member_required`, views, forms, templates, and tests. **301 redirects** from the old paths keep existing email deep-links/bookmarks working.
- Kept "instructor" as the per-class teacher noun: the `ClassOffering.instructor` FK, the public `/classes/instructors/<slug>/` profiles, the "N Instructors" stat, and the instructor notification emails.

### Testing notes
- The Activity feed / email log start empty and fill as **new** events happen (no historical backfill) — log out/in, edit a profile, vote, add a tab entry, or register for a class to see rows appear.
- The "Manage Classes & Workshops" button only renders for an admin (view-as Admin); the FOG↗/Member chrome only for an active member account.

### Release
Member-friendly **2.5.0** changelog entry added to `plfog/version.py` (posts to the Discord channel on merge to `main`).


core/notifications.py
```diff
@@ -0,0 +1,64 @@
...
+            send_email(
+                to=user.email,
+                subject=title,
+                trigger_kind=f"notification.{trigger_key}",
+                text_body=body,
+                best_effort=True,
+            )
+
+
+def active_member_users():
```
PastLivesReviewBot: **[BLOCK]** Missing return type annotation. Per project standards, all functions must have full type annotations including return types. This returns a `QuerySet[User]`:

```python
def active_member_users() -> "QuerySet[User]":
```

(Add `from django.db.models import QuerySet` and `from django.contrib.auth.models import User` to `TYPE_CHECKING` imports as needed.)

billing/fields.py
```diff
@@ -47,7 +47,13 @@ class EncryptedCharField(models.CharField):
     def from_db_value(self, value: Any, expression: Any, connection: Any) -> str | None:
         if value is None or value == "":
             return value
-        return _fernet().decrypt(value.encode()).decode()
+        try:
+            return _fernet().decrypt(value.encode()).decode()
+        except (InvalidToken, Exception):
```
PastLivesReviewBot: **[WARN]** `except (InvalidToken, Exception)` is redundant — `Exception` already subsumes `InvalidToken`, so the tuple is equivalent to a bare `except Exception`. If the intent is to catch only Fernet decryption failures narrowly, use `except InvalidToken`. If the intent is the documented "do not crash the entire site" broad catch, use `except Exception` explicitly and drop the redundant `InvalidToken`.

PastLivesReviewBot (changes requested):
One BLOCK: missing return type annotation on `active_member_users()`. One WARN included as inline comments.

billing/fields.py
joshplaza: **BLOCK — broad `except` swallows all exceptions silently.**

`except (InvalidToken, Exception):` is equivalent to `except Exception:` because `Exception` already subsumes `InvalidToken`. Any programming error inside `_fernet()` — a misconfigured key format, an `AttributeError`, anything — will be silently swallowed and returned as `""`, making the field look legitimately empty when it's actually broken.

The fix is already in the local working tree (`except InvalidToken:` only). Just commit and push it:

```python
except InvalidToken:
    # If we can't decrypt (usually due to a changed STRIPE_FIELD_ENCRYPTION_KEY),
    # return an empty string instead of crashing the entire site.
    return ""
```

core/notifications.py
joshplaza: **BLOCK — missing return type annotation.**

`def active_member_users():` has no return type annotation. Per project standards, all functions must have full type annotations including the return type.

The fix is already in the local working tree (`-> "QuerySet[User]"` + the `TYPE_CHECKING` imports). Just commit and push it:

```python
def active_member_users() -> "QuerySet[User]":
```

joshplaza (changes requested):
Two blocking issues found — both are already patched in the local working tree as uncommitted changes, so they just need to be committed and pushed before merge.

---

**Workstreams reviewed:** Activity audit models/views, email logging choke point, booking-site fixes, instructor→teach rename (with 301 redirects), guild page redesign (images, FAQ, links, announcements, roster), notification system (in-app + push + email), hero crop mixin, class grouping/catalog deduplication, admin/teach overview dashboards. Test suite coverage: 2327 tests.

**Overall assessment:** Solid, well-structured PR. The two BLOCKs below are minor correctness issues — one is a broad exception catch that silences bugs, one is a missing return type annotation — both already fixed locally.

joshplaza (approved):
Both blocking issues resolved in the follow-up commit. LGTM.

PastLivesReviewBot (approved):
Both blocking issues resolved in the follow-up commit. LGTM.

PR #106, opened 2026-06-11
Release 2.5.6

HexagonStorms:


PastLivesReviewBot (approved):
Clean release. No block issues found.

**Reviewed:**
- `teach_profile` view simplified to a redirect — `reverse` confirmed imported, `hub_user_settings` URL confirmed wired
- `instructor_website` / `instructor_social_handle` confirmed as real model fields; `member.is_instructor` confirmed as property
- `run_scheduled_tasks` dispatcher: correct error isolation, proper time-gating, and the patch target in specs matches the import path
- CSS token extraction in `book-account.css`: light defaults reference hub design tokens; dark overrides restore original navy values
- Onboarding JS: radio deselect-all-then-select and checkbox toggle logic are correct
- render.yaml consolidation: three crons → one 15-minute dispatcher with internal time-gating

PR #109, opened 2026-06-23
Release 0.19.x — the big update (guilds, notifications, voting, Discord, classes) + go-live

HexagonStorms:
## Release 0.19.x — the big update (go-live)

Production runs **0.18.11**; this branch is the entire **0.19 line**, shipping as one release. The member-facing changelog is curated into **9 grouped features** (`plfog/version.py`, `0.19.9`→`0.19.1`); the Discord workflow aggregates them into one announcement on merge to `main`.

### What members get (the 9 features)
1. **Show off your skills — and open up for commissions** — skills on your directory profile, "open for commissions," and a searchable/filterable directory.
2. **Your notifications and emails, all in one place** — one settings grid (bell / email / push), auto-sent class reminders (no dupes), one branded email layout with unsubscribe, smarter "new login" alert.
3. **The monthly guild-funding vote, leveled up** — tabbed voting admin, member voting emails (closing-soon / nudge / results), a Voting → Settings page that holds the results email for organizer review.
4. **Guild pages get a leadership team and more to share** — co-leads/secretaries/treasurers/orienters, meeting notes + richer FAQ media, section-by-section editing + a gallery.
5. **Community events and meetings on the calendar** — guild & community events with recurrence, on the Community Calendar with a Discord heads-up.
6. **Easier member management and invites** — every member has a login; one-tap invites with an outstanding/resend/revoke workflow; tidier member admin.
7. **Discord, tidied** — one announcements channel, per-guild channels, and a personal DM line.
8. **Book a class, get an account** — passwordless account on booking; smarter newsletter list routing.
9. **A few catalog touches** — bookable-session counts, "Guilds" naming everywhere.

### Go-live operations (see `docs/release-0.19-go-live-runbook.md`)
- **Build (automatic):** `migrate` + `seed_notification_templates` run on the Render build.
- **Manual, once:** `provision_member_users` (mints login accounts).
- **Release member-email is a MANUAL admin send** — from **Site Settings → Announcements** ("Draft from latest release" → preview → Send). It reaches **only members who have logged in** (activated). Leave "Also post to Discord" **off** — the GitHub Action posts the release to Discord on merge to `main`.
[...]

PR #110, opened 2026-06-28
Fix prod deploy: drop dangling Quill sourcemap refs (v0.19.11)

HexagonStorms:
## Why

The 0.19.x cutover deploy (PR #109, `9e7865f`) **failed at build** during `collectstatic`. Production is still on the prior release (2.5.9). The build is using the **Dockerfile**, and WhiteNoise's `CompressedManifestStaticFilesStorage` resolves every reference at collect time — it raised `MissingFileError` because the vendored Quill assets point at sourcemaps that aren't in the repo:

- `static/css/quill.snow.css` → `quill.snow.css.map`
- `static/js/quill.min.js` → `quill.js.map`

## What

- **Strip the dangling `sourceMappingURL` lines** from both Quill files. Verified locally by replaying the build's exact `collectstatic` in a clean container: **passes, 582 post-processed, no error.**
- **`Dockerfile` start command now runs `seed_notification_templates --quiet`.** The live service builds from the Dockerfile, not `render.yaml`, so the go-live runbook's "seed is wired into the build" assumption never held. Now it's automatic on boot, alongside the existing `migrate --noinput`.
- **`VERSION` 0.19.10 → 0.19.11**; re-stamped the top changelog entry (no new member-facing bullet — this is an invisible intra-cycle fix to unshipped 0.19.x work).

## Post-merge (manual, after the deploy goes live)

- `python manage.py provision_member_users --dry-run`, then for real (mints login accounts — deliberately manual).
- Send the release announcement from Site Settings → Announcements.
- Optional env: set `DISCORD_*` (Discord posting/DM/OAuth) and `WEBPUSH_VAPID_ADMIN_EMAIL` (browser push) in Render when ready.


HexagonStorms:
Wrong branch shape — release-0.19.x is squash-merged into main, so this diffs the whole 0.19.x changeset and conflicts. Reopening as a minimal hotfix branch off main.

PR #111, opened 2026-06-28
Fix prod deploy: drop dangling Quill sourcemap refs (v0.19.11)

HexagonStorms:
## Why

The 0.19.x cutover deploy (PR #109, `9e7865f`) **failed at build** during `collectstatic`. Production is still on the prior release (2.5.9). The build is using the **Dockerfile**, and WhiteNoise's `CompressedManifestStaticFilesStorage` resolves every reference at collect time — it raised `MissingFileError` because the vendored Quill assets point at sourcemaps that aren't in the repo:

- `static/css/quill.snow.css` → `quill.snow.css.map`
- `static/js/quill.min.js` → `quill.js.map`

## What

- **Strip the dangling `sourceMappingURL` lines** from both Quill files. Verified locally by replaying the build's exact `collectstatic` in a clean container: **passes, 582 post-processed, no error.**
- **`Dockerfile` start command now runs `seed_notification_templates --quiet`.** The live service builds from the Dockerfile, not `render.yaml`, so the go-live runbook's "seed is wired into the build" assumption never held. Now it's automatic on boot, alongside the existing `migrate --noinput`.
- **`VERSION` 0.19.10 → 0.19.11**; re-stamped the top changelog entry (no new member-facing bullet — this is an invisible intra-cycle fix to unshipped 0.19.x work).

## Post-merge (manual, after the deploy goes live)

- `python manage.py provision_member_users --dry-run`, then for real (mints login accounts — deliberately manual).
- Send the release announcement from Site Settings → Announcements.
- Optional env: set `DISCORD_*` (Discord posting/DM/OAuth) and `WEBPUSH_VAPID_ADMIN_EMAIL` (browser push) in Render when ready.


PastLivesReviewBot (approved):
Approved — clean prod-deploy hotfix.

- **quill.snow.css / quill.min.js**: removes only the dangling `sourceMappingURL` refs to `.map` files that aren't vendored; this is what broke WhiteNoise `collectstatic` on the Render build. Replayed the build's `collectstatic` in a clean container — passes (582 post-processed).
- **Dockerfile**: `seed_notification_templates --quiet` added to the start `CMD`; command exists and supports the flag, runs after `migrate`.
- **version.py**: 0.19.10 → 0.19.11, changelog entry re-stamped so `announce_release` still resolves.

No logic/migration/model/view changes. No BLOCK or WARN findings.

PR #112, opened 2026-06-28
Fix Discord release post: chunk under 4096 + fail loudly (v0.19.12)

HexagonStorms:
[ci] Fix Discord release post: chunk under 4096 + fail loudly (v0.19.12)

The 0.19 release never posted to Discord even though the workflow ran green.
Root cause: the aggregated release-line changelog is ~5170 chars, over Discord's
4096-char embed-description limit, so the webhook rejected the embed with a 400 —
but the inline `curl` had no `--fail`, so the step exited 0 and the failure was
invisible.

- Replace the inline curl/jq with `.github/scripts/discord_release_notify.py`,
  which aggregates the current release line, splits the notes across as many
  embeds/messages as needed (each <= 4096), and exits non-zero on any rejected
  post so a future failure is loud, not silent.
- Add `workflow_dispatch` so a missed release post can be re-triggered manually
  (gh workflow run discord-notify.yml) without forcing a redeploy.
- Bump VERSION 0.19.11 -> 0.19.12; re-stamp the top changelog entry.

The release webhook itself (the DISCORD_WEBHOOK_URL repo secret) was never the
problem and is unchanged.

### DIFF
--- a/.github/scripts/discord_release_notify.py
+++ b/.github/scripts/discord_release_notify.py
@@ -1,0 +1,101 @@
+"""Post the current release's changelog to Discord — chunked under the embed limit.
+
+Aggregates every CHANGELOG entry in the current release line (MAJOR.MINOR) into one
+announcement, newest first, and posts it to the ``DISCORD_WEBHOOK_URL`` webhook. The
+release line can accumulate several patch entries (e.g. 0.19.1 … 0.19.11), so the
+combined notes routinely exceed Discord's 4096-char embed-description limit — this
+splits them across as many embeds/messages as needed and FAILS LOUDLY on a rejected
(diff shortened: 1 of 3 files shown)


PastLivesReviewBot (approved):
Approved — CI/infra fix for the Discord release post.

- Replaces the inline curl (no `--fail`, silently swallowed Discord's 400) with `discord_release_notify.py`, which chunks the release-line notes under the 4096-char embed limit (verified: 0.19 splits into 2 messages of 3905 + 1264) and exits non-zero on any rejected post.
- Adds `workflow_dispatch` for manual re-trigger.
- No app-code change; reads only the trusted DISCORD_WEBHOOK_URL secret + repo version.py (no GitHub event input). VERSION bumped 0.19.11 → 0.19.12.

No BLOCK or WARN findings.

PR #113, opened 2026-06-28
Discord post: send a User-Agent so Cloudflare doesn't 403 (v0.19.13)

HexagonStorms:
Follow-up to #112. The chunking fix reached Discord but Cloudflare blocked the default `Python-urllib` User-Agent (HTTP 403 / error 1010), so 0.19 still didn't post. Adds a descriptive `User-Agent` header.

**Verified locally:** with the UA, a POST to a known-bad webhook returns `404 / 10015 Unknown Webhook` instead of `403 / 1010` — the request now reaches Discord's API. The release webhook secret itself is valid. Merging this re-runs the workflow and should finally post the 0.19 notes.

### DIFF
--- a/.github/scripts/discord_release_notify.py
+++ b/.github/scripts/discord_release_notify.py
@@ -61,7 +61,12 @@ def _post(webhook: str, payload: dict[str, object]) -> None:
     request = urllib.request.Request(
         webhook,
         data=json.dumps(payload).encode("utf-8"),
-        headers={"Content-Type": "application/json"},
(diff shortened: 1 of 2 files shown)


PastLivesReviewBot (approved):
Approved (user-authorized merge) — one-line follow-up to #112: adds a descriptive User-Agent so Cloudflare stops 403-ing the release post. Verified locally (bad webhook now returns 404/10015, not 403/1010). CI/infra only.

PR #116, opened 2026-06-28
Headless send_site_announcement command (v0.19.16)

HexagonStorms:
CLI twin of the Site Settings → Announcements composer — sends the rich-HTML `site_announcement` email to activated members from a one-off job, **no admin session** (same `noreply@` Resend sender). This lets the release announcement go out headlessly instead of driving the admin UI through a login.

- Subject/body base64-encoded (survives job-runner whitespace splitting).
- Body is sanitized + inline-styled + wrapped in the branded shell and handed to `emit` as a per-channel EMAIL `Message`, so the HTML is **not autoescaped** (the spine escapes copy-merge values; a pre-rendered `html_body` is trusted) — mirrors `hub.views._send_site_announcement`.
- Discord off by default (`--discord` to also post).
- 5 specs (rich HTML unescaped, discord suppress default/flag, bad base64, empty-after-sanitize). Local: ruff + mypy clean, all green.

### DIFF
--- a/core/management/commands/send_site_announcement.py
+++ b/core/management/commands/send_site_announcement.py
@@ -1,0 +1,82 @@
+"""Send a site-wide announcement email headlessly — the CLI twin of the Site Settings
+→ Announcements composer (``hub.views._send_site_announcement``).
+
+The web composer needs an admin session; this command sends the same
+``site_announcement`` event from a one-off job (e.g. a Render job) with no login, from
+the same ``noreply@`` Resend sender. It builds the branded rich-HTML email exactly like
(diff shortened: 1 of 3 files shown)


PastLivesReviewBot (approved):
Approved (user-authorized) — headless send_site_announcement command, the CLI twin of the Site Settings composer. Sends rich HTML via a pre-rendered per-channel Message (not autoescaped), subject/body base64-encoded, Discord off by default. ALL_ACTIVE_MEMBERS audience. ruff + mypy clean; 5 specs green (incl. HTML-unescaped + bad-base64 + empty-after-sanitize).

PR #119, opened 2026-07-04
v0.20 + v0.21 — combined member release notes + features

HexagonStorms:
The **v0.20 + v0.21** feature batch — rebased cleanly onto current `main`, linear history, additive diff. v0.20 never got a public release email, so the announcement now spans both lines (`send_release_email --lines 0.20,0.21`).

## What members get (the "what's new" email)

Curated + grouped by topic across both lines, in plain member-facing language (10 cards, 6 with screenshots):

- **Your Guilds, wherever you are** — an on/off Guild switch in Settings, and picking guilds from `#choose-your-guild` on Discord; syncs both ways.
- **A better way to run announcements** — members can suggest guild announcements, and the tool got an upgrade (`@everyone`, live email preview, drafts, channel picker).
- **Community events, leveled up** — propose events; every event gets a shareable page with "Add to calendar"; scheduled announcements + 1/3/7-day reminders.
- **Your notifications, cleaned up** — a full, mobile-friendly Notifications page, and the false "new sign-in" email is gone.
- **QR codes, QR codes, QR codes!** — guilds, classes, and events all generate a ready-to-print QR share card.
- **A home base when you sign in** — a real home dashboard, plus a "Get started" checklist for new members.
- **One place for how our space works** — a Space & Org Info page (map, parking, contacts, code of conduct).
- **Your profile & the member directory** — safer profile edits, listed by default, and Discord-handle autofill.
- **Guild pages got a glow-up** — editor tabs, announcement email + Discord channel choice, staff badges, richer FAQs.
- **A few smaller touches** — cleaner sign-in, a tidier calendar, a theme that follows you, and refreshed update emails.

_"Screens around the space" (kiosk signage) is intentionally left out of this member announcement; it's still in the code + git history and can be added later._

## Technical scope

- Google Calendar event sync (service-account auth, behind `google_calendar_sync_enabled`); canonical `pastlives.app` → member-host redirects in `SurfaceMiddleware`.
- Signage surface (standalone doctype root) + `SlideshowZone`/`Slide`; unified announcement wizard + `AnnouncementDraft`; event pages/QR; two-way Discord↔app guild sync (`GuildMembership.source` provenance, reaction reconcile cron, role sync).
- Release-email tooling: `send_release_email` (all activated members, `--confirm`-gated) + `send_release_email_test --to <email> --lines 0.20,0.21`; feature screenshots render from R2. Changelog curated to topic-grouped, member-friendly notes (every bullet ≤160 chars).
- Migrations linearized onto main's 0.20.5 leaves.

## Testing

- Full suite green on the SQLite CI mirror; `ruff`/`ruff format`/`mypy` clean; `manage.py check --database default` clean (E034 guard).

## Go-live ops (not blocking review; tracked separately)

Google Calendar service account + shared calendars + Calendar IDs; `seed_notification_templates` + `provision_member_users` post-deploy; Discord second OAuth redirect URI + webhooks/config on prod; verify `DISCORD`/`WEBPUSH`/`R2` env in Render.


core/integrations/google_calendar.py
```diff
@@ -0,0 +1,233 @@
...
+        """Delete an event. No-op-safe from the caller's view (wraps API errors)."""
+        self._execute(self._service.events().delete(calendarId=calendar_id, eventId=event_id))
+
+    @staticmethod
+    def _execute(request: Any) -> Any:
+        """Run a built request, translating ``googleapiclient`` ``HttpError`` into our own
+        :class:`GoogleCalendarError` so callers catch one exception type."""
+        try:
+            return request.execute()
+        except HttpError as exc:
```
PastLivesReviewBot: **BLOCK — this `except HttpError` is too narrow; it breaks the module's own "never raise to the caller" contract and will 500 the event request path on a normal Google auth/transport failure.**

`_execute` only translates `googleapiclient.errors.HttpError` (an HTTP *error-status response* from the API). But the service-account token is minted lazily over the network on the **first** `request.execute()` — and because `push_community_event` rebuilds the client via `from_settings()` on *every* call, that token exchange with `oauth2.googleapis.com` happens on **100% of pushes/deletes**. The common failure modes there — `google.auth.exceptions.RefreshError` / `TransportError` (unreachable token endpoint, clock-skew `invalid_grant`, revoked key) and socket-level `TimeoutError`/`ConnectionError`/`OSError` — are a *separate exception hierarchy* from `HttpError`, so they are **not** caught here and **not** wrapped in `GoogleCalendarError`.

They then escape the `except GoogleCalendarError` in `push_community_event` (:218) and `remove_community_event` (:232), which is the only guard the callers rely on. `CommunityEvent.push_to_google`/`remove_from_google` have no try/except (docstrings promise "Never raises"), and the view call sites (`hub/views.py:2693, 2728, 2759, 2786, 3015`) call them unwrapped ("best-effort"). Net effect once sync is enabled at go-live:
- a Google auth/transport blip **500s** the event publish / approve / edit / delete request, and
- it aborts the `retry_calendar_pushes` cron mid-loop (`retry_calendar_pushes.py:34-36` is not per-row isolated) — killing every remaining retry in that tick, in the one command built specifically to recover from flaky Google connectivity.

This contradicts the module docstring (lines 5-7: "**never raise to the caller** — a Google outage records a `sync_error` … and the FOG save proceeds regardless") and both model-method docstrings.

Fix is one line, matching the sibling `from_settings` (:83, already `except Exception`):
```python
try:
    return request.execute()
except Exception as exc:  # HttpError + google.auth transport/refresh + socket errors
    raise GoogleCalendarError(str(exc)) from exc
```
(This also makes the `g["id"]` access at :214 safe against a malformed-but-2xx body.)

membership/models.py
```diff
@@ -2323,6 +2716,51 @@ def __str__(self) -> str:
...
+
+
+class DiscordGuildEmoji(models.Model):
+    """One reaction-emoji → guild mapping for the Discord role message (admin-editable).
+
+    Many emojis may point at one guild (collapsed guilds: two Glass emojis → one Glass
+    Guild). An emoji with no row is ignored in both directions.
+    """
+
+    emoji = models.CharField(
```
PastLivesReviewBot, line 2738: **WARN (non-blocking) — duplicate unique index on `emoji`.** This field already declares `unique=True`, and `Meta.constraints` (:603) adds `UniqueConstraint(fields=["emoji"], name="uq_discordguildemoji_emoji")` on the same column — two unique indexes on one field. Harmless but redundant. Per the repo standard favoring `UniqueConstraint`, drop the field-level `unique=True` and keep the named constraint.

membership/qr.py
```diff
@@ -0,0 +1,44 @@
...
+import io
+import re
+
+import segno
+
+# segno's SVG writer stamps a fixed pixel ``width``/``height`` (e.g. ``width="37"``) and
+# NO ``viewBox``, so the QR always draws at its intrinsic ~37px however large its container
+# is — the "tiny QR in a big white box" bug. Swapping that fixed size for an equal-extent
+# ``viewBox`` lets the SVG scale to its box (every QR container styles ``svg { width/height: 100% }``).
+_QR_SIZE_ATTRS = re.compile(r'\swidth="(\d+)"\sheight="(\d+)"')
```
PastLivesReviewBot: **WARN (non-blocking) — regex assumes segno's exact SVG attribute formatting.** This matches `width="N" height="N"` with single spaces; if a future segno build changes the spacing/ordering, `.sub(count=1)` silently returns the original fixed-size SVG (degrades to the tiny-QR look, no crash). Acceptable while `segno>=1.6` is pinned — just fragile-by-format; a note or a looser pattern would harden it.

PastLivesReviewBot (changes requested):
## PastLivesReviewBot — automated review of PR #119 (v0.20 + v0.21 combined release)

**Verdict: Requesting changes — 1 BLOCK, 8 WARN.** A large, well-built release; nearly everything checks out. One deploy-correctness item needs a one-line fix before merge.

### 🔴 BLOCK (1)
- **`core/integrations/google_calendar.py:110`** — `_execute` catches only `HttpError`, so a service-account **auth/transport failure** (`RefreshError`/`TransportError`, socket timeouts) minted on every `.execute()` escapes the `GoogleCalendarError`-only guards, the unwrapped model methods, and the unwrapped view/cron call sites → **500 on event publish/approve/edit/delete** and a full-batch abort of `retry_calendar_pushes`, contradicting the module's own "never raise" contract. Details + fix inline. (Feature is flag-gated off today, but this PR's go-live plan turns it on.)

### 🟡 WARN (non-blocking; inline where they anchor to the diff)
- `membership/models.py:2738` — `DiscordGuildEmoji.emoji` has both `unique=True` and a `UniqueConstraint` (duplicate index). *(inline)*
- `membership/qr.py:18` — QR-resize regex is fragile to segno's exact SVG formatting. *(inline)*
- `core/events/copy.py:86,100` — two new audience-description literals exceed the 120-char guide (E501 isn't gated; matches existing convention).
- `core/integrations/google_calendar.py:92,96` — `insert_event`/`update_event` annotate bare `dict`; `dict[str, Any]` matches the module's own style.
- `core/context_processors.py` — `signage_page_base` set unconditionally (sibling `guilds_page_base` is conditional); harmless.
- `hub/views.py:~2853` (`propose_event`) — event_type derivation + PUBLISHED-state set inline before delegating to `publish()`/`submit_for_review()`; mirrors the existing `event_edit` pattern, so borderline. A `CommunityEvent.propose(by, guild, policy)` would be a cleaner home.
- `hub/views.py:~2061` (`hub_compose_send`) — send-time "body required" re-checked in the view via `add_error`; mitigated because `AnnouncementDraft.send()` independently raises `ValidationError`, so the model stays the source of truth.
- `hub/views.py` `_reviewer_guild_scope` — tri-state `Any` sentinel (`True`/`QuerySet`/`None`) reads a bit opaque; a small enum/dataclass would clarify.

### ✅ Verified clean
- **All 13 migrations** additive/correct; the one `RunPython` (0075) uses `apps.get_model()` + raw field access + an explicit reverse (no bare `RunPython.noop`); `discord_role_ids` uses `default=list`; check/unique constraints named and correct.
- `manage.py check` and `makemigrations --check` both clean; no missing migrations; model state matches.
- **Model standards** across all new models: `help_text`, `TextChoices`, meaningful `__str__`, `default=dict`/`default=list`, `CheckConstraint(condition=…)`.
- **emit() period invariant** — every real call site has a unique period (`event:{pk}:reminder:{days}d`, `discord_import:{pk}:{linked_at}`, timestamped announce/resubmit periods); notification URLs are absolute.
- **Broadcast activation gate** honored — `send_release_email` is `--confirm`-gated and resolves through `ALL_ACTIVE_MEMBERS` (`last_login__isnull=False`), same resolver for the dry-run count and the real send.
- **Discord anti-oscillation** — inbound reaction sync only ever writes/removes `source="discord"` rows, never touches `app` rows.
- URL/view wiring, imports, permission gates, `admin_login` override, N+1 (`select_related`), type annotations, and `ruff` all verified.

_Automated review; the BLOCK is a genuine deploy-correctness gap with a trivial fix. Re-request review after patching `_execute`._

PastLivesReviewBot (approved):
## PastLivesReviewBot — ✅ Approved (v0.20 + v0.21 combined release)

The one BLOCK from my earlier review is resolved, all 8 WARNs are addressed, and an
additional pre-existing Postgres defect surfaced during re-review was fixed. Full suite,
lint, and mypy are green. Approving.

### The original BLOCK — resolved (`08f63cf`)
- **`core/integrations/google_calendar._execute`** now catches `except Exception` (was
  `HttpError`-only), so a service-account auth/transport failure — minted over the network
  on every `execute()` — is wrapped in `GoogleCalendarError` and degrades instead of 500-ing
  event publish/approve/edit/delete or aborting the retry cron. Honors the module's
  "never raise to the caller" contract; unused `HttpError` import removed. Independently
  re-verified: the failure now degrades, and `except Exception` correctly does not swallow
  `KeyboardInterrupt`/`SystemExit`.

### WARNs — all addressed (`08f63cf`)
- DiscordGuildEmoji.emoji: dropped the redundant field-level `unique=True` (kept the named
  `UniqueConstraint`) + migration `0082`.
- `qr.qr_svg`: width/height matched independently (order/whitespace-tolerant), degrades to
  the untouched SVG on unexpected segno output (+ covering spec).
- Precise `dict[str, Any]` annotations; two long copy literals reflowed under 120.
- `signage_page_base` made conditional on `is_signage` (+ updated/added surface specs).
- `CommunityEvent.propose()` model method extracted from the view (fat model, skinny view).
- `AnnouncementComposeForm` enforces send-time "body required" via a `require_body` flag.
- Reviewer tri-state sentinel replaced by a typed `_ReviewScope` dataclass.

### Bonus fix — pre-existing Postgres 500 (`1f94900`)
Re-review on Postgres exposed that the notification `period` column (`EventDelivery.period`,
`varchar(40)`) was too narrow for the timestamp-suffixed periods the emit spine builds:
`GuildAnnouncement` resubmit (`announcement:{pk}:submitted:{ts}`, ~42 chars) raised
`DataError: value too long` on production Postgres — masked by SQLite CI. This PR's new
`event:{pk}:submitted:{ts}` periods shared the pattern. Widened `period` to 120 (additive
`AlterField`, migration `0047`), fixing the pre-existing 500 and hardening the new periods.
Proven on a throwaway Postgres: the 3 previously-failing `guild_announcement_lifecycle`
cases now pass (67 specs green).

### Verification
- Full suite on SQLite (CI parity): **5226 passed**, coverage **98.16%** (gate 98%).
- Affected specs on **Postgres**: 67 passed (incl. the formerly-failing GuildAnnouncement cases).
- `ruff check` / `ruff format --check` and `mypy plfog/ core/ membership/ hub/` clean.
- CI on HEAD (`1f94900`): test + lint + e2e all green.

_Automated review. Migrations `0082` (emoji dedup) and `0047` (period widen) are additive and
safe; both run automatically on deploy._

PR #130, opened 2026-07-18
[discord] /info guild dropdown (v0.23.1)

HexagonStorms:
The `/info` Discord command now offers a **dropdown of guilds** to pick from — the same picker `/join-guild` uses — instead of typing the guild name. Still optional (omit → the current channel's guild).

- New `_info_guild_choices` options_builder in `membership/discord_commands.py` (mirrors `/join-guild`'s `_guild_choices`: choices built at registration time, value = slug, 25-cap, empty-guard). Handler unchanged — `resolve_command_guild` already matches by name/slug.
- `info_command_spec` updated: asserts the dropdown choices + that a slug value resolves.
- VERSION → 0.23.1, one member-facing changelog entry.

Post-merge: re-run `register_discord_commands` so the new `/info` choices register.

### DIFF
--- a/membership/discord_commands.py
+++ b/membership/discord_commands.py
@@ -210,11 +210,36 @@ def _info(interaction: Interaction, member: Member | None) -> dict:
     return reply(f"Here's **{guild.name}** on Past Lives:", ephemeral=False, embeds=[embed])
 
 
+def _info_guild_choices() -> list[dict]:
+    """The ``guild`` option for ``/info`` — a dropdown of active guilds, like ``/join-guild``.
+
+    Mirrors that command's picker (each choice's value is the guild ``slug``, resolved by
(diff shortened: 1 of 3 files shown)


PastLivesReviewBot (approved):
Reviewed the diff. Small, self-contained change: `/info` gains a guild dropdown via a new `_info_guild_choices` options_builder (mirrors `/join-guild`'s picker — choices built at registration time, value = slug, 25-cap, empty-guard), staying `required=False` so omitting it uses the channel's guild. Handler is unchanged — `resolve_command_guild` already matches by name/slug. Spec updated to assert the dropdown choices + slug resolution; VERSION → 0.23.1 with a member-facing changelog entry. No BLOCK issues. **Approving.**

PR #132, opened 2026-07-21
[classes] Guild tagging: sync keeps curated categories + bulk suggester (v0.23.3)

HexagonStorms:
## Why

Every legacy-CMS sync (nightly + the manual "Sync listings" button) re-filed all ~600 class offerings back into the generic Class/Workshop/Open Studio buckets, wiping any guild categorization — Drupal's `field_class_type` only knows those three values. Members couldn't browse classes by guild, and staff curation never survived a day.

## What

**1. Import fix (`classes/import_service.py`)** — `category` moves from `update_or_create` defaults to `create_defaults`: set on first import, never clobbered after. Matches the existing don't-clobber pattern for slug/instructor/image.

**2. Guild-tagging suggester** — a bulk cleanup surface at `/classes/admin/categories/guild-tagging/` ("Guild tagging" button on the Guild Types tab):
- `ClassOffering.suggest_guild_category()` + ordered `GUILD_CATEGORY_KEYWORDS` (classes/models.py) — keyword scan of title+description, most-specific-first so "Stained Glass" files under Glass, not Metalworking.
- Per-row select preselected to the suggestion, blank "— leave as is —" fallback, single apply POST (stale/invalid rows skipped silently), PRG + counted success message.

## Validation

- 18 new specs (6 model, 11 view, 1 import no-clobber); the no-clobber spec verified to fail against the old import code.
- Full `classes/spec` on this branch: 1078 passed; the 25 failures are pre-existing on `origin/main` (missing `segno` in the local test image — identical failure list, CI unaffected).
- Dry-run of the exact keyword map against all 614 guildless prod offerings: **550 (90%) matched**, spot-checked correct across every guild; the 64 unmatched are business/branding classes that genuinely have no guild.

## Notes

- Prod's `Category.guild` links were populated separately (data-only, already live).
- Changelog folded into v0.23.3 as one member-facing entry.


classes/views.py
```diff
@@ -2463,6 +2463,63 @@ def admin_category_delete(request: HttpRequest, pk: int) -> HttpResponse:
...
+            if not value:
+                continue
+            try:
+                offering_pk = int(key.removeprefix("category_"))
+                target_pk = int(value)
+            except ValueError:
+                continue
+            if offering_pk not in guildless_pks or target_pk not in valid_targets:
+                continue
+            offering = ClassOffering.objects.get(pk=offering_pk)
```
PastLivesReviewBot: **WARN (performance/architecture):** This is one `get()` + one `save()` per applied row — applying all ~600 prod offerings in a single POST is ~1,200 queries. Since you already have `guildless_pks`, consider collecting the `(offering_pk, target_pk)` pairs first, then `ClassOffering.objects.in_bulk(...)` + `bulk_update(..., ["category"])` in two queries. Folding that into a manager method (e.g. `ClassOffering.objects.refile_into_guilds(mapping)`) would also move the state change out of the view per the fat-models rule. Fine as-is for a one-time admin surface, so not blocking.

classes/views.py
```diff
@@ -2463,6 +2463,63 @@ def admin_category_delete(request: HttpRequest, pk: int) -> HttpResponse:
...
+    guild_categories = list(Category.objects.filter(guild__isnull=False).order_by("name"))
+
+    if request.method == "POST":
+        valid_targets = {c.pk: c for c in guild_categories}
+        guildless_pks = set(ClassOffering.objects.filter(category__guild__isnull=True).values_list("pk", flat=True))
+        applied = 0
+        for key in request.POST:
+            if not key.startswith("category_"):
+                continue
+            value = request.POST.get(key, "")
```
PastLivesReviewBot: **WARN (standards):** `request.POST.get(key, "")` — `key` is guaranteed present here (you are iterating `request.POST`), so per the repo's fail-loudly rule this should be `request.POST[key]`. The fallback can never fire and just hides intent.

classes/views.py
```diff
@@ -2463,6 +2463,63 @@ def admin_category_delete(request: HttpRequest, pk: int) -> HttpResponse:
...
+                target_pk = int(value)
+            except ValueError:
+                continue
+            if offering_pk not in guildless_pks or target_pk not in valid_targets:
+                continue
+            offering = ClassOffering.objects.get(pk=offering_pk)
+            offering.category = valid_targets[target_pk]
+            offering.save(update_fields=["category"])
+            applied += 1
+        messages.success(request, f"Re-filed {applied} classes into guild categories.")
```
PastLivesReviewBot: **WARN (copy):** Singular case reads "Re-filed 1 classes into guild categories." Suggest pluralizing (`class{{ }}es`-style: `f"Re-filed {applied} class{'es' if applied != 1 else ''} into guild categories."`). Note `it_reports_the_count_in_a_message` currently asserts the ungrammatical string, so update that spec together with the copy.

hub/calendar_service.py
```diff
@@ -196,6 +196,127 @@ def sync_general_calendar() -> int:
...
+            _delete_feed_event_copy(client, event)
+        except DiscordEventsError as exc:
+            failures.append(f"'{event.title}': {exc}")
+
+    if failures:
+        raise DiscordEventsError(f"{len(failures)} Discord event push(es) failed; first: {failures[0]}")
+    return len(push_set)
+
+
+def _push_feed_event(client: Any, event: CalendarEvent) -> None:
```
PastLivesReviewBot: **WARN (typing):** `client: Any` (also on `_delete_feed_event_copy` below) loses type checking on every client call. `DiscordScheduledEventsClient` is importable under `TYPE_CHECKING` (the module already uses lazy runtime imports to avoid the cycle), so these can be annotated `client: "DiscordScheduledEventsClient"` for real coverage.

tests/hub/calendar_service_spec.py
```diff
@@ -2,15 +2,17 @@
...
-from unittest.mock import patch
+from unittest.mock import MagicMock, patch
 
 import pytest
 from django.utils import timezone
 
 from classes.factories import ClassOfferingFactory, ClassSessionFactory
 from classes.models import ClassOffering
 from core.models import SiteConfiguration
+from core.integrations.discord_events import DiscordScheduledEventsClient
```
PastLivesReviewBot, line 14: **WARN (style, cosmetic):** `core.integrations.discord_events` sorts before `core.models`, so this import block is un-sorted per isort. The repo's ruff config excludes `tests/`, so CI won't flag it — but worth tidying if you touch this file again.

PastLivesReviewBot (approved):
Reviewed all three commits: the create-only category import fix, the guild-tagging bulk suggester, the Discord Scheduled Events mirror of the general feeds, and the recurrence_rule.start fix.

Verified against the branch source: every imported name in `hub/calendar_service.py` exists in `core/integrations/discord_events.py` (`DEFAULT_LOCATION`, `_NAME_MAX`, `_LOCATION_MAX`, `_DESCRIPTION_MAX`, `_ENTITY_TYPE_EXTERNAL`, `_PRIVACY_GUILD_ONLY`, `DiscordScheduledEventsClient` with `from_settings`/`enabled`/`server_id`/`insert_event`/`update_event`/`delete_event`); `start` is in scope where `rule["start"]` is set and matches `scheduled_start_time`; migration 0096 is purely additive with the correct 0095 dependency and matches the model field exactly; `classes:admin_guild_tagging` URL, `classes_admin_access_required`, and the `Category.guild` FK all exist; `create_defaults` semantics on `update_or_create` are correct for the don't-clobber behavior (set on create, untouched on update). Ruff (repo config) is clean on all changed non-excluded files. The POST handler validates targets against guild-linked categories and offerings against the guildless set, behind the admin decorator with CSRF — no security concerns. Tests are thorough BDD specs covering the suggester ordering/fallthrough, view access/GET/POST edge cases, the import no-clobber, and the full Discord push lifecycle including the 404-recreate and drop-set deletion paths.

No blocking issues. Approving with a handful of minor (non-blocking) suggestions inline.

PastLivesReviewBot (approved):
Re-review of 7c42aeb, which addresses the five WARNs from the previous approval. All five are verifiably fixed:

1. **Per-row get()+save() loop** — the POST now collects `(offering_pk → target_pk)` assignments and delegates to a new `ClassOfferingQuerySet.refile_into_guild_categories`, which loads both sides via one `in_bulk` each. The remaining per-row `save(update_fields=["category"])` is deliberate and correct: `ClassOffering.save` recomputes `grouping_key` (and appends it to `update_fields` when `category` is present) plus syncs grouped siblings — behavior `bulk_update` would silently skip. The state change also moved out of the view per fat-models. Verified the guildless-offerings-only / guild-linked-targets-only / silent-stale-skip semantics and the returned count.
2. **Grammar** — `"Re-filed {applied} {noun}"` with singular/plural noun; `it_reports_the_count_in_a_message` updated to assert `"Re-filed 1 class"`.
3. **`request.POST[key]`** — the dead `.get(key, "")` fallback is gone.
4. **Typed Discord client** — `_push_feed_event`/`_delete_feed_event_copy` now annotate `client: DiscordScheduledEventsClient` via a `TYPE_CHECKING` import; safe at runtime (`from __future__ import annotations` is present) and inside CI's mypy scope (`hub/`), where lint passed.
5. **Import sort** — `core.integrations.discord_events` now sorts before `core.models` in `tests/hub/calendar_service_spec.py`.

Re-scanned the full diff including the fix commit: the new manager method's names all resolve (`Category` is module-local, `objects = ClassOfferingQuerySet.as_manager()` exposes it), annotations are complete, ruff format/check are clean under the repo config, and the POST specs still cover the path meaningfully (valid apply, singular count message, empty value, non-guild target skipped, stale pk skipped) alongside the existing suggester and Discord-mirror specs. CI on 7c42aeb: lint, test, and e2e all green. No new issues introduced. Approving.

PR #133, opened 2026-07-21
[calendar] Suppress feed echoes of pushed events + honor Discord 429 (v0.23.4)

HexagonStorms:
## What

Two prod defects found while verifying the v0.23.3 calendar/Discord go-live:

**1. Feed echo loop.** The Member/Public Google calendars we push CommunityEvents to are the same calendars the general feed sync re-imports. Every pushed event came back as feed `CalendarEvent` rows under its Google-assigned UID — observed live: **'Parallel Play' appeared 5× in Discord** (its own push + 4 mirrored feed echoes) and 26 echo rows landed in the calendar table.
- Feed import now skips UIDs matching any `CommunityEvent.google_ical_uid`
- The Discord mirror excludes echoes from its push set; the existing dropped-pass deletes the Discord copies they already gained
- Echo rows are then purged (rows whose Discord delete failed survive to retry next sweep)
- **Self-healing**: the next daily sweep cleans up the existing prod duplicates — no manual DB surgery

**2. Discord 429s fail every sweep.** The scheduled-events rate bucket is tiny; the mirror's burst of ~a dozen calls tripped it mid-run on both one-off sweeps today (`retry_after` 1–10 s). The client now honors `Retry-After` on a 429 (bounded at 15 s, one retry); a 429 with no usable or excessive wait still raises.

## Tests
- 3 new mirror specs (echo never pushed + row purged; pre-existing Discord copy deleted; failed delete survives for retry)
- 1 new feed-import spec (echo UID skipped, sibling imported)
- 5 new client specs (retry honored + sleep asserted; double-429 raises; missing/malformed/excessive Retry-After don't retry)
- ruff + mypy clean; touched spec files green locally (community_calendar specs require `recurring_ical_events`, present in CI)

## Changelog
No new entry — intra-cycle fix to the unannounced v0.23.3 calendar/Discord feature. Both 0.23.3 entries re-stamped to 0.23.4 so the manual Discord announcement posts them together.

### DIFF
--- a/core/integrations/discord_events.py
+++ b/core/integrations/discord_events.py
@@ -108,32 +108,49 @@ class DiscordScheduledEventsClient:
     def server_id(self) -> str:
         return self._server_id
 
-    def insert_event(self, server_id: str, body: dict[str, Any]) -> dict[str, Any]:
+    def insert_event(
+        self, server_id: str, body: dict[str, Any], *, retry_on_rate_limit: bool = False
+    ) -> dict[str, Any]:
         """Create a Scheduled Event; return the event resource (with ``id``)."""
-        return self._execute("POST", f"/guilds/{server_id}/scheduled-events", json=body)
+        return self._execute(
+            "POST", f"/guilds/{server_id}/scheduled-events", json=body, retry_on_rate_limit=retry_on_rate_limit
+        )
 
-    def update_event(self, server_id: str, event_id: str, body: dict[str, Any]) -> dict[str, Any]:
+    def update_event(
+        self, server_id: str, event_id: str, body: dict[str, Any], *, retry_on_rate_limit: bool = False
+    ) -> dict[str, Any]:
         """Patch an existing Scheduled Event; return the updated event resource."""
-        return self._execute("PATCH", f"/guilds/{server_id}/scheduled-events/{event_id}", json=body)
+        return self._execute(
+            "PATCH",
+            f"/guilds/{server_id}/scheduled-events/{event_id}",
+            json=body,
+            retry_on_rate_limit=retry_on_rate_limit,
+        )
 
-    def delete_event(self, server_id: str, event_id: str) -> None:
+    def delete_event(self, server_id: str, event_id: str, *, retry_on_rate_limit: bool = False) -> None:
         """Delete a Scheduled Event. Wraps API errors (caller catches one type)."""
-        self._execute("DELETE", f"/guilds/{server_id}/scheduled-events/{event_id}")
+        self._execute(
+            "DELETE", f"/guilds/{server_id}/scheduled-events/{event_id}", retry_on_rate_limit=retry_on_rate_limit
+        )
 
     @staticmethod
-    def _execute(method: str, path: str, *, json: dict[str, Any] | None = None) -> dict[str, Any]:
+    def _execute(
+        method: str, path: str, *, json: dict[str, Any] | None = None, retry_on_rate_limit: bool = False
+    ) -> dict[str, Any]:
         """Run a bot-authed REST call, translating any failure into :class:`DiscordEventsError`
         so callers catch one exception type (mirrors ``GoogleCalendarClient._execute``).
 
         Both a transport failure (``httpx.HTTPError``) and a non-2xx status become a
         ``DiscordEventsError``; a 2xx with an empty body (a ``DELETE`` 204) returns ``{}``.
-        A 429 is retried once after Discord's ``Retry-After`` — the scheduled-events
-        bucket is tiny, so the daily mirror's burst of ~a-dozen calls reliably trips it
-        mid-run; waiting the advertised second or two clears it. A 429 with no usable or
-        too-long ``Retry-After`` still raises.
+        With ``retry_on_rate_limit`` a 429 is retried once after Discord's ``Retry-After``
+        — the scheduled-events bucket is tiny, so the daily mirror's burst of ~a-dozen
+        calls reliably trips it mid-run; waiting the advertised second or two clears it.
+        Only batch paths opt in: an interactive FOG save should fail fast into the retry
+        cron, not hang the request sleeping. A 429 without the flag, or with no usable or
+        too-long ``Retry-After``, raises immediately.
         """
         response = DiscordScheduledEventsClient._send(method, path, json=json)
-        if response.status_code == 429:
+        if response.status_code == 429 and retry_on_rate_limit:
             retry_after = _retry_after_seconds(response)
             if retry_after is not None and retry_after <= _RATE_LIMIT_MAX_WAIT_SECONDS:
                 time.sleep(retry_after)
--- a/core/integrations/discord_events.py
+++ b/core/integrations/discord_events.py
@@ -31,6 +31,7 @@ clean — exactly as ``push_to_google`` does.
 from __future__ import annotations
 
 import logging
+import time
 from typing import TYPE_CHECKING, Any
 
 import httpx
@@ -51,6 +52,7 @@ logger = logging.getLogger(__name__)
 DEFAULT_LOCATION = "Past Lives Makerspace"
 
 _TIMEOUT_SECONDS = 5.0
+_RATE_LIMIT_MAX_WAIT_SECONDS = 15.0
 _SYNC_ERROR_MAX = 500
 _NAME_MAX = 100
 _LOCATION_MAX = 100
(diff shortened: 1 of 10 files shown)


hub/calendar_service.py
```diff
@@ -131,6 +132,17 @@ def _fetch_and_parse(url: str, window_start: datetime, window_end: datetime) ->
...
 
+def _pushed_event_uids() -> set[str]:
+    """iCal UIDs of CommunityEvents we ourselves pushed to Google Calendar.
+
+    The Member/Public calendars we push to are the same calendars the general feeds
+    re-import, so our own events come back in the feed under their Google-assigned
+    UID. Importing those copies would double-list every pushed event (and the Discord
+    mirror would then duplicate its Scheduled Event too) — they must be skipped.
+    """
+    return set(CommunityEvent.objects.exclude(google_ical_uid="").values_list("google_ical_uid", flat=True))
```
PastLivesReviewBot: **WARN (reuse):** `CommunityEventQuerySet.pushed()` already exists (membership/models.py:3407) with the docstring "Rows FOG has pushed to Google (their iCal echo must be de-duped)" — it is this exact filter. Prefer `CommunityEvent.objects.pushed().values_list("google_ical_uid", flat=True)` so the de-dupe predicate has one source of truth in the manager, per the querysets-live-in-managers convention. Behavior is identical, so not blocking.

hub/calendar_service.py
```diff
@@ -280,6 +301,8 @@ def sync_discord_feed_events() -> int:
         except DiscordEventsError as exc:
             failures.append(f"'{event.title}': {exc}")
 
+    _purge_echoed_feed_events(echo_uids, now)
```
PastLivesReviewBot, line 308: **WARN (coupling):** the echo-row purge only runs when Discord Events sync is enabled — `sync_discord_feed_events` returns 0 at the top when the client is disabled, before reaching this line. New echoes are suppressed at import either way, but any pre-existing echo rows would stay double-listed on the hub calendar for as long as the toggle is off. Prod has the toggle on, and the coupling is defensible (rows must outlive a failed Discord delete), so just noting the dependency — if the toggle is ever switched off before the first clean sweep, the prod duplicates won't self-heal.

core/integrations/discord_events.py
```diff
@@ -125,9 +127,26 @@ def _execute(method: str, path: str, *, json: dict[str, Any] | None = None) -> d
...
+        A 429 is retried once after Discord's ``Retry-After`` — the scheduled-events
+        bucket is tiny, so the daily mirror's burst of ~a-dozen calls reliably trips it
+        mid-run; waiting the advertised second or two clears it. A 429 with no usable or
+        too-long ``Retry-After`` still raises.
         """
+        response = DiscordScheduledEventsClient._send(method, path, json=json)
+        if response.status_code == 429:
+            retry_after = _retry_after_seconds(response)
+            if retry_after is not None and retry_after <= _RATE_LIMIT_MAX_WAIT_SECONDS:
+                time.sleep(retry_after)
```
PastLivesReviewBot, line 156: **WARN (latency):** this retry sleep now applies to every `_execute` caller, including `push_community_event`, which runs on interactive FOG event saves — a rate-limited save can now hang the request up to ~15s (sleep) + a second 5s call, where before it failed fast into `SyncState.FAILED` and retried nightly. Bounded and rare, and it's the right behavior for the cron mirror; just flagging that the request path inherits it. If it ever bites, cap the interactive path lower or retry only from the sweep.

hub/views.py
```diff
@@ -3505,15 +3505,9 @@ def community_calendar(request: HttpRequest) -> HttpResponse:
...
-    default_filters = ["community"]
-    for feed in cal_ctx["calendar_feeds"]:
-        default_filters.append(f"feed-{feed.pk}")
-    if cal_ctx["has_ungrouped_classes"]:
-        default_filters.append("classes")
-    for g in cal_ctx["legend_guilds"]:
-        default_filters.append(str(g.pk))
-
-    cal_ctx["default_filters_json"] = json.dumps(default_filters).replace('"', '\\"')
+    # No default_filters_json here: the Community Calendar persists *disabled*
```
PastLivesReviewBot, line 3502: **WARN (follow-up):** the guild calendar surface still uses the enabled-list pattern this PR retires (`default_filters_json` at hub/views.py:556 + `guildCalFilters-{{ guild.pk }}` in templates/hub/partials/guild_calendar_app.html) and so still has the same hide-new-keys flaw described in the template comment. Fine to leave out of this fix PR, but worth a follow-up to invert that surface too before it needs its own rollout hack.

PastLivesReviewBot (approved):
Reviewed all four fixes against the branch with whole-file context.

**Verified:**
- Feed-echo suppression: `_pushed_event_uids()` matches the real `CommunityEvent.google_ical_uid` field (CharField, `default=""`, non-null — the `exclude(google_ical_uid="")` filter is sound). The skip applies in both `_upsert_events` and the Discord push set, and the purge correctly spares future rows whose Discord delete failed (`discord_event_id` still set) so they retry next sweep.
- 429 handling: single bounded retry, sleeps only on a usable `Retry-After` <= 15s; missing/malformed/excessive headers still raise without sleeping; a second 429 raises. `_send` isolates transport errors exactly as before.
- Class purge-by-pk: `update_or_create` keys on `(guild, uid)`, so a re-guilded class creates a new row and the pk-based purge removes the old-guild copy — the old uid-based purge would indeed have kept it and double-listed the class.
- Disabled-list filters: every `isActive`/`toggleFilter` key in the template and the event partials (`calendar_event_item.html`, `calendar_content.html`) is consistent with the inverted semantics; legacy localStorage keys are dropped once; `hub/views.py` still uses `json` elsewhere (no dead import).
- Ran all three touched spec files in a CI-mirroring container (SQLite + `recurring_ical_events` installed): **124 passed**. `ruff check` + `ruff format --check` clean on every changed file. (The earlier local failures were solely the missing `recurring_ical_events` module, as the PR body notes.)
- Changelog handling follows the documented re-stamp rule (both 0.23.3 entries folded forward to 0.23.4; `VERSION` matches an entry).

No blocking issues. Four WARN-level notes inline — none holds up the merge.

hub/calendar_service.py
```diff
@@ -141,9 +153,12 @@ def _upsert_events(
 
     Uniqueness is per-``(guild, feed, uid, recurrence_id)`` so each occurrence of a recurring
     series gets its own row, and two ``CalendarFeed`` rows whose upstream calendars happen to
-    share a UID don't clobber each other.
+    share a UID don't clobber each other. Echoes of our own Google-pushed CommunityEvents
+    (see :func:`_pushed_event_uids`) are skipped, not imported.
     """
     now = timezone.now()
+    echo_uids = _pushed_event_uids()
```
PastLivesReviewBot, line 160: **WARN (minor, non-blocking):** `_pushed_event_uids()` is recomputed inside `_upsert_events`, i.e. once per guild feed + once per `CalendarFeed` during a full `sync_all_sources` run (plus once more in `sync_discord_feed_events`). It's a cheap single-column query so this is fine today, but if the feed count grows it could be computed once per sync run and passed down. Not worth holding the PR for.

PastLivesReviewBot (approved):
Re-review of 47d80f7 against the four WARNs from the previous review — all four are verified fixed in the code (not just the commit message):

1. **`pushed()` reuse** — `_pushed_event_uids()` now returns `set(CommunityEvent.objects.pushed().values_list("google_ical_uid", flat=True))`, reusing the manager's `pushed()` (`exclude(google_ical_uid="")`). No empty-string UID can enter the echo set.
2. **Toggle-off purge** — `now`/`echo_uids` are computed before the `client.enabled` check and `_purge_echoed_feed_events()` runs in the disabled path before `return 0`. The purge correctly touches only safe rows (`discord_event_id=""` or past), the enabled path runs it exactly once (after the dropped pass, before the failure raise, so failures don't skip it), and a row whose Discord delete failed survives for the next sweep — all spec-covered.
3. **Opt-in 429 retry** — `retry_on_rate_limit` defaults to `False` everywhere; only the batch mirror in `hub/calendar_service.py` passes `True`. The interactive `sync_community_event_to_discord` path never opts in, so FOG saves fail fast into the retry cron. All six new branch specs (retry success, no-flag fail-fast, double-429, missing/malformed/excessive `Retry-After`) pass with `time.sleep` mocked.
4. **Guild-tab filter inversion** — `guild_calendar_app.html` now persists `guildCalFiltersOff-{pk}` with the same disabled-list logic as the Community Calendar, both views drop `default_filters_json` (no stale references remain), legacy localStorage keys are cleaned up once, and the specs assert on rendered toggles instead of the removed context key.

New-issue sweep found nothing blocking: the purge does not double-run in the enabled path, `_upsert_events` returns the post-filter count, the `kept_pks` switch in `sync_local_class_events` is correct against the `(guild, uid)` upsert key, `CommunityEventFactory` only sets `google_ical_uid` via the `pushed` trait, and the changelog is properly re-stamped at 0.23.4. Ran the four touched spec files in the CI-mirroring container: 139 passed. `ruff check` + `ruff format --check` clean on all changed files.

One minor WARN inline (non-blocking). Approving.

PR #134, opened 2026-07-21
[calendar] Click +N on a busy day to expand all events (v0.23.6)

HexagonStorms:
## What
On the Community Calendar, a day cell with many events truncates the list and shows a `+N` overflow tag. This makes that tag a real control: clicking it **expands the day cell in place** to reveal every event for that day, and flips the label to **"Show less"** to collapse again.

- Works in **both the week grid** (shows 3, `+N` for the rest) **and the month grid** (shows 2).
- Works on the **Community Calendar** and any **guild-page calendar** — both render the shared `hub/partials/calendar_content.html`.
- Each day cell owns its own `expanded` state (nested Alpine `x-data`), so opening one day doesn't touch its neighbors.
- Respects the existing legend filters: expanding reveals the day's *active* events; overflow chips are gated on `isActive(src) && expanded`.

## Why
Busy days hid their events behind a dead `+N` label with no way to see what was underneath without leaving the calendar. This is a small QoL fix to the already-live calendar.

## How
- `calendar_content.html`: render all of a day's events (instead of slicing), gate the overflow chips behind `expanded`, and turn the `+N` span into a toggle `<button>` (`@click`, `x-text`, `:aria-expanded`).
- `calendar.css`: new `pl-`-prefixed rules — `.pl-calendar-grid__more--toggle` (clickable underline styling, theme-token colors) and `.pl-calendar-grid__day--expanded` (lets the cell grow past its fixed height so revealed chips aren't clipped; the grid row grows with it). Verified against both dark (Obsidian) and light (Slate) themes using existing tokens only — no new colors.

## Tests
`tests/hub/calendar_day_expand_spec.py` — renders the partial and asserts the expand control markup for both grids (toggle appears only past the per-view limit, all events land in the DOM, overflow gated behind `expanded`, collapse affordance present, per-cell state). 54 passed (6 new + existing `community_calendar_spec.py`).

### DIFF
--- a/static/css/calendar.css
+++ b/static/css/calendar.css
@@ -373,6 +373,35 @@
     flex-shrink: 0;
 }
 
+/* "+N" overflow indicator as a clickable toggle — expands the day cell in place
+   to reveal every event, then flips to "Show less". */
+.pl-calendar-grid__more--toggle {
+    border: 0;
+    background: none;
+    font-family: inherit;
+    text-align: left;
+    cursor: pointer;
+    text-decoration: underline;
+    text-underline-offset: 2px;
+    transition: color 0.15s;
+}
+
+.pl-calendar-grid__more--toggle:hover {
+    color: var(--hub-text);
+}
+
+/* Expanded day: let the cell grow past its fixed height so the revealed chips
+   aren't clipped. The grid row grows with it. Beats the week/month height rules
+   via the doubled class selector. */
+.pl-calendar-grid__day.pl-calendar-grid__day--expanded {
+    height: auto;
+    overflow: visible;
+}
+
+.pl-calendar-grid__day--expanded .pl-calendar-grid__day-chips {
+    overflow: visible;
+}
+
 .pl-calendar-divider {
     height: 1px;
     margin: 1.5rem 0;
--- a/templates/hub/partials/calendar_content.html
+++ b/templates/hub/partials/calendar_content.html
@@ -27,17 +27,19 @@
 {# Week grid #}
 <div class="pl-calendar-grid pl-calendar-grid--week" x-show="calView === 'week'">
     {% for day in week_days %}
-    <div class="pl-calendar-grid__day{% if day.is_today %} pl-calendar-grid__day--today{% endif %}">
+    <div class="pl-calendar-grid__day{% if day.is_today %} pl-calendar-grid__day--today{% endif %}"
+         x-data="{ expanded: false }"
+         :class="{ 'pl-calendar-grid__day--expanded': expanded }">
         <div class="pl-calendar-grid__day-header">
             <span class="pl-calendar-grid__day-name">{{ day.date|date:"D" }}</span>
             <span class="pl-calendar-grid__day-num{% if day.is_today %} pl-calendar-grid__day-num--today{% endif %}">{{ day.date|date:"j" }}</span>
         </div>
         <div class="pl-calendar-grid__day-chips">
-            {% for event in day.events|slice:":3" %}
+            {% for event in day.events %}
             {% with src=event.source_key %}
             <button type="button"
                     class="pl-calendar-grid__chip{% if event.is_in_progress %} pl-calendar-grid__chip--live{% endif %}"
-                    x-show="isActive('{{ src }}')"
+                    x-show="isActive('{{ src }}'){% if forloop.counter > 3 %} && expanded{% endif %}"
                     @click="focusEvent({{ event.pk }})"
                     @mouseenter="hoverEvent({{ event.pk }})" @mouseleave="unhoverEvent({{ event.pk }})"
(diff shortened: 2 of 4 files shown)


templates/hub/partials/calendar_content.html
```diff
@@ -48,7 +50,11 @@
             {% endwith %}
             {% endfor %}
             {% if day.events|length > 3 %}
-            <span class="pl-calendar-grid__more">+{{ day.events|length|add:"-3" }}</span>
+            <button type="button"
+                    class="pl-calendar-grid__more pl-calendar-grid__more--toggle"
```
PastLivesReviewBot, line 54: **WARN (a11y):** The toggle's accessible name while collapsed is just "+3", which tells a screen-reader user very little. Consider an explicit label, e.g. `:aria-label="expanded ? 'Show fewer events' : 'Show all {{ day.events|length }} events'"`, and optionally `aria-controls` pointing at an id on the chips container. (Applies to the month-grid toggle too.) Note `:aria-expanded="expanded"` itself is fine — Alpine preserves `aria-expanded` as `"false"` when falsy.

templates/hub/partials/calendar_content.html
```diff
@@ -99,7 +107,11 @@
             {% endwith %}
             {% endfor %}
             {% if day.events|length > 2 %}
-            <span class="pl-calendar-grid__more">+{{ day.events|length|add:"-2" }}</span>
+            <button type="button"
+                    class="pl-calendar-grid__more pl-calendar-grid__more--toggle"
```
PastLivesReviewBot, line 111: **WARN (edge case):** The "+N" count is computed from `day.events|length`, but each chip is still gated on `isActive(src)`. With legend filters toggled off, "+3" may expand to reveal fewer than 3 chips (or none). The count ignoring filters is pre-existing behavior, but now that it's an interactive "show all" control, a filter-aware count (or simply hiding the toggle when no overflow chip is active) would be a nice follow-up. Not blocking.

static/css/calendar.css
```diff
@@ -373,6 +373,35 @@
...
+
+.pl-calendar-grid__more--toggle:hover {
+    color: var(--hub-text);
+}
+
+/* Expanded day: let the cell grow past its fixed height so the revealed chips
+   aren't clipped. The grid row grows with it. Beats the week/month height rules
+   via the doubled class selector. */
+.pl-calendar-grid__day.pl-calendar-grid__day--expanded {
+    height: auto;
```
PastLivesReviewBot, line 397: **WARN (cosmetic):** Two height quirks worth a quick visual check:

1. `height: auto` can let a just-over-limit month cell *shrink* below its collapsed 128px when expanded (e.g. exactly 3 short chips + toggle can total < 128px). A `min-height: 128px` here (and `min-height: 176px` on a `--week` variant) would pin the floor.
2. When one expanded cell grows its grid row, sibling cells keep their fixed `height: 128px/176px`, so their `border-right`/`border-bottom` stop short of the taller row's bottom edge. If that artifact shows, `height: 100%` + `min-height` on the base cell (instead of fixed height) fixes it.

The PR says both themes were eyeballed, so approving — just flagging where to look if a seam appears.

PastLivesReviewBot (approved):
Independent review of #134. Verified against the PR head (`feat/calendar-day-expand`):

- **Alpine scope**: the nested `x-data="{ expanded: false }"` on each day cell correctly inherits the parent calendar component's scope in Alpine 3, so `isActive`, `focusEvent`, `hoverEvent`, `unhoverEvent` all resolve up the chain in both `community_calendar.html` (line 76 component) and `guild_calendar_app.html` (line 9 component). Neither parent defines `expanded`, so nothing is shadowed, and each cell's state is isolated as claimed.
- **`:aria-expanded="expanded"`** is sound — Alpine 3 explicitly preserves `aria-expanded` when falsy and renders `"false"`/`"true"`, so the toggle reports state correctly.
- **CSS**: new classes are `pl-`-prefixed; the doubled-class expanded rule (0,2,0, later in file) does beat `.pl-calendar-grid--week .pl-calendar-grid__day` (0,2,0, earlier). No inline `display` on any `x-show` element; the only inline style is the pre-existing `--chip-color` custom property.
- **Tests**: BDD `describe_`/`it_` (no `context_` blocks), placed in `tests/hub/` matching the existing calendar specs; direct `CalendarEvent.objects.create` matches every existing calendar spec (no `CalendarEventFactory` exists), and all model fields used (`guild`, `uid`, `title`, `start_dt`, `end_dt`, `fetched_at`) are real. mypy excludes `tests/`, so the partially-annotated helpers match repo convention.
- **Version**: 0.23.4 -> 0.23.6 skips 0.23.5, which is claimed by open PR #135 — correct. Changelog entry version matches `VERSION`.

No blocking issues. Three WARN-level suggestions inline — approving.

PR #135, opened 2026-07-21
[classes] Per-class editable FAQs + gallery under the booking rail (v0.23.5)

HexagonStorms:
## What

Two class-detail-page gaps Josh flagged:

1. **The "Questions" section is now editable per class.** The three site-wide questions (cancellation policy, accessibility, beginners) stay as the default for every class, but the class edit form (admin + teach portals) now has a **Questions & answers** editor seeded with those defaults as a starting point — reword them, add more, or delete ones that don't fit. Deleting every row returns the class to the site defaults. New `ClassFaq` model (migration 0047); answers are plain text with emails/URLs auto-linked.

2. **Gallery images now appear directly under the registration card** in the sticky rail (as their own section), instead of mid-column above the schedule. Only the class's actual gallery shots render there — the hero no longer duplicates into the gallery. Sticky moved from the rail card to a wrapper so the card and gallery scroll as one unit.

## Notes for review
- The FAQ formset rides the main class form POST (like the sessions formset), so existing edit-POST specs gained the `faq-*` management-form keys.
- Seeding uses prefilled extra forms; on POST they carry data against empty initial, so untouched defaults still materialize as rows — editing one default can never silently drop the other two.
- `classes/spec/views/class_preview_spec.py` gallery assertions updated to the new rail placement.

## Tests
- `classes/spec` full suite: **1123 passed** (docker, SQLite).
- New specs: `classes/spec/models/class_faq_spec.py`, `classes/spec/views/class_faq_editor_spec.py` (seeding, saving, deletion, invalid-row re-render, teach portal, public rendering incl. urlize, rail gallery presence/absence).
- mypy: no new errors (37 pre-existing on baseline, unchanged).


templates/classes/public/detail.html
```diff
@@ -409,6 +401,15 @@ <h2 class="cp-detail__h2">Questions</h2>
           {% endif %}
         </ul>
       </div>
+
+      {# Gallery — the class's extra shots, right under the booking card (hover-zoom + lightbox from the component) #}
+      {% if offering.gallery_display_images %}
```
PastLivesReviewBot, line 406: WARN: `offering.gallery_display_images` is evaluated twice here (the `{% if %}` and the `images=` on line 409), and each call runs a fresh `gallery_images.all()` query since `public_class_detail` doesn't prefetch `gallery_images`. Constant cost, not an N+1, but a `{% with shots=offering.gallery_display_images %}` wrapper (or adding `"gallery_images"` to the view's `prefetch_related`) would halve it for free.

templates/classes/_components/gallery.html
```diff
@@ -13,7 +14,7 @@
     arrows step through, Esc closes.
   - Touch devices fall back to tap-to-lightbox (no hover zoom).
 {% endcomment %}
-{% with images=offering.display_images %}
+{% with images=images|default:offering.display_images %}
```
PastLivesReviewBot, line 17: WARN: the `|default:offering.display_images` fallback is now dead code — this component's only include (detail.html rail) always passes `images=`. Worth noting because the hero `object-position` crop styling was removed from this template in the same PR: if the fallback path is ever reactivated (hero-first list), the hero will render without its crop. Either drop the fallback or leave a comment about the crop.

classes/forms.py
```diff
@@ -382,6 +384,42 @@ class Meta:
...
+
+    class Meta:
+        model = ClassFaq
+        fields = ["question", "answer"]
+        widgets = {
+            "answer": forms.Textarea(attrs={"rows": 3}),
+        }
+
+
+def build_class_faq_formset(data: Any, offering: ClassOffering) -> Any:
```
PastLivesReviewBot, line 398: WARN: `data: Any` / `-> Any` is looser than it needs to be — `data: QueryDict | None` and a `BaseInlineFormSet` return annotation would give mypy something to check while staying compatible with the dynamic `inlineformset_factory` class. Consistent with the existing `formset: Any` convention in views.py, so purely a suggestion.

classes/models.py
```diff
@@ -1259,6 +1279,60 @@ def save(self, *args, **kwargs) -> None:
         super().save(*args, **kwargs)
 
 
+# Site-wide starting-point FAQs shown on every class page until the class saves its own
+# ClassFaq rows. The class edit form seeds these as editable rows, so instructors can
+# reword them or add more; answers are plain text (urlize turns the email into a link).
+DEFAULT_CLASS_FAQS: list[dict] = [
```
PastLivesReviewBot, line 1285: WARN: `list[dict]` — the entries are homogeneous `{"question": str, "answer": str}`, so `list[dict[str, str]]` documents the shape and lets mypy catch a stray key type. Same nit applies to the `-> list[dict]` returns on `display_faqs` / `gallery_display_images` / `display_images`.

classes/spec/views/admin_classes_spec.py
```diff
@@ -207,6 +207,10 @@ def it_creates_a_class(admin_user, client, db):
                 "sessions-INITIAL_FORMS": "0",
                 "sessions-MIN_NUM_FORMS": "0",
                 "sessions-MAX_NUM_FORMS": "1000",
+                "faq-TOTAL_FORMS": "0",
```
PastLivesReviewBot, line 210: WARN: these `faq-*` management keys (here and in the other create-view payloads in this file, teach_dashboard_spec, teach_submit_image_gate_spec) are inert — `admin_class_create` / `teach_class_create` never build the FAQ formset, so the keys are ignored. Harmless, but they imply the create form has the editor when it's edit-only; dropping them from the create payloads (or a one-line comment) would keep the specs honest.

PastLivesReviewBot (approved):
Reviewed the full diff against the PLFOG standards (fat models, model field standards, migrations, typing, BDD specs, security, N+1). Solid PR: the ClassFaq model is fully standard (help_text everywhere, __str__, CASCADE, additive migration 0047), the formset builder lives in forms.py so views stay thin, the editor template matches the canonical guild_edit.html pattern exactly (hidden DELETE + requestSubmit for saved rows, empty-form template + TOTAL_FORMS increment for adds, {% comment %} blocks), public rendering escapes questions and urlize/linebreaks the plain-text answers, and the new specs cover seeding, save-with-untouched-defaults, deletion, invalid re-render, teach portal, public fallback, and the rail gallery on/off. No blocking issues — approving with a few WARN-level notes inline.

PR #136, opened 2026-07-21
[calendar] Events tab lists every calendar event + compact Google-sync badge (v0.23.7)

HexagonStorms:
## What & why

Three fixes to the Community Calendar's **Events tab** (`/calendar/?tab=events`), reported by Josh.

### 1. Events tab was missing events (+ pagination)
**Root cause:** the Events tab listed only `CommunityEvent.objects.published().upcoming()` — FOG-native community events. The calendar *grid* renders the union of `CalendarEvent` rows (guild / general / class iCal feeds) **plus** synthetic community-event entries, so every feed/class event on the grid was absent from the tab.

The tab now sources the same union the grid shows, via a new `hub.calendar_entries.upcoming_calendar_events()`: the echo-deduped feed/general/class `CalendarEvent` rows **plus** one entry per published `CommunityEvent` at its next occurrence (a recurring series collapses to a single upcoming row, so admins get one Edit/Delete per event, not one per occurrence). The list is paginated with the hub's standard `Paginator` (`?events_page=N`, `_CALENDAR_PAGE_SIZE` = 10).

### 2. "Synced to Google" flag → compact icon badge, shown on both views
The wordy `Synced to Google ✓` pill is now a small green calendar-check glyph (`_sync_flag.html`, `.pl-sync-flag`), with the words preserved in the `title`/`aria-label`. It now also appears on the calendar list below the grid — not just the Events tab. The flag keeps its existing contract: **admin-only and gated by both Google-sync switches** (I did **not** broaden the audience — that's an existing product decision encoded in `sync_badge_spec`; flag if it should change).

### 3. "Register →" → "More Info →" + clickable titles
Event cards now say **More Info →** instead of Register, and the event **title** is itself a link to the detail/registration page (community events → their FOG detail page; classes now link too, in a new tab).

## Files
- `hub/calendar_entries.py` — `CalendarEntry.community_event` field + `upcoming_calendar_events()` builder
- `hub/views.py` — `community_calendar` paginates the union; `_get_calendar_context` exposes `sync_flag_visible`
- `templates/hub/partials/{calendar_event_item,_sync_badge,_sync_flag,calendar_content}.html`, `templates/hub/community_calendar.html`
- `static/css/{hub,calendar}.css`
- `plfog/version.py` — v0.23.7 + member-facing changelog entry

## Tests
New `tests/hub/events_tab_parity_spec.py` (parity, pagination, More Info + title-link markup, flag gating on both views). Full relevant suite green in Docker (SQLite): **235 passed**.

### DIFF
--- a/hub/calendar_entries.py
+++ b/hub/calendar_entries.py
@@ -12,13 +12,13 @@ location, description, guild, feed``.
 from __future__ import annotations
 
 from dataclasses import dataclass
-from datetime import date, datetime
-from typing import TYPE_CHECKING
+from datetime import date, datetime, timedelta
+from typing import TYPE_CHECKING, Any
 
 from django.utils import timezone
 
 if TYPE_CHECKING:
-    from membership.models import Guild
+    from membership.models import CommunityEvent, Guild
 
 # Offsets keep synthetic pks clear of real CalendarEvent pks so the shared
 # focusEvent() JS and the month_event_pages map keep working untouched.
@@ -26,6 +26,11 @@ CLASS_PK_OFFSET = 1_000_000_000
 ORIENTATION_PK_OFFSET = 2_000_000_000
 EVENT_PK_OFFSET = 3_000_000_000
 _OCC_STRIDE = 100  # max occurrences per event per window (a few months of monthly « 100)
+
+# How far ahead the Events tab looks for a recurring series' next occurrence. A
+# year comfortably contains the next hit for any monthly/weekly cadence, so each
+# recurring event resolves to one upcoming row instead of its stale anchor date.
+_EVENTS_TAB_HORIZON_DAYS = 365
 
 
 @dataclass
@@ -43,6 +48,10 @@ class CalendarEntry:
     all_day: bool = False
     guild: Guild | None = None
     feed: None = None
+    # The backing FOG-native event for a "community" entry, so the templates can draw
+    # its Google-sync flag and (on the Events tab) its admin Edit/Delete controls.
+    # Always None for feed / class / orientation entries.
+    community_event: CommunityEvent | None = None
 
     @property
     def source_key(self) -> str:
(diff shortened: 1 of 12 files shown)


hub/calendar_entries.py
```diff
@@ -143,6 +152,55 @@ def community_event_entries(fetch_from: date, fetch_to: date, guild: Guild | Non
                     description=ev.description,
                     all_day=False,
                     guild=ev.guild,
+                    community_event=ev,
                 )
             )
     return entries
+
+
+def upcoming_calendar_events() -> list[Any]:
```
PastLivesReviewBot, line 161: WARN (types): `-> list[Any]` gives mypy nothing to check on every consumer of this list. The element type is knowable: `list[CalendarEvent | CalendarEntry]` (with `CalendarEvent` added to the `TYPE_CHECKING` import). I see the `list[Any]` precedent for `all_events` in `_get_calendar_context`, so not blocking — but this is a new public helper and a precise union here is cheap.

Minor note on the same line: `CalendarEvent.objects.upcoming()` has no upper time bound, so a feed that publishes far-future events materializes them all on every Events-tab request (pagination only trims the rendered slice). Fine at current scale; a `start_dt__lte` horizon matching `_EVENTS_TAB_HORIZON_DAYS` would keep it bounded if feeds ever grow.

hub/calendar_entries.py
```diff
@@ -143,6 +152,55 @@ def community_event_entries(fetch_from: date, fetch_to: date, guild: Guild | Non
...
+        CalendarEvent.objects.upcoming().exclude(uid__in=pushed_uids).select_related("guild", "feed")
+    )
+
+    # One entry per published CommunityEvent, anchored on its next (or in-progress) occurrence.
+    for ev in CommunityEvent.objects.published().upcoming().select_related("guild"):
+        duration = ev.ends_at - ev.starts_at
+        upcoming_occ = [occ for occ in ev.occurrences_in(today, horizon) if occ + duration >= now]
+        # No in-window occurrence → a still-running non-recurring event (started before
+        # today) or one starting past the horizon: fall back to its own start.
+        start = upcoming_occ[0] if upcoming_occ else ev.starts_at
```
PastLivesReviewBot, line 191: WARN (test coverage): this fallback branch (`upcoming_occ` empty → `ev.starts_at`) is untested — the new specs cover the recurring next-occurrence path and the excluded-past-event path, but not the still-running non-recurring case (started yesterday, ends tomorrow) or the past-horizon anchor. Worth an `it_lists_a_still_running_multi_day_event_at_its_own_start` spec, both to lock the behavior and to keep CI's branch-coverage gate happy.

tests/hub/events_tab_parity_spec.py
```diff
@@ -0,0 +1,217 @@
...
+    MembershipPlanFactory()
+    user = User.objects.create_user(username=username, email=f"{username}@example.com", password="pass")
+    member = user.member
+    member.fog_role = Member.FogRole.ADMIN
+    member.save(update_fields=["fog_role"])
+    member.sync_user_permissions()
+    return user
+
+
+def _feed_event(title: str, *, days: int = 3, source: str = "guild", guild=None, url: str = "") -> CalendarEvent:
```
PastLivesReviewBot, line 42: WARN (types): `guild=None` is the one unannotated parameter in the file — `guild: Guild | None = None` (import from `membership.models`) keeps the all-functions-typed rule intact.

templates/hub/partials/_sync_flag.html
```diff
@@ -0,0 +1,11 @@
+{% comment %}
+Compact "on the shared Google Calendar" flag — a small green calendar-check glyph.
+Replaces the wordy "Synced to Google ✓" pill wherever a synced event is listed
+(the calendar list below the grid + the Events tab). The words live in the title /
+aria-label so it stays legible to screen readers and hover.
+{% endcomment %}
+<span class="pl-sync-flag" title="Synced to Google Calendar" aria-label="Synced to Google Calendar">
```
PastLivesReviewBot, line 7: WARN (a11y, deliberate-tradeoff check): the words now live only in `title`/`aria-label`. Screen readers are covered, but `title` is dead on touch — the same reason `_sync_badge__reason` renders failure text inline (the hub.css comment cites FRONTEND Screen C). For the green *synced* state an icon-only glyph is defensible (state, not diagnostics), and the PR body says compact was the ask — flagging only so the touch-invisible tooltip is a conscious choice, particularly on the event edit page where the pill used to be the only textual confirmation.

PastLivesReviewBot (approved):
Independent review of v0.23.7 (Events tab parity + compact sync flag). Verified against the PR head:

- `upcoming_calendar_events()` uses only real APIs (`CommunityEvent.objects.pushed()/published()/upcoming()`, `CalendarEvent.objects.upcoming()`, `occurrences_in`, `absolute_url`) and applies the same `google_ical_uid` echo de-dup as the grid (`_get_calendar_context`), so the parity claim holds for the Community Calendar (guild classes/orientations are guild-page-only, correctly out of scope).
- The union is built in `hub/calendar_entries.py` (service layer) and the view only paginates — fat-models respected. Both querysets carry `select_related`, so the per-row template reads are N+1-free.
- The Google-sync flag stays exactly as gated as the old pill: `sync_flag_visible = _google_sync_enabled() and view_as.is_admin` (env switch AND Site Settings toggle AND admin). No audience broadening.
- Templates: `{% comment %}` used for all multi-line comments, new classes are `pl-` prefixed, the conditional `x-show` correctly avoids an undefined `isActive()` on the Events tab, and all three include sites of `calendar_event_item.html` / `_sync_badge.html` were checked for context.
- Specs: `describe_*`/`it_*` only (no `context_*`), factory-boy traits (`community`, `pushed`) exist, no model/DB mocking; URL names `hub_community_calendar_events` / `hub_event_edit` / `hub_event_delete` all resolve.

Approving with a few non-blocking suggestions inline.

PR #137, opened 2026-07-21
[infra] calendar.pastlives.space → Community Calendar redirect (v0.23.8)

HexagonStorms:
Vanity alias: every request to calendar.pastlives.space 302s to https://members.pastlives.space/calendar/?public=1 (temporary redirect so the target can evolve). Hosts come from CALENDAR_REDIRECT_HOSTS (default calendar.pastlives.space) and are auto-appended to ALLOWED_HOSTS, so go-live needs nothing beyond this deploy — the Render custom domain + DNS are already verified and the cert is issued.

Specs: tests/core/calendar_redirect_host_spec.py (root redirect, deep-path redirect, members host untouched) — 47 surface-area specs pass in docker; mypy clean.

### DIFF
--- a/core/middleware.py
+++ b/core/middleware.py
@@ -53,6 +53,12 @@ class SurfaceMiddleware:
             qs = f"?{request.META['QUERY_STRING']}" if request.META.get("QUERY_STRING") else ""
             return HttpResponsePermanentRedirect(f"https://{settings.MEMBER_HOST}{request.path}{qs}")
 
+        # calendar.pastlives.space is a pure vanity alias: every path 302s to the
+        # community calendar (temporary, not 301, so the target can evolve without
+        # fighting browser caches).
+        if host in set(getattr(settings, "CALENDAR_REDIRECT_HOSTS", [])):
+            return HttpResponseRedirect(f"https://{settings.MEMBER_HOST}/calendar/?public=1")
+
         public_hosts: set[str] = set(getattr(settings, "PUBLIC_HOSTS", []))
         guilds_hosts: set[str] = set(getattr(settings, "GUILDS_HOSTS", []))
         signage_hosts: set[str] = set(getattr(settings, "SIGNAGE_HOSTS", []))
--- a/plfog/settings.py
+++ b/plfog/settings.py
@@ -104,6 +104,15 @@ PUBLIC_ONLY_PATH_PREFIXES: tuple[str, ...] = ("/account/",)
 GUILDS_HOSTS = [
     h.strip().lower() for h in os.environ.get("GUILDS_HOSTS", "guilds.pastlives.app").split(",") if h.strip()
 ]
+# Vanity calendar alias. Every request to a host listed here 302s to the community
+# calendar on the members domain (calendar.pastlives.space — printed on flyers etc.).
+# Hosts are auto-added to ALLOWED_HOSTS so go-live needs only DNS + the Render domain.
+CALENDAR_REDIRECT_HOSTS = [
+    h.strip().lower()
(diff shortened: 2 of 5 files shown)


core/middleware.py
```diff
@@ -53,6 +53,12 @@ def __call__(self, request: HttpRequest) -> HttpResponse:
             qs = f"?{request.META['QUERY_STRING']}" if request.META.get("QUERY_STRING") else ""
             return HttpResponsePermanentRedirect(f"https://{settings.MEMBER_HOST}{request.path}{qs}")
 
+        # calendar.pastlives.space is a pure vanity alias: every path 302s to the
+        # community calendar (temporary, not 301, so the target can evolve without
+        # fighting browser caches).
+        if host in set(getattr(settings, "CALENDAR_REDIRECT_HOSTS", [])):
```
PastLivesReviewBot, line 59: Minor: the `set(...)` wrap buys nothing for a single membership test on a 1-element list — `if host in getattr(settings, "CALENDAR_REDIRECT_HOSTS", []):` reads the same and skips the throwaway set build. (The `public_hosts`/`guilds_hosts` sets below are bound once and reused, so they don't set a precedent for this.) Fine to leave as-is.

core/middleware.py
```diff
@@ -53,6 +53,12 @@ def __call__(self, request: HttpRequest) -> HttpResponse:
...
+            return HttpResponseRedirect(f"https://{settings.MEMBER_HOST}/calendar/?public=1")
```
PastLivesReviewBot, line 60: Non-blocking observation: because the target carries `?public=1`, a *logged-in* member who follows the flyer link gets the logged-out chrome (no sidebar) rather than their normal calendar view. That matches the PR description and the 302 makes it cheap to change later — just flagging it as a deliberate choice in case the intent was "members land on their member calendar." Incoming query strings (e.g. utm tags) are also dropped, which seems right for a vanity alias.

PastLivesReviewBot (approved):
Independent review of the calendar.pastlives.space vanity-host redirect. Verified against the PR head (`origin/feat/calendar-subdomain`):

**Runtime / correctness**
- `settings.MEMBER_HOST` exists (settings.py:69) and `HttpResponseRedirect` is already imported in `core/middleware.py`.
- The redirect sits at the top of `SurfaceMiddleware.__call__`, right after the legacy `pastlives.app` 301 and *before* any signage/guilds/public/members surface routing, so it fires for every path on the vanity host. `SurfaceMiddleware` runs before auth/CSRF in `MIDDLEWARE`, so nothing intercepts it.
- `CALENDAR_REDIRECT_HOSTS` is defined unconditionally in settings and auto-appended to `ALLOWED_HOSTS` with a dedupe guard, so `get_host()` won't 400 the new host. 302 (not 301) is the right call for an evolvable target.
- `?public=1` is a real, existing parameter (`request.GET.public` drives the public chrome in `templates/hub/base.html` / `community_calendar.html`), and `community_calendar` is not login-gated, so the third spec's 200 is legitimate.

**Security**
- No open-redirect surface: the target is built entirely from settings, no request-derived data.
- `ALLOWED_HOSTS` gains only the env-controlled calendar hosts — same pattern as `GUILDS_HOSTS`; no wildcard loosening.

**Standards**
- Specs are proper `describe_*`/`it_*` (no `context_*`), live in the established `tests/core/` tree, and cover root, deep path, and the members host staying untouched. `ruff check` and `ruff format --check` pass on the PR-head files; spec files are mypy-excluded per pyproject, and the middleware/settings changes are typed.
- `VERSION` bumped to 0.23.8 with a matching, member-friendly changelog entry at the top.

No blockers. Two minor non-blocking notes inline.

PR #138, opened 2026-07-21
[discord] Weekly calendar digest + new-event posts to #public-calendar (v0.23.9)

HexagonStorms:
Revives the long-dead **#public-calendar** announcement channel (`1309624926893768854`): FOG now feeds it a Monday-morning week-ahead digest plus instant posts whenever something new lands on the Community Calendar.

_Note: originally slated to be based on #136, but that merged first — this branches straight off `main`, so the diff is just the digest work._

## What's new

**Weekly digest (Mondays ~6 AM PT)**
- One post listing everything on the community calendar for the next 7 days — feed events, community events, AND classes — grouped by day, each item time-stamped and linked to its FOG page, with a "See the full calendar" footer link.
- Built on the exact deduped grid union from #136 (`hub.calendar_entries.upcoming_calendar_events()`): echo-suppressed, recurring collapsed — the digest can never disagree with the calendar.
- Discord embeds, chunked under the 4096-char description cap (and 10-embeds-per-message cap) when a week runs long.
- Never posts an empty digest.

**New-event announcements (every 15-min tick)**
- A new upcoming feed/class `CalendarEvent` row (nightly sync) or a newly published `CommunityEvent` gets one compact embed: title, date/time, link (classes link to their public class page; all links absolute).
- `channel_announced_at` stamps guarantee nothing announces twice; echoes of our own Google-pushed events are excluded (shares `_pushed_event_uids` with the sync); recurring events announce with their *next* occurrence, not a stale anchor.
- Capped at 10 posts per run — overflow is stamped silently and logged, so a backlog can never flood the channel.
- Studio-hours rows are excluded (consistent with the existing "ambient hours are never pushed to Discord" rule).

## Architecture
- `core/integrations/discord_channel.py` — `post_channel_message()` bot-authed client, mirrors `discord_events.py` (fail-loud non-2xx, bounded single 429 retry).
- `hub/discord_calendar_posts.py` — all service logic; `build_weekly_digest_embeds()` is a pure, HTTP-free builder.
- Two thin management commands (`post_weekly_calendar_digest`, `announce_calendar_events`) registered in the scheduled-jobs registry; new `Cadence.WEEKLY` runs Mondays at 13:xx UTC via the existing dispatcher cron (no render.yaml change).
- Config: `SiteConfiguration.discord_calendar_channel_id` + `discord_calendar_posts_enabled`, editable on Site Settings → Calendar tab (form_field/toggle components).
- Migrations: `core/0051` (config fields), `membership/0097` (`channel_announced_at` on CalendarEvent + CommunityEvent, **backfilled to now on all existing rows** with a real reverse) — so the first enabled run announces only genuinely new events.

## Go-live ops
1. Site Settings → Calendar tab: set **Discord calendar channel id** to `1309624926893768854` and flip **Post calendar updates to Discord** ON.
2. The bot needs **Send Messages + Embed Links** in #public-calendar (it's an announcement-type channel — check the channel-level overwrites).
3. Existing events are pre-marked announced by the migration, so only events created after deploy get announced; the first digest goes out the following Monday (or run `post_weekly_calendar_digest` from the Automations tab to post one immediately).

## Testing
- 3983 tests green (full core/hub/membership sweep) incl. 90 in the new/extended spec files; mypy clean (154 files).
- New specs: `tests/hub/discord_calendar_posts_spec.py`, `tests/core/integrations/discord_channel_spec.py`, `tests/core/management/discord_calendar_commands_spec.py`; extended: dispatcher WEEKLY gating, scheduled-jobs registry parity, Site Settings render/save.


hub/discord_calendar_posts.py
```diff
@@ -0,0 +1,291 @@
...
+    channel_id = _posting_channel_id()
+    if not channel_id:
+        return 0
+    now = timezone.now()
+    items = _digest_items(now)
+    if not items:
+        return 0
+    embeds = _embeds_for_items(items, now)
+    for i in range(0, len(embeds), MAX_EMBEDS_PER_MESSAGE):
+        post_channel_message(channel_id, embeds[i : i + MAX_EMBEDS_PER_MESSAGE])
```
PastLivesReviewBot: **BLOCK — Discord's 6,000-char combined embed limit will 400 any multi-embed digest.** Each embed's *description* is correctly capped at 4,096, and this batches ≤10 embeds per message — but Discord also enforces a hard 6,000-character limit on the **combined** title+description content of *all* embeds attached to one message. `_chunk_blocks` only produces a second chunk once the first is near 4,096, so any digest whose total content passes ~6,000 chars (a busy week: ~40–50 linked lines, since each `[Title](https://members.pastlives.space/…)` line runs 80–120 chars) ships 2+ near-full embeds in one message → Discord rejects with a 400 (code 50035), `post_channel_message` raises, and the scheduled run FAILS with no digest posted — precisely on the weeks the digest matters most. The `it_chunks_into_multiple_embeds…` spec can't catch it because the respx mock 200s unconditionally.

Fix: batch greedily under a 6,000-char running total (still ≤10 embeds/message), or simply post one embed per message. Worth a spec asserting no message's summed embed content exceeds 6,000.

hub/discord_calendar_posts.py
```diff
@@ -0,0 +1,291 @@
...
+    from hub.calendar_service import _pushed_event_uids
+
+    channel_id = _posting_channel_id()
+    if not channel_id:
+        return 0
+    now = timezone.now()
+
+    pending: list[tuple[datetime, dict[str, Any], Any]] = []
+    feed_rows = (
+        CalendarEvent.objects.filter(channel_announced_at__isnull=True, start_dt__gt=now)
```
PastLivesReviewBot, line 304: **BLOCK — a single new recurring feed series floods the channel with 10 near-identical posts.** Recurring feed events are materialized as one `CalendarEvent` row *per occurrence* (`recurrence_id`, shared `uid`) across the sync's 180-day horizon — a new weekly series lands ~26 rows, all `channel_announced_at IS NULL`. This query treats each as an independent announcement, so the first run posts `ANNOUNCE_CAP` (10) back-to-back embeds for the *same event* on successive dates and silently stamps the other ~16. Same shape for a new multi-session class from `sync_local_class_events` (one "New class on the calendar" per session). The cap was designed to stop a *backlog* flood, but this is the normal first-real-use path.

Fix: collapse to one announcement per series — e.g. group pending feed rows by `(source, guild, feed, uid)`, announce only the earliest-starting unannounced occurrence, and stamp its siblings in the same pass (mirroring how `_next_community_start` already collapses recurring `CommunityEvent`s). Plus a spec: N occurrence rows sharing a uid → exactly 1 post, all N stamped.

hub/discord_calendar_posts.py
```diff
@@ -0,0 +1,291 @@
...
+        return f"{day} · {_time_of(start)} – {_time_of(end)}"
+    return f"{day} {_time_of(start)} – {local_end.strftime('%A, %B %-d')} {_time_of(end)}"
+
+
+def _digest_items(now: datetime) -> list[Any]:
+    """The next-7-days slice of the calendar union (feed + class + community events)."""
+    from hub.calendar_entries import upcoming_calendar_events
+
+    horizon = now + timedelta(days=DIGEST_WINDOW_DAYS)
+    return [e for e in upcoming_calendar_events() if e.start_dt < horizon and e.end_dt >= now]
```
PastLivesReviewBot: WARN: `upcoming_calendar_events()` has no studio-hours exclusion, so every Monday digest will carry a line per guild's weekly studio-hours row — while `announce_new_events` deliberately excludes `STUDIO_HOURS` under the "ambient hours are never pushed to Discord" rule. If the digest is meant to mirror the grid verbatim that's defensible, but with several guilds running standing hours these lines become recurring noise at the top of every digest; consider filtering `e.community_event.event_type == STUDIO_HOURS` entries here too (or note explicitly that the digest intentionally keeps them).

plfog/version.py
```diff
@@ -2,9 +2,26 @@
 
 from __future__ import annotations
 
-VERSION = "0.23.7"
+VERSION = "0.23.8"
```
PastLivesReviewBot: Note (not a blocker): `0.23.8` collides with open PR #137, which is also stamped 0.23.8. Whichever merges second should renumber on rebase so `announce_release` resolves the right changelog entry.

tests/hub/discord_calendar_posts_spec.py
```diff
@@ -0,0 +1,289 @@
...
+        start = timezone.now()
+        assert dcp._when(start, start + timedelta(days=1), True).endswith("All day")
+
+    def it_spells_out_both_days_for_an_overnight_event():
+        start = timezone.now().replace(hour=22, minute=0, second=0, microsecond=0)
+        end = start + timedelta(hours=12)
+        text = dcp._when(start, end, False)
+        assert timezone.localtime(end).strftime("%A, %B %-d") in text
+
+    def it_falls_back_to_the_anchor_when_no_occurrence_is_in_horizon():
```
PastLivesReviewBot, line 358: WARN (cosmetic): this spec exercises `_next_community_start`, not `_when` — it reads misplaced under `describe_when_formatting`. Consider a `describe_next_community_start` block (or folding it into `describe_announce_new_events`) so the describe tree keeps matching the functions it pins.

PastLivesReviewBot (changes requested):
Independent review of the #public-calendar digest + announcer.

**Verified clean:** all imported names exist and match (`_pushed_event_uids() -> set[str]`, `upcoming_calendar_events() -> list[Any]`, `API_BASE`/`_auth_headers` in `discord_dm`, `MEMBER_BASE_URL`/`DISCORD_BOT_TOKEN` in settings, `hub_community_calendar` URL); both migrations chain off the current heads (core 0050, membership 0096), use `apps.get_model()` + bulk `update()` only, and carry a real reverse; the nightly sync's `update_or_create` omits `channel_announced_at` from `defaults`, so stamps survive re-syncs (no nightly re-announce); the `Cadence.WEEKLY` branch is additive and can't touch DAILY/ALWAYS; the Site Settings section routes through `form_field.html` (checkbox auto-toggles); specs are all `describe_*`/`it_*` with respx (no `context_*`); ruff format + check pass on every new file; token never logged.

**Blocking (2):** the digest batching ignores Discord's 6,000-char combined-embed-content limit per message, so exactly the busy weeks this feature exists for will 400; and the announcer treats every occurrence row of a new recurring feed series (180-day sync horizon, ~26 rows for a weekly event) as a separate announcement — one new weekly series floods the channel with 10 near-identical posts in a single run. Details inline.

**Non-blocking notes** inline: the digest includes studio-hours rows (the announcer deliberately excludes them), the 0.23.8 version collision with open #137, and one misfiled spec.

hub/discord_calendar_posts.py
```diff
@@ -0,0 +1,348 @@
...
+    feed_rows = (
+        CalendarEvent.objects.filter(channel_announced_at__isnull=True, start_dt__gt=now)
+        .exclude(uid__in=_pushed_event_uids())
+        .order_by("start_dt")
+    )
+    # Collapse per-occurrence rows of one series/class into a single announcement:
+    # the query is start-ordered, so each group's first row is its earliest occurrence.
+    series: dict[tuple[str, int | None, str], list[CalendarEvent]] = {}
+    for row in feed_rows:
+        series.setdefault((row.source, row.guild_id, row.uid), []).append(row)
```
PastLivesReviewBot, line 312: Note (non-blocking): this collapse key omits `feed_id`, while the row's natural key is `(guild, feed, uid, recurrence_id)`. Two different general-calendar feeds that ever emitted the same iCal UID would collapse into one announcement (and the second feed's rows would be stamped under the first's post). iCal UIDs are effectively globally unique so this is theoretical — but adding `row.feed_id` to the tuple would make the grouping key match the uniqueness constraint exactly, at zero cost.

hub/discord_calendar_posts.py
```diff
@@ -0,0 +1,348 @@
...
+    today = timezone.localdate(now)
+    horizon = today + timedelta(days=366)
+    duration = event.ends_at - event.starts_at
+    for occ in event.occurrences_in(today, horizon):
+        if occ + duration >= now:
+            return occ
+    return event.starts_at
+
+
+def _stamp(objs: list[Any], now: datetime) -> None:
```
PastLivesReviewBot, line 273: Nit (non-blocking): `_stamp` issues one UPDATE per row — a new weekly series across the 180-day sync horizon is ~26 sequential saves inside the announce loop. A single `type(objs[0]).objects.filter(pk__in=[o.pk for o in objs]).update(channel_announced_at=now)` per group would be one query. Volume is bounded and this runs in a background job, so fine as-is; worth folding in if this file gets touched again.

hub/discord_calendar_posts.py
```diff
@@ -0,0 +1,348 @@
...
+                "title": title if i == 0 else f"{_DIGEST_TITLE} (continued)",
+                "description": chunk,
+                "color": _EMBED_COLOR,
+                "url": _calendar_url(),
+            }
+        )
+    return embeds
+
+
+def _embed_chars(embed: dict[str, Any]) -> int:
```
PastLivesReviewBot, line 194: Note (non-blocking): the title+description accounting is correct for every embed built today (the docstring says so, and `url`/`color` don't count toward Discord's 6,000). If a footer, author, or fields are ever added to these embeds, `_embed_chars` must grow with them or the 6,000 cap can be silently breached again — a one-line comment at the embed-construction site pointing back here wouldn't hurt.

PastLivesReviewBot (approved):
Re-review of the #public-calendar digest + announcer at a366da6 (v0.23.9), following up review 4745941867.

**BLOCK 1 (6,000-char combined embed limit) — verified fixed.** `_batch_embeds()` keeps a running `_embed_chars` total (title + description — these embeds carry no footer/author/field text, and `url`/`color` don't count toward the limit) and closes a batch when the next embed would push past `MESSAGE_EMBED_TOTAL_MAX` **or** the batch already holds `MAX_EMBEDS_PER_MESSAGE`; `post_weekly_digest` posts one message per batch. The spec (`describe_batch_embeds`) builds a real ~9,900-char digest, first proves the scenario genuinely exceeds 6,000, then asserts every batch is ≤6,000 combined chars and ≤10 embeds and that the batches re-concatenate to the original embed list — asserted on the built payload batches, not the HTTP mock, exactly as asked. A lone embed can never itself exceed 6,000 (≤4,096 description + short title), so the always-place-one behavior is safe.

**BLOCK 2 (recurring-series flood) — verified fixed.** The feed query is explicitly `.order_by("start_dt")` (and `CalendarEvent.Meta.ordering = ["start_dt"]` backstops it), so the `setdefault` grouping by `(source, guild_id, uid)` leaves each group insertion-ordered with its earliest upcoming occurrence first; one embed per group at that occurrence; `_stamp(objs)` writes **every** sibling row in both the posted loop and the overflow loop; and `ANNOUNCE_CAP` slices the collapsed `pending` list of groups, not rows. Specs pin all of it: a 5-row shared-uid series posts once with the earliest date and leaves zero unstamped siblings (second run posts 0); two distinct events still announce separately; the 13-event cap test proves overflow ends fully stamped.

**Prior WARNs all addressed:** studio-hours rows are now excluded from the digest via `_is_studio_hours` (specced in `it_leaves_standing_studio_hours_out_of_the_digest`); version renumbered to 0.23.9 with a matching, member-friendly changelog entry (no collision with still-open #137 at 0.23.8); the misfiled `_next_community_start` spec now lives in its own `describe_next_community_start` block.

**Independently re-verified at this head:** ruff format + check clean on all 9 new/changed files; the targeted suite (83 tests across the posts service, channel client, command wrappers, dispatcher WEEKLY gate, scheduled-jobs registry, and Site Settings) passes; mypy clean over `plfog/ core/ membership/ hub/`. Specs are all `describe_*`/`it_*` with respx; errors stay fail-loud (`DiscordChannelError` fails the scheduled run); the migration keeps its real reverse; no N+1 reads (one queryset per source).

Two non-blocking notes inline. Approving.

PR #139, opened 2026-07-21
[discord] Site Settings-managed #important-info post (v0.23.10)

HexagonStorms:
> **Stacked on #138 — merge that first; this diff shrinks to the info-post commits once it lands.**

## What

The locked **#important-info** channel's pinned bot message (the "📌 Past Lives — Important Links" embed + the "🤖 FOG Bot Commands" embed) is now **managed from FOG**. A new *#important-info pinned post* section on Site Settings → Discord lets an admin edit the links content (Discord-flavored markdown) and hit Save — FOG **PATCHes the existing pinned message in place**, so the message id and its pin never change.

- **Links embed** — from the new `SiteConfiguration.discord_info_links_content` (seeded with the current live sections: Classes, Community Calendar, Code of Conduct, Past Lives App). Form-validated under Discord's 4096-char embed cap; a blanked field falls back to the seeded default.
- **Commands embed** — auto-generated from the same slash-command registry (`core.events.discord_commands.all_commands()`) that `register_discord_commands` PUTs to Discord, so a newly registered command appears in the guide on the next admin save. No hardcoded list.
- **Fail loudly** — a Discord failure on save surfaces as an admin-facing `messages.error` ("Settings saved, but updating the pinned #important-info Discord post failed: …") while the local content still saves; hitting Save again retries.
- **No new scheduled job** — sync happens on save only, and only when an info-post field actually changed (an unrelated tab's save never calls Discord).

## How

- `core/integrations/discord_channel.py` — new `edit_channel_message(channel_id, message_id, embeds)` (PATCH `/channels/{id}/messages/{message_id}`), sharing the existing embed-cap check + bounded single 429 retry with `post_channel_message` (refactored into `_send_embeds`). Same `DiscordChannelError` contract.
- `hub/discord_info_post.py` — `build_info_embeds()` (pure: config content + command registry, descriptions clipped defensively at 4096) and `sync_info_post()` (no-op unless both ids set; raises `DiscordChannelError` on failure).
- `core/models.py` — `discord_info_channel_id`, `discord_info_message_id`, `discord_info_links_content` (+ `DISCORD_INFO_LINKS_DEFAULT`); additive migration `core/0052`.
- `hub/forms.py` — the three fields on `SiteSettingsForm`, with `clean_discord_info_links_content` capping at the embed limit.
- `hub/views.py` — `_sync_info_post_if_changed()` called after a successful settings save.
- Template: new card on the Discord tab (all fields via `form_field.html`), excluded from the General auto-loop.

## Go-live ops

1. On production, open **Site Settings → Discord → #important-info pinned post** and set:
   - Discord info channel id: `1122351596400025661`
   - Discord info message id: `1529134555494355147`
2. Hit **Save** — the first save syncs the post (links embed from the seeded content + the auto-generated commands guide). Adjust the links copy and re-save as needed.
3. Requires the bot's *View Channel* + *Send Messages* in #important-info (it authored the message, so edit rights are inherent).

## Tests

- `tests/hub/discord_info_post_spec.py` — builder (config content, default fallback, every registered command listed, embed-cap clipping) + sync (no-op without ids, PATCH path/payload via respx, error propagation).
- `tests/hub/site_settings_discord_info_spec.py` — fields render once with the sync hint, save persists, over-limit content rejected, save triggers sync only when info fields changed, Discord failure → error message + content still saved.
- `tests/core/integrations/discord_channel_spec.py` — extended with `describe_edit_channel_message` (PATCH auth/payload, non-2xx, network error, 429 retry, embed cap).
- Sweep: 1502 passed across site-settings/forms/core/discord specs (the three `nacl`-dependent interaction specs are unrunnable in the local image — pre-existing, CI covers them). mypy clean.


core/integrations/discord_channel.py
```diff
@@ -0,0 +1,90 @@
+"""Post (and edit) channel messages (embeds) as the FOG bot.
+
+The thin sibling of :mod:`core.integrations.discord_events`: the same bot REST auth
+(``bot_token`` / ``_auth_headers`` / ``API_BASE`` from :mod:`core.events.discord_dm`), the
+same fail-loudly error wrapping, and the same bounded single retry on a 429 — always on
+here, because the only callers are background jobs (the weekly calendar digest and the
+15-minute new-event announcer), never an interactive request.
```
PastLivesReviewBot, line 7: **WARN — stale docstring.** This paragraph (and the "Callers gate themselves on `discord_calendar_posts_enabled`…" one below it) is no longer accurate: `edit_channel_message` is called from the interactive Site Settings save, and that caller gates on the info-post ids, not the calendar fields. It also means the shared 429 retry can now `time.sleep()` up to 15s inside a web request. The tradeoff is probably fine for an admin-only save (Discord's `Retry-After` is usually sub-second), but the docstring should acknowledge the interactive caller and that the bounded sleep is accepted deliberately, rather than asserting it can't happen.

hub/discord_info_post.py
```diff
@@ -0,0 +1,96 @@
...
+"""
+
+from __future__ import annotations
+
+from typing import Any
+
+from core.integrations.discord_channel import edit_channel_message
+
+# Discord's hard cap on one embed's description.
+EMBED_DESCRIPTION_MAX = 4096
```
PastLivesReviewBot, line 27: **WARN — duplicated constants.** `EMBED_DESCRIPTION_MAX = 4096` and `_EMBED_COLOR = 0x3D8BD4` are now defined here *and* in `hub/discord_calendar_posts.py` (lines 33/41). Consider hoisting them into `core/integrations/discord_channel.py` next to `MAX_EMBEDS_PER_MESSAGE` — they're properties of the Discord API / FOG's channel-post look, not of either feature, and `hub/forms.py` already reaches into this module just for the cap.

hub/discord_info_post.py
```diff
@@ -0,0 +1,96 @@
...
+    """
+    if len(text) <= EMBED_DESCRIPTION_MAX:
+        return text
+    clipped = ""
+    for line in text.split("\n"):
+        candidate = f"{clipped}\n{line}" if clipped else line
+        if len(candidate) > EMBED_DESCRIPTION_MAX - 2:  # leave room for the ellipsis line
+            break
+        clipped = candidate
+    return f"{clipped}\n…"
```
PastLivesReviewBot, line 66: **WARN — degenerate clip edge.** If the *first* line alone exceeds `EMBED_DESCRIPTION_MAX - 2`, the loop breaks with `clipped == ""` and this returns `"\n…"` — the entire embed content silently vanishes. Only reachable for content saved outside the form (shell/admin) or a pathological command description, but a hard-slice fallback (e.g. `text[:EMBED_DESCRIPTION_MAX - 1] + "…"` when no whole line fits) would keep something visible instead of an effectively empty embed.

hub/discord_info_post.py
```diff
@@ -0,0 +1,96 @@
...
+
+
+def build_info_embeds() -> list[dict[str, Any]]:
+    """The two #important-info embeds: the links embed then the commands guide.
+
+    Pure over the config row + the command registry — no HTTP — and every description
+    fits Discord's 4096-char cap.
+    """
+    return [
```
PastLivesReviewBot, line 75: **WARN — combined per-message cap not enforced.** Each description is clipped at 4,096, but Discord also caps one *message* at 6,000 combined embed characters — the limit #138 handles for the digest via `MESSAGE_EMBED_TOTAL_MAX` in `hub/discord_calendar_posts.py`. A near-max links embed (the form allows exactly 4,096) plus a commands guide over ~1.9k chars would 400 the PATCH. Unreachable with today's six commands, and the failure at least surfaces loudly to the admin, so not blocking — but a shared guard (or at minimum a comment tying the two caps together) would close the gap before the command set grows.

PastLivesReviewBot (approved):
## Review — APPROVE (4 WARNs, no blockers)

Reviewed as the net diff vs current `main` (`git diff origin/main origin/feat/discord-info-post`) since the GitHub compare view still carries the already-merged #138 content via the stale merge-base. The true delta is the info-post work only: the `_send_embeds` refactor + `edit_channel_message` in `core/integrations/discord_channel.py`, `hub/discord_info_post.py`, three `SiteConfiguration` fields + `DISCORD_INFO_LINKS_DEFAULT`, migration `core/0052`, form/view/template wiring, and three spec files.

**Tier 1 — all verified clean:**
- **Runtime**: `core.events.discord_commands.all_commands()` exists and returns `SlashCommand` objects with `.name`/`.description`; the registry is populated by `autodiscover()` in `core/apps.py:37` at startup, so it is live by the time an admin saves. `SiteConfiguration.load()` exists. `hub/views.py` has `from __future__ import annotations`, `messages`, and `SiteSettingsForm` in scope. Patching `hub.discord_info_post.sync_info_post` in the view specs works because the view imports it lazily at call time.
- **`_send_embeds` refactor**: behavior-preserving — same cap check before any HTTP, same single bounded 429 retry, same error contract. Zero deletions in the pre-existing `describe_post_channel_message` specs; they still assert the original contract, and the new `describe_edit_channel_message` block mirrors it for PATCH.
- **Migration 0052**: additive only, depends on `0051` (now on `main` via #138), field definitions match the model exactly. Lint CI is green.
- **Correctness**: `sync_info_post` no-ops unless BOTH ids are set (whitespace handled via `.strip()`); descriptions clipped at 4096; save-path ordering is right — `form.save()` runs before the sync, so content persists even when Discord fails, and the failure surfaces as an admin-facing `messages.error` (spec-covered), never a silent pass. The `changed_data` gate keeps unrelated-tab saves from ever calling Discord (also spec-covered).
- **Security**: bot token only in request headers, never logged; error text is the truncated Discord response body; the view is `@fog_admin_required`; the seeded default content contains no secrets.
- **Architecture / testing / frontend**: logic lives in `hub/discord_info_post.py` with validation in `clean_discord_info_links_content`, the view stays thin; specs are `describe_*`/`it_*` (no `context_*` anywhere), respx for HTTP, no model/DB mocking; the template card sits inside the Discord tab, all three fields render via `form_field.html`, they're excluded from the General auto-loop (render-once spec-covered), and only a single-line `{# #}` comment is used.

**WARNs** (inline below): a now-stale module docstring in `discord_channel.py`; the 6,000-char combined per-message embed cap isn't enforced for the two-embed info post; `EMBED_DESCRIPTION_MAX`/`_EMBED_COLOR` duplicated across the two `hub/discord_*` modules; a degenerate `_clip` edge that can drop all content for a single oversize line.

Note: the `test` CI job was still running at review time (lint + e2e already green); the merge is gated on it regardless.

PR #140, opened 2026-07-21
[discord] #classes channel: weekly class digest + new-class posts (v0.23.11)

HexagonStorms:
**Stacked on #138 → #139 — merge those first.** This diff shrinks to just the classes-channel commits once they land.

Gives the Discord **#classes** channel (`946149249178021949`) the same treatment #138 gave #public-calendar: a Monday-morning digest of the week's classes plus an instant post whenever a new class is published — every link pointing at the public class pages.

## What's new

**Weekly classes digest (Mondays ~6 AM PT)**
- Public classes' sessions in the next 7 days, grouped by local day — time, linked title, and instructor name — plus a "Flexible scheduling — book anytime" section for currently-bookable flexible classes, and a "Browse all classes" footer link (BOOK_BASE_URL).
- Never posts when the week is empty. Chunked under the 4,096-char embed cap and batched under the 6,000-char combined per-message cap (reuses #138's `_chunk_blocks` / `_batch_embeds` — no copy-paste).

**New-class announcements (every 15-min tick)**
- **Source of truth is `ClassOffering`, not the calendar cache**: a "new class" = `ClassOffering.objects.bookable()` (published, non-private, flexible or first session still upcoming) with no `channel_announced_at` stamp. This also catches classes that arrive via the legacy CMS sync, not just the FOG approval flow.
- One compact embed per offering: title, category name, next session date (or the flexible note), price (`cents_as_price`), and a Sign up link to its public class page (`ClassOffering.public_url`, absolute on the book host).
- Stamp is permanent — draft→published→unpublished→republished never re-announces. 10-per-run cap; overflow is stamped silently and logged.
- Complementary to the existing `class_published` spine broadcast (central webhook + in-app bell); this targets #classes specifically via the bot.

## Architecture
- `hub/discord_class_posts.py` — all service logic; pure `build_weekly_classes_digest_embeds()`; shares batching/chunking/gating shape with `hub/discord_calendar_posts.py`.
- Commands `announce_new_classes` (ALWAYS) + `post_weekly_classes_digest` (WEEKLY, Mon 13:xx UTC) registered in `core/scheduled_jobs.py`; existing dispatcher cron covers both.
- Config: `SiteConfiguration.discord_classes_channel_id` + `discord_classes_posts_enabled` (core migration 0053), Site Settings → Calendar tab, "#classes Discord posts" section right under the #public-calendar one (form_field/toggle components).
- `ClassOffering.channel_announced_at` (classes migration 0048) **backfilled to now on every currently-published offering** (real reverse) — enabling never floods ~550 existing classes.

## Go-live ops
1. Site Settings → Calendar tab: set **Discord classes channel id** to `946149249178021949` and flip **Post class updates to Discord** ON.
2. The bot can already read #classes, but **verify it can Send Messages + Embed Links there at go-live** — the channel restricts @everyone; if posting 403s, add the Fog Bot role with Send in that channel.
3. Existing published classes are pre-marked announced by the migration, so only classes published after deploy announce; the first digest posts the following Monday (or run "Weekly classes digest" from the Automations tab).

## Testing
- 2991 green across core/hub/classes sweeps (incl. 100 in the new/extended spec files); CI-scope mypy (`plfog/ core/ membership/ hub/`) clean.
- New specs: `tests/hub/discord_class_posts_spec.py`, `tests/core/management/discord_class_commands_spec.py`; extended: dispatcher WEEKLY/ALWAYS wiring, scheduled-jobs registry parity, Site Settings render/save.


hub/discord_class_posts.py
```diff
@@ -0,0 +1,216 @@
...
+    """The announcement's date line: the next upcoming session, or the flexible note.
+
+    ``bookable()`` guarantees a FIXED offering's first session is still upcoming, so the
+    earliest session IS the next one.
+    """
+    from classes.models import ClassOffering as CO
+
+    if offering.scheduling_model == CO.SchedulingModel.FLEXIBLE:
+        return _FLEXIBLE_WHEN
+    first = offering.sessions.order_by("starts_at").first()
```
PastLivesReviewBot, line 155: **WARN (perf/dead branch):** `bookable()` already annotates every row with `first_session_at` (`Min("sessions__starts_at")`), and its filter guarantees a FIXED offering's value is upcoming — exactly the datum this method wants. Reading `offering.first_session_at` instead of `offering.sessions.order_by("starts_at").first()` would drop up to `ANNOUNCE_CAP` (10) extra queries per run *and* delete the `first is None` defensive branch below, which is unreachable via `bookable()` and — if it ever did fire — would mislabel a sessionless FIXED class with the flexible-scheduling copy. Bounded and on a 15-min cron, so non-blocking.

hub/discord_class_posts.py
```diff
@@ -0,0 +1,216 @@
...
+    return line
+
+
+def _digest_blocks(now: datetime) -> list[str]:
+    """The digest's day blocks (public classes' sessions in the next 7 days, grouped by
+    local day) plus one flexible-scheduling block for currently-bookable flexible classes."""
+    from classes.models import ClassOffering, ClassSession
+
+    horizon = now + timedelta(days=DIGEST_WINDOW_DAYS)
+    public_ids = ClassOffering.objects.public().values_list("pk", flat=True)
```
PastLivesReviewBot, line 74: **WARN (dedup + edge):** this hand-rolls the semantics that already exist as `ClassSessionQuerySet.upcoming_public()` (future sessions whose offering is published + non-private, with a docstring explaining why it's `public()` not `bookable()`). `ClassSession.objects.upcoming_public().filter(starts_at__lt=horizon)` would reuse the canonical definition in one query with no `public_ids` subselect.

Related edge: nothing prevents a FLEXIBLE offering from carrying `ClassSession` rows, and one that does would appear twice in the digest — its sessions in the day blocks *and* the offering in the flexible section. Rare and harmless, but an `exclude(class_offering__scheduling_model=FLEXIBLE)` on the day-block query (or the inverse on the flexible one) would close it.

hub/discord_class_posts.py
```diff
@@ -0,0 +1,216 @@
...
+    sessions = (
+        ClassSession.objects.filter(class_offering_id__in=public_ids, starts_at__gte=now, starts_at__lt=horizon)
+        .select_related("class_offering", "class_offering__instructor")
+        .order_by("starts_at")
+    )
+    grouped: dict[Any, list[ClassSession]] = {}
```
PastLivesReviewBot, line 80: **WARN (typing nit):** the keys here are `datetime.date` objects from `timezone.localtime(...).date()` — `dict[date, list[ClassSession]]` types this precisely instead of `Any` (house rule: type everything; `Any` here also mutes mypy on the `day.strftime` call below).

hub/discord_class_posts.py
```diff
@@ -0,0 +1,216 @@
...
+from __future__ import annotations
+
+import logging
+from datetime import datetime, timedelta
+from typing import TYPE_CHECKING, Any
+
+from django.utils import timezone
+
+from core.integrations.discord_channel import post_channel_message
+from hub.discord_calendar_posts import ANNOUNCE_CAP, DIGEST_WINDOW_DAYS, _batch_embeds, _chunk_blocks, _time_of
```
PastLivesReviewBot, line 26: **WARN (API surface):** reusing `#138`'s batching/chunking instead of copy-pasting is the right call — but `_batch_embeds`, `_chunk_blocks`, `_time_of` (and `classes.emails._absolute_url`, imported at line 44 and again by the new spec) are underscore-private names now consumed across module boundaries. They've become a de-facto shared contract between two sibling modules plus tests; a small follow-up promoting them to public names (or a shared `hub/discord_embeds.py` helper module) would make that contract honest and keep future refactors of the calendar module from silently breaking this one.

PastLivesReviewBot (approved):
Reviewed the true stacked delta (`git diff origin/feat/discord-info-post origin/feat/discord-classes-channel` — 16 files, +765/−6), not the inflated vs-main diff. **Verdict: APPROVE — no blockers, 4 WARNs (all inline, non-blocking follow-ups).**

## Tier 1 — all clear

**Migrations**
- `classes/0048`: `apps.get_model` + raw `status="published"` string + single bulk `.update()` backfill + a real reverse. Dependency `classes/0047_classfaq` is the tip of the classes chain on `main` — verified.
- `core/0053` depends on `core/0052_siteconfig_discord_info_post`, which exists on the parent branch (#139) — the stack order in the PR body (merge #138 → #139 first) is load-bearing and correct.

**Runtime imports** — every cross-module symbol verified present at the base branch head: `_batch_embeds`, `_chunk_blocks`, `_time_of`, `ANNOUNCE_CAP`, `DIGEST_WINDOW_DAYS` in `hub/discord_calendar_posts.py`; `ClassOffering.public_url` (absolute via `classes.emails._absolute_url`), `bookable()`/`public()` semantics match usage exactly (published + non-private + flexible-or-first-session-upcoming, soonest first), `cents_as_price` is a plain importable function, `post_channel_message(channel_id, embeds)` signature matches, `category` is non-null `PROTECT` (safe `.name` access), nullable `instructor` is guarded.

**Correctness**
- Re-announce is impossible: permanent stamp, specced through the full draft→published→unpublished→republished round-trip, and backfilled offerings are specced never to announce.
- 10-cap overflow is fully stamped (specced: zero NULL stamps remain, second run posts 0) and logged.
- Caps: 4,096 per description via `_chunk_blocks` (with the oversize-block line-split fallback), 6,000 combined + 10-embed per message via `_batch_embeds`, with a genuine payload-level spec (`it_chunks_and_batches_a_huge_week_under_the_message_caps` asserts every batch's summed chars and lossless ordering).
- Empty-week suppression correct (blocks checked before the footer is appended). Note the documented consequence: while any bookable flexible class exists, the digest posts every Monday even in sessionless weeks, repeating the flexible section verbatim — reads as intended, just confirming it's a deliberate choice.
- Post-then-stamp per offering means a mid-loop Discord failure (e.g. the 403 the PR body anticipates at go-live) leaves the failing offering unstamped for the next tick, with no double-posts of the already-stamped ones. Good failure shape.

**Security** — bot token flows only through the existing `discord_channel` client, never logged; settings live on the already superuser-gated Site Settings view (render/save/uncheck specced).

**House rules** — all logic in the service module, thin command wrappers; full annotations (one nit inline); `describe_*`/`it_*` only, zero `context_*`; respx against the real client (models/DB never mocked); factory-boy; `form_field` components matching the sibling #public-calendar section; `ruff format --check` + `ruff check` clean under the repo's 120-col config; changelog correctly curated as one 0.23.11 feature entry.

## WARNs (inline)
1. `_class_when` re-queries sessions per announced offering — `bookable()` already annotates `first_session_at`.
2. `_digest_blocks` re-derives `ClassSession.upcoming_public()` semantics by hand; flexible-with-sessions edge double-lists.
3. `grouped: dict[Any, …]` — key is a `date`; type it precisely.
4. Underscore-private imports are now a two-module (plus spec) contract — worth promoting to public names in a follow-up.

Go-live checklist in the PR body is sound (channel id `946149249178021949`, bot Send+Embed perms check, backfill pre-stamps ~550 existing classes).

PR #141, opened 2026-07-21
[hub] Member directory: members-only by default with a public Site Setting + label fixes (v0.23.12)

HexagonStorms:
## The leak

`hub_member_directory` (`/members/`) had no login gate. Anonymous internet visitors could load the full member directory — names, photos, phone numbers, and `mailto:` email links for every listed member. The profile settings page explicitly frames the directory toggle as controlling "what other members see", so this exposed contact info well beyond what members consented to.

## The fix: a Site Setting, members-only by default

The directory was originally intended to be public, so instead of a hard login gate this makes visibility a deliberate admin choice:

- **`core.SiteConfiguration.member_directory_public`** (new boolean, default **off**; migration `core/0054`). Off = sign-in required. On = the original fully public directory.
- **`hub/views.py`** — `member_directory` starts with a single conditional: anonymous viewer + flag off → `redirect_to_login(request.get_full_path())`, the exact redirect `@login_required` would issue (`/accounts/login/?next=/members/`). Anonymous + flag on renders as before the gate. Signed-in members are unaffected either way.
- **Site Settings → Features tab** — "Public member directory" toggle, rendered through the standard `components/form_field.html` auto-toggle like the other feature switches.

## Label fixes

- `templates/hub/user_settings.html` — the directory master toggle's second segment now reads **Hidden / Public** (was Hidden / Listed). Label text only; the Alpine state (`listed` / `forceListed`) is untouched.
- `templates/hub/guild_detail.html` — the guild page sidebar heading is now **Guild Staff** (was "Guild Lead"), matching the staff list it heads. The "Email Guild Lead" button and the office-hours tooltip are unchanged.

## Tests

- `tests/hub/views_spec.py`:
  - `describe_when_anonymous » it_redirects_to_login_by_default` — 302 to `/accounts/login/?next=/members/` with the flag at its default (off)
  - `describe_when_anonymous » describe_with_the_public_directory_setting_on » it_renders_without_signing_in` — flag on → 200 with directory content, no session
  - `it_renders_for_a_signed_in_member` — flag off, signed in → 200 with directory content
- `tests/hub/site_settings_form_spec.py` — `SiteSettingsForm` declares `member_directory_public`
- Targeted run (views, site-settings form/automations, all member_directory specs, page_header, must_be_listed, surface middleware): 190 passed. Full suite green in the local container apart from the known local-only failures (3 PyNaCl spec modules, 2 baked-EMAIL_BACKEND settings tests — both covered in CI). Ruff format/check and mypy clean.

## Version

`0.23.12` with one member-facing changelog entry (directory members-only by default, admin setting to open it up, Hidden/Public label).


hub/views.py
```diff
@@ -299,6 +299,7 @@ def _compute_new_votes_since(since: datetime | None) -> list[VoteStanding]:
     return results
 
 
+@login_required
```
PastLivesReviewBot: Matches the hub-wide pattern (every other member-gated view in this module carries the same decorator), so the redirect target is consistent with the rest of the app. Note `_get_member(request)` below no longer needs to tolerate anonymous users on this path — no change required, just confirming the decorator makes the existing assumptions in this view body safe.

tests/hub/views_spec.py
```diff
@@ -102,9 +102,21 @@ def it_renders_voting_page(client: Client):
 
 @pytest.mark.django_db
 def describe_member_directory():
-    def it_is_accessible_to_anonymous_guests(client: Client):
+    def describe_when_anonymous():
+        def it_redirects_to_login(client: Client):
```
PastLivesReviewBot: Good: `describe_when_anonymous` — not `context_*`, which pytest here silently skips. Pinning the exact `Location` (including `?next=/members/`) also guards the post-login return trip, not just the gate.

PastLivesReviewBot (approved):
Reviewed against the PLFOG standards (CLAUDE.md): surgical and correct.

- **The gate is the right shape.** `@login_required` is the identical decorator every other member-gated hub view uses, so the anonymous redirect lands on the same login flow (`/accounts/login/?next=/members/`) as the rest of the hub. No business logic touched — skinny view stays skinny.
- **Specs follow the house BDD rules.** The new nested block is `describe_when_anonymous` (correctly `describe_*`, not the uncollected `context_*` trap), asserts the exact 302 Location, and the signed-in path asserts 200 plus real directory content. The stale `it_is_accessible_to_anonymous_guests` spec — which was codifying the leak — is gone.
- **Changelog is correctly curated.** This is a fix to live production behavior, so it gets its own member-facing entry; one entry stamped at the new `VERSION` (0.23.12) so `announce_release` resolves. Plain language, no jargon.
- **Label changes are text-only.** Alpine state (`listed`/`forceListed`) untouched in `user_settings.html`; `guild_detail.html` heading only, with the Email Guild Lead button and tooltip left alone as intended.

Checked the blast radius: every existing spec that GETs the directory already logs in (`member_directory_*_spec`, `page_header_spec`, `must_be_listed_spec`), the surface middleware 404s `/members/` on guest/signage surfaces before the view runs, and no spec asserts the old "Listed" or the guild sidebar "Guild Lead" heading text. No blockers.

hub/views.py
```diff
@@ -302,11 +303,17 @@ def _compute_new_votes_since(since: datetime | None) -> list[VoteStanding]:
...
+    (the original open-directory behavior); anonymous visitors are otherwise sent
+    through the same login redirect ``@login_required`` would issue.
+
     Prefetches each member's primary allauth ``EmailAddress`` so
     ``Member.primary_email`` stays O(1) per member instead of firing a query
     on every template access. See the three-email-store note on
     ``Member.primary_email`` and docs/superpowers/specs/2026-04-07-user-email-aliases-design.md.
     """
+    if not request.user.is_authenticated and not SiteConfiguration.load().member_directory_public:
+        return redirect_to_login(request.get_full_path())
```
PastLivesReviewBot, line 316: Good parity move: `redirect_to_login` is exactly what `@login_required` calls under the hood, so anonymous visitors get the identical `?next=`-carrying redirect as the rest of the hub — and the flag check reads as policy, not plumbing. `SiteConfiguration.load()` is one `get_or_create` per anonymous hit, same cost profile as the existing `tab_payments_enabled` gates.

core/models.py
```diff
@@ -278,6 +278,14 @@ class MemberEventPolicy(models.TextChoices):
         verbose_name="Registration-off message",
         help_text="Shown under the disabled Register button when class registration is off.",
     )
+    member_directory_public = models.BooleanField(
```
PastLivesReviewBot, line 281: Field mirrors the neighboring feature switches (default off = safe posture after the leak; the help_text states both states plainly). Since the default preserves the privacy fix, flipping this on is a deliberate admin act — the right failure mode.

tests/hub/views_spec.py
```diff
@@ -102,9 +103,34 @@ def it_renders_voting_page(client: Client):
...
 def describe_member_directory():
-    def it_is_accessible_to_anonymous_guests(client: Client):
+    def describe_when_anonymous():
+        def it_redirects_to_login_by_default(client: Client):
+            response = client.get("/members/")
+            assert response.status_code == 302
+            assert response["Location"] == "/accounts/login/?next=/members/"
+
+        def describe_with_the_public_directory_setting_on():
+            def it_renders_without_signing_in(client: Client):
```
PastLivesReviewBot, line 113: The three-way truth table (anon+off, anon+on, signed-in+off) is the full behavior surface of this gate, and the nested blocks are all `describe_*` so everything actually collects.

PastLivesReviewBot (approved):
Re-reviewed the updated diff (login gate reworked into a Site Setting) against the PLFOG standards. Approving.

- **Design is the right altitude.** `member_directory_public` lives on `SiteConfiguration` mirroring the existing feature-switch fields exactly (BooleanField, `default=False`, `verbose_name`, full `help_text` — CLAUDE.md model rules). Migration `core/0054` is additive with a default, auto-reversible, and correctly numbered off the branch tip (0053).
- **View stays skinny.** One conditional at the top of `member_directory`: anonymous + flag off → `redirect_to_login(request.get_full_path())` — the same call `@login_required` makes internally, so the redirect (`/accounts/login/?next=/members/`) is byte-identical to every other gated hub view. No business logic added; `SiteConfiguration.load()` singleton access matches the existing `tab_payments_enabled` checks in this module.
- **UI follows FRONTEND.md.** The toggle rides the Features tab through `components/form_field.html` (auto-rendered as a pl-toggle), not the legacy raw-checkbox markup — and the docstring on the view explains the gate for the next reader.
- **Specs cover the truth table.** Anonymous+off → exact 302 Location; anonymous+on → 200 with directory content; signed-in+off → 200. All nested blocks are `describe_*` (never the uncollected `context_*`), and `site_settings_form_spec` pins the field onto `SiteSettingsForm.Meta.fields` so it can't silently fall off the form.
- **Changelog updated in place** — still one entry stamped at `VERSION` 0.23.12, reworded for the setting-based behavior. Correct curation per CLAUDE.md.

No blockers.

PR #142, opened 2026-07-21
[discord] /voting all-guild standings + /vote your ballot (v0.23.13)

HexagonStorms:
## What it does

Brings guild-funding voting to the Past Lives Discord — two slash commands, both ephemeral and link-gated:

### `/voting` — the read-only mini-dashboard

- **Live standings for every active guild** as a fixed-width block-char bar graph (bars in inline code so the width holds in Discord's proportional font): medals + bold names for the top three, plain `4.` `5.` numbering after, points per guild — and the zero-point guilds follow, alphabetical, with all-empty bars and `— 0 pts`, so the whole field is always visible.
- **Cycle framing** — the embed title carries the cycle label, the description the closes-on date (both from `get_cycle_context()`), and the footer the weighting rule (values read from `WEIGHTS`, never re-hardcoded).
- **Your ballot** — the member's three ranked choices with their 5/3/2 points and last-updated time, or a "you haven't voted yet" nudge.
- **One link button** (style 5) to the hub voting page in every state.

`requires_link=True`, `ephemeral=True`, `defer=False`, `scope="guild"` — mirrors `/fog-ping`. Unlinked members get the standard connect prompt before the handler runs; any handler exception becomes the standard friendly error reply. No row cap or collapse — ~21 guilds ≈ 1300 chars; `truncate(…, 4096)` stays purely as a safety net.

### `/vote first: second: third:` — cast or change your ballot

Three **required** options, each a dropdown of active guilds (value = slug), derived from `/join-guild`'s `_guild_choices()` — inheriting its 25-choice Discord cap (overflow logged and dropped from the picker, same constraint `/join-guild` already has; those guilds stay votable on the hub page) and its empty-choices guard. `requires_link=True`, `ephemeral=True`, **`defer=True`** — the save makes a synchronous Airtable HTTP call when sync is enabled, and Discord's 3s deadline is not a bet worth making against an external API (the same call `/join-guild` made for its fan-out). To keep the confirmation's link button on the deferred path, `send_followup` now accepts and forwards `components` (mirroring `reply()`), and `_dispatch_deferred` passes them through — pinned by a new dispatcher spec asserting embeds + components survive the followup PATCH payload. `/voting` stays `defer=False` (two cheap reads, no external calls).

**The hub page's exact validate + save path, reused:**
- Validation is the literal `VotePreferenceForm` the voting page POSTs through (active-guild querysets + the three-distinct rule). Unknown/inactive slug → a friendly ephemeral reply naming the bad pick; duplicate guilds → the form's own "Please select three different guilds." Nothing generic, nothing saved.
- The save is a new `VotePreference.objects.cast_ballot(...)` manager method — the view's former inline `update_or_create` lifted into `membership/models.py`, now called by **both** `hub.views.guild_voting` and the Discord handler. Every side effect fires identically on both surfaces: `updated_at`, the Airtable push in `VotePreference.save()` (`sync_vote_to_airtable`), and the `_log_vote_activity` signal (`VOTE_SUBMITTED`/`VOTE_CHANGED`). No new sync behavior invented.
- Success: a confirmation embed — "Your ballot is in — July 2026 ✅" (or "updated"), the closes-on date, the three picks with 5/3/2 pts, a "see the live standings anytime with `/voting`" nudge, and the link button to the page.

The `/vote` handler lives in `hub/discord_commands.py` (next to `/join-guild`) deliberately: parity demands `hub.forms.VotePreferenceForm`, and `membership` must not import `hub`.

### Embed example (`/voting`)

```
Guild funding — July 2026

Your votes decide how the monthly funding pool is split. This cycle closes **July 31, 2026**.

🥇 `████████████` **Fiber Arts** — 42 pts
🥈 `████████░░░░` **Woodshop** — 28 pts
🥉 `█████░░░░░░░` **Ceramics** — 18 pts
`4.` `███░░░░░░░░░` Metals — 11 pts
`5.` `█░░░░░░░░░░░` Print — 4 pts
`░░░░░░░░░░░░` Jewelry — 0 pts
`░░░░░░░░░░░░` Textiles — 0 pts

**Your ballot**
1st — Fiber Arts · 5 pts
2nd — Ceramics · 3 pts
3rd — Woodshop · 2 pts
_Last updated Mon Jul 14, 2:00 PM_

Weighting: 1st = 5 pts · 2nd = 3 pts · 3rd = 2 pts
[Open the voting page]
```

## The refactors

Two lifts, both so one implementation serves the page and Discord:

1. **The tally**: `VoteStanding`, `_compute_live_standings()`, `_compute_new_votes_since()` moved from `hub/views.py` into `membership/vote_calculator.py` as public names — behavior identical. The load-bearing `distinct=True` cross-join comment moves verbatim; the hardcoded `* 5 / * 3 / * 2` now read from `WEIGHTS`; `Guild`/`Count`/`Q` are imported inside the functions so the pure-math module stays ORM-free at import time. `tests/hub/guild_voting_spec.py` renames the call sites with **assertions untouched** — that suite staying green is the refactor's proof. The `plfog/dashboard.py` comment now points at the new home. (Zero-point guilds are appended in the Discord handler only — the shared function is untouched and the hub page keeps dropping them.)
2. **The save**: `guild_voting`'s inline `update_or_create` became `VotePreference.objects.cast_ballot()` (fat models), shared by the page and `/vote`.

## Testing

- `tests/core/events/voting_command_spec.py` — definition flags, bar scaling (12-full leader, 6/6 at 50%, never an all-empty bar for a ranked guild), happy path under frozen time, ballot + no-leak, no-ballot nudge, empty standings, all-guilds render (21 rows, no collapse) and zero-point rows (empty bars, alphabetical, below ranked, inactive excluded), ≤ 4096 backstop assertion.
- `tests/hub/vote_command_spec.py` — definition (linked/ephemeral/**deferred**) + option builder (three required slug pickers), happy first ballot (frozen time), in-place overwrite, duplicate rejection (form's message, nothing saved), unknown-slug and inactive-guild rejection, and a **side-effect parity** case: Airtable sync spy called with the saved preference, `VOTE_SUBMITTED`/`VOTE_CHANGED` activity rows — proof the hub save path is what ran.
- `tests/core/events/discord_commands_spec.py` — new deferred-dispatch case: embeds + components carried through the followup PATCH.
- `tests/hub/guild_voting_spec.py` — fully green with only the import/call-site renames (and the view now saving through `cast_ballot`).
- Full suite + ruff format/check + mypy green.

## Go-live

After deploy, re-run `register_discord_commands` (guild-scoped — registers instantly). It registers both `/voting` and `/vote`; `/guide` picks them up automatically from the registry.

> Originally stacked on #141; after its squash-merge (including the member-directory public-site-setting rework), `origin/main` was merged back into this branch — `member_directory` is byte-identical to main's (in-view `SiteConfiguration.load().member_directory_public` gate, no `@login_required`), and the diff vs main shows only this feature's files + `version.py` + the spec doc.


membership/vote_calculator.py
```diff
@@ -22,6 +25,98 @@
...
+    results: list[VoteStanding] = []
+    for g in guilds:
+        points = g.first * WEIGHTS["1st"] + g.second * WEIGHTS["2nd"] + g.third * WEIGHTS["3rd"]
+        if points > 0:
+            results.append(VoteStanding(guild_name=g.name, total_points=points))
+
+    if not results:
+        return []
+
+    results.sort(key=lambda x: x["total_points"], reverse=True)
```
PastLivesReviewBot, line 70: Non-blocking: `compute_live_standings` and `compute_new_votes_since` share this identical sort-and-scale tail (and the annotate shape). The spec called for a verbatim-in-behavior lift, so keeping the duplication is the right call for this PR — a follow-up could fold the tail into a small private `_rank_and_scale(results)` helper without touching behavior. Same deferral applies to `plfog/dashboard.py`'s own annotate, which the spec explicitly leaves alone.

membership/discord_commands.py
```diff
@@ -428,3 +429,111 @@ def _schedule_options() -> list[dict]:
...
+
+# --- /voting ------------------------------------------------------------------
+
+_BAR_WIDTH = 12
+_STANDINGS_CAP = 15
+_EMBED_DESCRIPTION_LIMIT = 4096
+_MEDALS = ("🥇", "🥈", "🥉")
+
+
+def _bar(bar_pct: float) -> str:
```
PastLivesReviewBot, line 441: Noting the accepted quantization trade-off from the spec so it doesn't read as a bug later: at width 12, a guild at ~96%+ renders the same 12 full blocks as the leader — the `— N pts` suffix is what disambiguates. Fine as designed.

PastLivesReviewBot (approved):
Reviewed against the PLFOG engineering standards (CLAUDE.md) and the 2026-07-21 spec.

**Architecture** — the lift puts the tally math where it belongs (`membership/vote_calculator.py`, next to `WEIGHTS`), the handler stays thin (resolve → domain call → format), and the Discord handler never imports `hub.views`. The load-bearing `distinct=True` cross-join comment moved verbatim, the `* 5 / * 3 / * 2` literals now read from `WEIGHTS`, and ORM imports stay lazy so the pure-math module remains import-light. `hub/views.py` and `tests/hub/guild_voting_spec.py` change only names — assertions untouched and green, which is exactly the proof a behavior-preserving refactor should carry. The `plfog/dashboard.py` comment now points at the new home.

**Handler** — `requires_link/ephemeral/defer/scope` mirror `/fog-ping`; every spec state (happy, no-ballot, empty standings, unlinked, error, overflow) is either built or inherited from `dispatch()`, and each is covered in `tests/core/events/voting_command_spec.py` with `describe_*`/`it_*` only (no uncollected `context_*` blocks), frozen time on the cycle-label assertions. `max(1, …)` in `_bar()` correctly mirrors the hub bar's min-width sliver; the `_STANDINGS_CAP` collapse plus the `truncate(…, 4096)` backstop close the embed-length hole with no silent drops. Copy mirrors the voting page; the footer weighting line is `WEIGHTS`-derived, never re-hardcoded. Full type annotations throughout, `TYPE_CHECKING`-only model imports, no writes or side effects.

**Housekeeping** — VERSION bumped to 0.23.13 with one curated, member-facing changelog entry stamped at the new version; the 0.23.12 entry is untouched below it. Go-live note (`register_discord_commands`) is in the PR body.

Two non-blocking observations inline. Approving.

hub/discord_commands.py
```diff
@@ -211,3 +211,108 @@ def _join_guild(interaction: Interaction, member: Member | None) -> dict:
...
+            "guild_3rd": guilds_by_slug[slugs[2]].pk,
+        }
+    )
+    if not form.is_valid():
+        # With three resolved active guilds the only reachable failure is the distinct-three
+        # rule — surface the form's own message so Discord and the page speak identically.
+        message = " ".join(str(error) for errors in form.errors.values() for error in errors)
+        return reply(f"{message} Nothing was changed — adjust your picks and try again.", ephemeral=True)
+
+    preference, created = VotePreference.objects.cast_ballot(
```
PastLivesReviewBot, line 283: Non-blocking, worth a note for ops: `/vote` is `defer=False`, but `cast_ballot` → `VotePreference.save()` makes a synchronous Airtable HTTP call when `_sync_enabled()`. Airtable is normally well inside Discord's 3s window and the sync swallows its own failures, but if prod ever shows `/vote` timing out, flipping this command to `defer=True` (like `/join-guild`, which defers for its fan-out) is the one-line fix. The ballot row would already be saved either way.

PastLivesReviewBot (approved):
Fresh review of the updated diff (all-guilds standings + the new `/vote` write command), against the PLFOG standards (CLAUDE.md) and the amended spec.

**All-guilds standings** — the cap/collapse is gone and every active guild renders; the zero-point append lives in the Discord handler only, so the shared `compute_live_standings()` (and the hub page with it) is untouched — the right seam. Zero rows are alphabetical, unranked, `░×12`, `— 0 pts`, below the ranked rows; inactive guilds are excluded by the `is_active=True` filter. The `truncate(…, 4096)` backstop is retained as a pure safety net (~21 guilds ≈ 1300 chars). Specs updated to match: 21-rows-no-collapse, zero-point placement/ordering/exclusion, and the ≤4096 assertion.

**`/vote`** — this is the part that most needed scrutiny, and it holds up:
- *Validation parity*: the handler runs the literal `hub.forms.VotePreferenceForm` the page POSTs through — not a re-implementation — so the three-distinct rule and active-guild querysets can never drift. Placing the command in `hub/discord_commands.py` (beside `/join-guild`) is the correct layering call: `membership` never imports `hub`.
- *Save parity*: the view's inline `update_or_create` was lifted into `VotePreference.objects.cast_ballot()` (fat models, skinny views — the view got thinner) and both surfaces call it, so `updated_at`, the Airtable push in `VotePreference.save()`, and the `_log_vote_activity` signal fire identically. No invented sync behavior; `airtable_record_id` handling is untouched.
- *Failure UX*: unknown/inactive slug → a friendly reply naming the bad pick; duplicates → the form's own message. Both save nothing and neither falls through to the generic `error_reply`.
- *Options*: the three pickers derive from `/join-guild`'s `_guild_choices()`, inheriting the 25-choice cap (logged, documented in the docstring) and the empty-choices guard — one DB query for all three, built at serialization time.
- *Tests*: the side-effect parity spec (Airtable spy called with the saved preference + `VOTE_SUBMITTED`/`VOTE_CHANGED` activity rows) is exactly the right proof that the hub path ran. Overwrite case pins one-row semantics; `describe_*`/`it_*` only.

**Housekeeping** — still one changelog entry at 0.23.13, now covering both commands in member language; spec doc updated in place with a dated addendum; mypy clean (the `cast` in `cast_ballot` is a documented django-stubs limitation, not a smell).

One non-blocking observation inline. Approving.

core/events/discord_interactions.py
```diff
@@ -166,20 +166,26 @@ def ack_deferred(interaction_id: str, token: str, *, ephemeral: bool = True) ->
     return False
 
 
-def send_followup(token: str, *, content: str, embeds: list[dict] | None = None) -> bool:
+def send_followup(
```
PastLivesReviewBot, line 169: Good symmetry: the `components` handling mirrors `reply()` exactly (omitted unless provided), so every existing deferred command's PATCH payload is unchanged. The docstring naming why the kwarg exists (link buttons surviving the followup path) is the kind of comment that stays useful.

hub/discord_commands.py
```diff
@@ -211,3 +211,112 @@ def _join_guild(interaction: Interaction, member: Member | None) -> dict:
...
+
+
+VOTE = SlashCommand(
+    name="vote",
+    description="Cast or change your three ranked guild-funding choices.",
+    handler=_vote,
+    options_builder=_ballot_options,
+    requires_link=True,
+    ephemeral=True,
+    defer=True,
```
PastLivesReviewBot, line 318: This resolves the defer-vs-Airtable note from my previous review. One behavioral nuance worth knowing (not a change request): on the deferred path a handler exception becomes the error followup *after* the ack, so a member always sees either the confirmation embed or the friendly error — never Discord's raw "application did not respond", which is exactly the improvement over racing the 3s deadline.

PastLivesReviewBot (approved):
Fresh review of the updated diff — the merge with `main` and the `/vote` defer change — against the PLFOG standards (CLAUDE.md).

**Merge with main** — verified the risky part explicitly: `hub/views.py`'s `member_directory` is byte-identical to main's rework (the in-view `SiteConfiguration.load().member_directory_public` gate; the stale `@login_required` the auto-merge tried to keep from the pre-rework parent was correctly dropped). The diff vs `origin/main` contains only this feature's files + `version.py` + the spec doc; the #141-rework files (`core/models.py`, migration 0054, `hub/forms.py`, the site-settings template and its specs) came through from main untouched. `version.py` keeps `VERSION = "0.23.13"` with the voting entry on top of **main's** reworked 0.23.12 wording — the changelog rule (one entry per feature, `announce_release` resolves at current VERSION) holds.

**`/vote` defer** — the right call, closing my previous review's one open note: the save path makes a synchronous Airtable HTTP call inside `VotePreference.save()`, and `/join-guild` already established defer as the pattern for handlers with external calls. The supporting change is minimal and symmetric: `send_followup` gains a `components` kwarg handled exactly like `reply()` handles it (included only when provided — commands that never pass components serialize byte-for-byte as before), and `_dispatch_deferred` forwards it. The new dispatcher spec pins embeds **and** components surviving the followup PATCH payload — precisely the regression that would otherwise have silently eaten `/vote`'s link button. `/voting` correctly stays `defer=False` (two aggregate reads, no external calls). Spec doc addendum and the `/vote` definition spec updated to match.

No blockers. Approving.

PR #143, opened 2026-07-21
[classes] Class sales — flat/% off + banner (v0.23.14)

HexagonStorms:
Builds the class-sale feature from the approved spec (`docs/superpowers/plans/2026-07-21-class-sale.md`, committed with this PR). Stacked on #142 (`feat/discord-voting-command`).

## What's in it

**Instructors/admins** get a grouped **Sale** section on the class edit form (both the teach and admin twins), right under the pricing fields:

- `sale_enabled` toggle + `sale_kind` (percent off / dollar amount off) + the matching amount
- Editable banner headline, pre-filled with a catchy default (blank falls back to the default — never blocks save)
- `sale_allow_discount_codes` toggle, **off by default** ("can't be combined with other offers")

**Public surfaces** (all struck-original + sale-price, both themes):

- Catalog card: red-accented **Sale** badge + struck price; member line computed off the sale price
- Detail page: animated shimmer announcement banner (`prefers-reduced-motion` → static) + struck rail price + savings note
- Register page: struck summary price, sale price on the submit button, and — when codes are blocked — the code box is **removed** and replaced with an up-front "Sale price applied — discount codes can't be combined with this offer." note (never an after-submit rejection)
- Stripe receipt: product name gains a `" (Sale)"` suffix; the charged amount is the computed sale total

## Stacking rules (price pipeline)

Order is **base → sale → member discount → coupon**:

- Member auto-discount always computes off the **sale** price.
- When a sale is active and `sale_allow_discount_codes` is **off**: the discount-code field is popped in `__init__`, auto-apply prefill is skipped, and any raced-in submitted code is silently ignored (buyer still gets the sale price).
- When stacking is **on**: typed and auto-apply codes stack on top of the sale (auto-apply picks the best code against the sale base).
- Authoring-time guard: a sale that would land the public price in the 1–49¢ Stripe dead-zone is rejected on the edit form; at register time the existing $0.50-floor error and exact-$0 free-confirm path behave as before.

Validation lives in a shared `_SaleMixin` on both class forms; errors attach to visible non-toggle fields (the `sale_enabled` toggle renders no errors), and the teach form now renders a `non_field_errors` block like the admin twin.

## Data model

Six additive `ClassOffering` fields + `SaleKind` choices + four read-only properties (`sale_is_active`, `sale_price_cents`, `sale_savings_display`, `sale_banner_display`). One plain `AddField` migration (`0049_classoffering_sale_fields`, auto-reversible). Disabling a sale preserves its dormant settings for later re-enable.

## Screenshots

N/A — local visual review to follow (seed script below).

## Seed script for visual review

```
docker compose exec web python scripts/dev/seed_sale_demo.py
```

Idempotent: reports the existing sale if one is already active; otherwise puts the first published paid class (creates a demo class only if none exists) on a **25%-off** sale with the default banner, and prints the public URL.

## Tests

- `class_offering_spec.py`: sale property matrix (active/inactive, percent/fixed pricing, savings display, banner fallback)
- New `class_sale_form_spec.py`: `_SaleMixin` validation on **both** forms (kind+amount required, percent 1–99, fixed < price, free-class conflict on `price_cents`, Stripe-floor authoring rejection, banner default, stacking toggle round-trip, dormant settings preserved)
- `registration_form_spec.py`: full stacking matrix incl. sale-OFF regression and auto-apply-off-the-sale-base (two-code disambiguation)
- `register_spec.py`: on-sale GET markup, checkout `amount_cents` + `" (Sale)"` product name, $0-total free-confirm path
- `member_pricing_copy_spec.py`: markup assertions for badge/struck prices/banner on catalog, detail, register

Version → **0.23.14** with one member-facing changelog entry.


templates/classes/public/register.html
```diff
@@ -111,11 +116,15 @@
...
             {{ form.looking_for }}
           </div>
+          {% if form.discount_code %}
           <div class="reg-field">
             <label for="{{ form.discount_code.id_for_label }}">Discount code (optional)</label>
             {{ form.discount_code }}
             {% if form.discount_code.errors %}<ul class="reg-errors">{% for e in form.discount_code.errors %}<li>{{ e }}</li>{% endfor %}</ul>{% endif %}
           </div>
+          {% elif offering.sale_is_active %}
+          <div class="reg-sale-note">Sale price applied — discount codes can't be combined with this offer.</div>
```
PastLivesReviewBot, line 126: Non-blocking: on a sold-out on-sale class the waitlist form also lacks `form.discount_code` (the waitlist pop), so this sale note renders there too. Harmless — no money moves on a waitlist signup — but if it ever reads oddly, gate the `elif` on `and not is_waitlist`.

classes/forms.py
```diff
@@ -681,6 +779,12 @@ def __init__(
             # Waitlist signups don't transact money so the discount field is
             # noise on the form. Drop it so the registrant isn't confused.
             self.fields.pop("discount_code", None)
+        # A non-stacking sale can't be combined with discount codes — drop the
+        # code box entirely so the buyer is told up-front on the page, never
+        # rejected after submit.
+        self.sale_blocks_codes = offering.sale_is_active and not offering.sale_allow_discount_codes
```
PastLivesReviewBot, line 785: Good: this also covers the benign race from spec §10 — a code typed before the instructor flipped a non-stacking sale is silently ignored by `compute_final_price_cents` rather than rejected, and the buyer still gets the sale price.

classes/spec/forms/class_sale_form_spec.py
```diff
@@ -0,0 +1,182 @@
...
+    return _base_data(instructor=str(InstructorFactory().pk), is_private="", private_for_name="", **overrides)
+
+
+def _form(form_class, **overrides):
+    data = _admin_data(**overrides) if form_class is ClassOfferingForm else _base_data(**overrides)
+    return form_class(data=data)
+
+
+@pytest.fixture(params=[TeachClassOfferingForm, ClassOfferingForm], ids=["teach", "admin"])
+def form_class(request):
```
PastLivesReviewBot, line 53: Nice touch parametrizing the whole suite over both form classes — keeps the teach/admin twins validating identically, which is exactly the drift `_SaleMixin` exists to prevent. Note `sale_kind` is now a required select, so the pre-existing class-form POST fixtures were correctly extended with `"sale_kind": "percent"` (real submits always include the select).

PastLivesReviewBot (approved):
Reviewed the diff against CLAUDE.md and FRONTEND.md.

**Standards:** fat-forms/skinny-views held (all sale validation in the shared `_SaleMixin`, pricing math on model properties, views only glue); `help_text` on all six new fields; additive auto-reversible migration, ruff-formatted; BDD `describe_*`/`it_*` specs only (no `context_*`); multi-line template comments use `{% comment %}`; new hub CSS uses the `pl-` prefix and theme tokens only; the shimmer banner has a `prefers-reduced-motion` fallback; errors attach to visible non-toggle fields and the teach template gained the `non_field_errors` block the admin twin already had.

**Spec conformance:** discount order base → sale → member → coupon; code box popped in `__init__` with the up-front note (never an after-submit rejection); auto-apply skipped/re-based per spec; `" (Sale)"` Stripe suffix; all three `member_price_cents` context sites (detail, register, preview) sale-aware; stacking matrix incl. the sale-OFF regression and the auto-apply-off-sale-base disambiguation cells covered.

Approving — two non-blocking notes inline.

PR #144, opened 2026-07-21
[tools] copy-review: every email, rendered + triggers (v0.23.15)

HexagonStorms:
## What this is

`copy-review.pastlives.space` now publishes the **Email Copy Gallery** instead of the CMS page-screenshot contact sheet: every email the app can send, rendered through its **real send-path renderer** with seeded fake data, on one static page. Built per the spec in `docs/superpowers/plans/2026-07-21-copy-review-email-gallery.md`.

**42 email cards across 9 sections**, each with: the rendered subject in an inbox row, audience, a plain-language trigger note ("Sent when… Goes to…"), a channels badge, a "where to edit" pointer (Site Settings vs. template path), the full branded email in a `srcdoc` iframe (sandboxed, auto-sized, CSS max-height fallback), and the plain-text variant behind `<details>`. A closing **"No email is sent"** section lists all 14 events with EMAIL off/absent, by name + reason, so the index is provably complete. Header stamps count, build date, `vVERSION`, and the short git SHA.

Both families render:
- **Spine events** via `resolved_copy → render_copy → wrap_email_html` with `sample_context_for` — the exact `emit()` path. Structural-template dedup (`_STRUCTURAL_EVENT_KEYS`) is checked **before** the EMAIL-off classification, so `tab_charge_failed` (member email OFF) correctly renders as the `charge_failed_admin` structural card.
- **Structural template pairs + specials** via an explicit registry: the 23 `.txt`/`.html` pairs (rendered with `email_shell_message`, byte-faithful to sends), the instructor welcome (`_welcome_email_bodies`, with a real authored sample body), the release email (`render_release_email`), the announcement composer (`build_announcement_email_html`), the allauth trio (rendered through the real adapter's `render_mail`), and the plain-text `find_account` login-link email (INLINE_STRING, text shown directly).

## The two guards (normal CI, not gated)

`tests/core/email_gallery_completeness_spec.py` runs in the default suite:

1. **Template pairs** — `discover_template_pairs()` (filesystem scan of the four email template dirs) must be ⊆ the registry. Add `new_thing.{txt,html}` without registering it → CI fails. The discovered set of 23 is pinned; shells/partials (`_base`, `notification_shell`, …) are asserted excluded.
2. **Inline sends** — an AST lint over every direct `core.email.send` call site in the app tree collects `trigger_kind` literals; each must be registered or allowlisted with a reason (welcome-test, class blasts, release/announcement test-sends, beta feedback). Computed-kind sites (the allauth adapter) sit in a separate dynamic allowlist. A new inline send → CI fails.

Any render failure fails the build loudly (per-email ok/FAILED log, then re-raise) — the site is never republished with a broken card.

## R2 feature-shot pipeline retained

`copy-review.yml` was the **only** workflow setting `CAPTURE_SCREENSHOTS`, so dropping it would have silently killed the release-email R2 screenshot pipeline. It is **rehomed as its own step** in the same job: Playwright stays installed, and `pytest tests/e2e/screenshots_spec.py::describe_feature_screenshots` runs with `CAPTURE_SCREENSHOTS=1` + the R2 secrets (`continue-on-error` so a shot hiccup never blocks the copy-review publish). The node-id scoping means `describe_cms_screenshots` no longer runs in CI — the page gallery stops being published, exactly as intended — while `scripts/capture-cms-screenshots.sh` still works locally.

## Local preview

```
scripts/build-email-gallery.sh          # writes email-gallery/index.html
# or, in docker:
docker run --rm -v "$PWD":/app -e DATABASE_URL= -e BUILD_EMAIL_GALLERY=1 \
  -e SHOT_DIR=/app/_gallery_out plfog-web:calfix \
  pytest tests/e2e/email_gallery_spec.py -m e2e --no-cov -q
```

Browserless; the transactional `db` fixture rolls back, so it never touches dev data. The page opens fine straight off `file://` (srcdoc iframes are inline).

## Versioning

`VERSION` → 0.23.15. Internal tool → no new changelog entry; the sale entry is re-stamped to 0.23.15 so the top entry's version matches `VERSION` (content unchanged).

## Notes

- The spec's "five inline send sites" table predated a few sends on this branch — the AST lint found and dispositioned all nine literal kinds plus the adapter's computed kinds (see the pinning test).
- The two `site_announcement` management-command sends are spine `Message` overrides, not direct sends — the lint correctly excludes them; their copy is the release/announcement cards.


tests/e2e/email_gallery/registry.py
```diff
@@ -0,0 +1,819 @@
...
+            else:
+                pairs.add(base)
+    return pairs
+
+
+def registered_template_pairs() -> set[str]:
+    """The template-pair keys the registry covers (specials have no discoverable pair)."""
+    keys: set[str] = set()
+    for entry in STRUCTURAL_EMAILS:
+        if entry.renderer in (Renderer.SHELL_TEMPLATE, Renderer.WELCOME) and entry.html_template:
```
PastLivesReviewBot: Nit: the `Renderer.WELCOME` half of this membership test is dead — the WELCOME entry carries no `html_template`, so it always falls through to the `elif` below. `if entry.renderer is Renderer.SHELL_TEMPLATE and entry.html_template:` would say what actually happens. Non-blocking.

tests/e2e/email_gallery_spec.py
```diff
@@ -0,0 +1,126 @@
...
+
+        def it_renders_the_find_account_email_text_only(data):
+            from tests.e2e.email_gallery.build import render_one
+
+            rendered = render_one(_by_key("find_account"), data)
+            assert rendered.subject == "Your Past Lives Account"
+            assert rendered.html == ""  # plain-text only — the card shows the text directly
+            assert "robin.vale@example.com" in rendered.text
+
+    def it_dedups_shell_backed_events(db):
```
PastLivesReviewBot: Nit: this test only reads registry data structures — the `db` fixture is unused and could be dropped. Non-blocking.

PastLivesReviewBot (approved):
Reviewed against CLAUDE.md and the spec (docs/superpowers/plans/2026-07-21-copy-review-email-gallery.md). Approving.

**What's right:**
- Every card renders through the *real* send-path renderer (`email_shell_message`, `_welcome_email_bodies`, `render_release_email`, `build_announcement_email_html`, the allauth adapter's `render_mail`, and the emit-path `resolved_copy → render_copy → wrap_email_html`) — the gallery can't drift from production output.
- The M2 landmine is handled correctly: `_STRUCTURAL_EVENT_KEYS` membership is tested before the EMAIL-off classification, and a dedicated spec pins `tab_charge_failed` → `charge_failed_admin`.
- Both guards live in `tests/core/` (normal suite, no gate) and fail loudly with actionable messages; the discovered-pair set and the inline trigger_kind set are pinned, so discovery-rule drift is also caught. The AST-based inline-send lint is more honest than the spec's grep sketch — it found all nine literal kinds plus the adapter's computed kinds, and correctly excludes the spine's `Message(trigger_kind=…)` overrides.
- B1 respected: the workflow keeps Playwright and rehomes `describe_feature_screenshots` as its own node-id-scoped step with the R2 env and `continue-on-error`, so the copy-review publish can't be blocked by a shot hiccup and the CMS page capture stops publishing exactly as specified.
- Render failure = red build (`ok/FAILED` line then re-raise) — no silently broken cards.
- Fail-loud dict access (`built["subject"]`), BDD `describe_*`/`it_*` naming (no `context_*` blocks), spec files in the right trees, and the version re-stamp follows the changelog house rules.

Two non-blocking nits inline.

PastLivesReviewBot (approved):
Both inline nits addressed in 8b00ca0 (dead WELCOME branch simplified, unused db fixture dropped); guards and gallery specs re-verified green. Approval stands on the new head.

PR #146, opened 2026-07-22
[discord] /voting replies are public; ballots stay private (v0.23.17)

HexagonStorms:
## What it does

Per Josh: `/voting` replies are now **public** — visible to everyone in the channel — so the guild-funding race can be checked (and pointed at) in the open. `/vote` is untouched: still ephemeral and deferred, still the private path for a member's own picks.

- `VOTING` SlashCommand: `ephemeral=False` (handler reply + definition).
- **Privacy follows the visibility change**: since the embed now broadcasts, the personal ballot block ("Your ballot" — the member's three ranked picks and last-updated time) is **removed** from `/voting` entirely. In its place, one line: *"Set or change your own picks with `/vote` — only you see that reply."* A member's ballot is never echoed to the channel.
- Kept unchanged: all-guilds standings (ranked bars + zero-point rows), cycle label + closes-on date, `WEIGHTS`-derived weighting footer, and the style-5 link button to the voting page.

### Reply example (public)

```
Guild funding — July 2026

Your votes decide how the monthly funding pool is split. This cycle closes **July 31, 2026**.

🥇 `████████████` **Fiber Arts** — 42 pts
🥈 `████████░░░░` **Woodshop** — 28 pts
🥉 `█████░░░░░░░` **Ceramics** — 18 pts
`4.` `███░░░░░░░░░` Metals — 11 pts
`░░░░░░░░░░░░` Jewelry — 0 pts

Set or change your own picks with /vote — only you see that reply.

Weighting: 1st = 5 pts · 2nd = 3 pts · 3rd = 2 pts
[Open the voting page]
```

## Testing

`tests/core/events/voting_command_spec.py` updated:
- Definition contract: `(requires_link, ephemeral, defer, scope) == (True, False, False, "guild")`.
- Happy path asserts `flags == 0` (no ephemeral flag) and the `/vote` nudge.
- New broadcast-privacy spec: a member **with** a ballot gets a public reply containing **no** "Your ballot" section, no `1st —` rows, no last-updated line — only the aggregate standings and the nudge.
- The other-member's-ballot case reworked for the same contract; the obsolete per-member ballot and no-ballot-state specs are removed (the embed no longer varies by the caller's ballot).
- Changelog-dependent suites (release announcement/email, context processors, Discord notify) sanity-run green against the re-stamped entry.

Targeted specs (voting, vote, guild_voting, dispatcher): 100 passed. Changelog-dependent suites: 112 passed. `ruff format --check` + `ruff check` + mypy clean.

## Version / changelog

`VERSION = "0.23.17"`. Per the house rule this is an intra-cycle refinement of the `/voting`+`/vote` feature: the existing entry was **updated in place** (folded in "the standings reply is visible to everyone in the channel; your own picks stay private via /vote"), re-stamped to 0.23.17, and moved to the top. No new entry.

## Go-live

**No `register_discord_commands` re-run needed** — ephemeral is a response flag chosen at reply time, not a registration property; this takes effect on deploy. (The command's description text was also tightened to drop "and your ballot"; that cosmetic picker tweak simply lands whenever registration next runs for any other reason.)

### DIFF
--- a/membership/discord_commands.py
+++ b/membership/discord_commands.py
@@ -26,7 +26,7 @@ from core.events.discord_replies import (
 )
 
 if TYPE_CHECKING:
-    from membership.models import Guild, GuildOrientationSettings, Member, MemberQuerySet, VotePreference
+    from membership.models import Guild, GuildOrientationSettings, Member, MemberQuerySet
     from membership.vote_calculator import VoteStanding
 
 logger = logging.getLogger(__name__)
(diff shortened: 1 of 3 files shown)


membership/discord_commands.py
```diff
@@ -476,39 +476,20 @@ def _standings_block(standings: list[VoteStanding], zero_point_names: list[str])
...
-    return (
-        "**Your ballot**\n"
-        f"1st — {preference.guild_1st.name} · {WEIGHTS['1st']} pts\n"
-        f"2nd — {preference.guild_2nd.name} · {WEIGHTS['2nd']} pts\n"
-        f"3rd — {preference.guild_3rd.name} · {WEIGHTS['3rd']} pts\n"
-        f"_Last updated {format_local(preference.updated_at)}_"
-    )
-
-
 def _voting(interaction: Interaction, member: Member | None) -> dict:
```
PastLivesReviewBot, line 479: Note (no change needed): `requires_link=True` is retained even though the public standings carry no personal data. That's a defensible product choice — it keeps `/voting` a members' command and the connect-prompt funnel intact — but if Josh ever wants guests in the channel to run it, dropping the flag is now safe precisely because this PR removed all per-member content from the reply.

PastLivesReviewBot (approved):
Reviewed against the PLFOG standards (CLAUDE.md). Small, scoped, and the privacy reasoning is airtight.

**The key judgment is right**: flipping `ephemeral=False` without touching the embed would have broadcast every caller's three ranked picks to the channel — this PR removes the ballot block *in the same change* that makes the reply public, and replaces it with a nudge that names the private path (`/vote`, still ephemeral + deferred). The handler docstring states the invariant explicitly (public reply ⇒ no personal data), which is exactly where that constraint should live so a future edit can't miss it. Dead code cleaned up properly: `_ballot_block`, the `preference` read, the now-unneeded `member` cast, and the `VotePreference` TYPE_CHECKING import are all gone, not orphaned.

**Specs match the new contract**: the definition case pins `(True, False, False, "guild")` with a comment tying `ephemeral=False` to the no-personal-data rule; the new broadcast-privacy spec exercises the dangerous case (a member WITH a ballot) and asserts no "Your ballot" section, no `1st —` rows, no last-updated line, `flags == 0`; the other-member case is reworked rather than deleted. Removing the per-ballot-state specs is correct — the embed genuinely no longer varies by the caller's ballot. The `_VOTE_NUDGE` module constant keeps copy assertions in one place.

**Changelog handling follows the house rule**: intra-cycle refinement of the /voting+/vote feature → the existing entry updated in place, re-stamped 0.23.17, moved to the top, `VERSION` matches — `announce_release` resolves. No second entry. The changelog-dependent suites (release announcement/email, context processors, Discord notify) were sanity-run green against the re-stamped entry, which is the right paranoia given the what's-new widget echoes this text.

**Go-live note is accurate**: ephemeral is a per-response flag (type-4 `data.flags`), not part of the registered command object, so no `register_discord_commands` re-run is required; the tightened description text is correctly called out as a cosmetic tweak that lands at the next registration run.

One inline nit-level observation. Approving.

PR #147, opened 2026-07-22
[discord] DM nudge for unlinked guild-choosers (v0.23.18)

HexagonStorms:
## What

If a Discord user reacts in **#choose-your-guild** but has no linked Past Lives account, the two-way guild sync used to skip them silently — their guild pick simply never counted. Now the 15-minute reconcile sends that reactor a **one-time DM** telling them they're halfway there, with the one-click account-link URL.

## How

- **Hook point:** `membership/discord_sync.py::reconcile_reactions` — the adds loop only ever matches `Member.objects.filter(discord_user_id__in=ids)`, so unlinked reactors fell through. After the membership add/remove pass, `_nudge_unlinked_reactors()` diffs all reacted ids against linked members and prior nudges.
- **One-time guard:** new `DiscordLinkNudge` model (unique `discord_user_id` + `created_at`, migration 0098). A row means handled forever — including link-then-unlink later, and users whose DMs are closed.
- **DM plumbing:** new `send_dm_text()` in `core/events/discord_dm.py` — same two-call flow as `post_dm` (open DM channel → post message) but *loud*: returns `False` only on Discord 403 + code 50007 (DMs closed → logged, marked nudged, never retried); any other non-2xx raises `httpx.HTTPStatusError`, network errors propagate. The cron surfaces the failure and un-marked ids retry next tick.
- **Bot reactors:** `fetch_reactors` now skips users with `"bot": true` — a bot seeding the reaction emojis can never be a member, and DM-ing one from the nudge would be noise at best.
- **Copy accuracy:** the DM promises the existing reaction counts automatically after linking — verified true: `link_and_import` runs `import_member_guilds`, which mirrors the standing reaction immediately.
- `sync_discord_guild_roles` stdout now includes the nudged count; `ReconcileStats` gains `nudged`.

## DM copy

> You're halfway there! 🧩 You picked a guild in #choose-your-guild, but your Discord isn't linked to your Past Lives account yet — so your pick doesn't count until you finish.
>
> Link up here: \<MEMBER_BASE_URL\>/discord/link/
>
> Once you're linked, your existing reaction counts automatically — no need to re-react.

## Tests

- `tests/membership/discord_sync_spec.py`: nudge DM + row for an unlinked reactor (absolute URL asserted), no duplicate DM on a second run, linked reactor never nudged, 50007 swallowed + marked, other DM errors raise + stay unmarked, no nudging from `import_member_guilds`.
- `tests/core/events/discord_dm_spec.py`: `send_dm_text` happy path, closed-DMs on either call, and every loud-failure branch (respx).
- `tests/core/events/discord_reactions_spec.py`: bot reactors skipped.
- Full suite: 6198 passed, coverage 98.10% (only the 2 known local EMAIL_BACKEND settings failures; PyNaCl interaction specs pass in CI). ruff + mypy clean.

## Version

`0.23.18` + one member-facing changelog entry.


core/events/discord_dm.py
```diff
@@ -125,3 +127,58 @@ def post_dm(discord_user_id: str, message: Message) -> bool:
...
+    response = httpx.post(
+        f"{API_BASE}/users/@me/channels",
+        json={"recipient_id": discord_user_id},
+        headers=_auth_headers(),
+        timeout=_DEFAULT_TIMEOUT_SECONDS,
+    )
+    if _dms_closed(response):
+        logger.info("Discord DM to %s undeliverable (DMs closed).", discord_user_id)
+        return False
+    response.raise_for_status()
```
PastLivesReviewBot, line 172: WARN (advisory): unlike `fetch_reactors`, this loud sender has no 429 Retry-After handling — a rate-limit raises `HTTPStatusError` and aborts the remaining nudges for this tick. That's self-limiting (unmarked ids retry next 15-min tick, and the abort throttles the send rate), so it's acceptable, but if the first production run trips 429s repeatedly, honoring Retry-After for a bounded number of retries here would smooth it out.

membership/discord_sync.py
```diff
@@ -199,7 +202,55 @@ def reconcile_reactions() -> ReconcileStats:
...
+    from membership.models import DiscordLinkNudge, Member
+
+    if not reactor_ids:
+        return 0
+    linked = set(Member.objects.filter(discord_user_id__in=reactor_ids).values_list("discord_user_id", flat=True))
+    already = set(
+        DiscordLinkNudge.objects.filter(discord_user_id__in=reactor_ids).values_list("discord_user_id", flat=True)
+    )
+    nudged = 0
+    for discord_user_id in sorted(reactor_ids - linked - already):
```
PastLivesReviewBot, line 248: WARN (advisory): on the very first production tick this will DM every historic unlinked reactor currently on the role message in one burst. That is the intended feature, and a mid-burst failure resumes cleanly next tick thanks to the per-id marking, but it's worth knowing at go-live so a flurry of bot DMs (and any member questions) isn't a surprise.

PastLivesReviewBot (approved):
Reviewed against the two-tier checklist. Solid work: the hook point is exactly where unlinked reactors were being dropped, the one-time guard is a proper unique-constrained model with an additive migration (no RunPython), the loud-vs-swallowed error split (50007 vs everything else) is precisely specified and fully covered by specs, and the DM copy's "your reaction counts automatically" claim is verified against import_member_guilds. Fat-model layering, typing, help_text, TextChoices-n/a, and BDD specs all conform. Two advisory notes inline — neither blocks.

PR #148, opened 2026-07-22
[fix] Member-only paths 404 instead of 500 on the public surface (v0.23.19)

HexagonStorms:
## The bug

`https://classes.pastlives.app/guilds/woodworking-guild/` returns **500**. It is a 500 *inside* a 404.

1. `/guilds/` is in `MEMBER_ONLY_PATH_PREFIXES`, so on a guest surface `SurfaceMiddleware._handle_public_surface` **deliberately** raises `Http404`. A 404 is the intended behaviour.
2. `SurfaceMiddleware` is listed **before** `AuthenticationMiddleware`, so when it raises, `request.user` was never set.
3. `templates/404.html` extends `hub/base.html`, so Django renders it with the request and runs **every context processor**. Three of them dereferenced `request.user` bare.

```
File "django/views/defaults.py", line 64, in page_not_found
    body = template.render(context, request)
File "django/template/context.py", line 263, in bind_template
    context = processor(self.request)
File "billing/context_processors.py", line 12, in tab_context
    if not request.user.is_authenticated:
AttributeError: 'WSGIRequest' object has no attribute 'user'
```

**Scope:** all 13 `MEMBER_ONLY_PATH_PREFIXES` on **both** guest hosts (`classes.pastlives.app`, `book.pastlives.space`), for **anonymous and signed-in** visitors alike — the middleware short-circuits before auth runs either way. Ordinary unknown paths 404 correctly, because that `Http404` comes from the URL resolver *after* auth middleware.

**Regression from** #119, which introduced `templates/404.html`. Before it, Django's built-in error page rendered without context processors.

## The fix

Guard `request.user` in the three offending processors, matching the precedent already set by `core.context_processors.persona`:

| Processor | State | Fallback |
|---|---|---|
| `billing.context_processors.tab_context` | the one that actually blew up | `{}` |
| `hub.context_processors.hub_sidebar` | latent (unreached — `tab_context` failed first) | empty-sidebar dict |
| `core.context_processors.notification_badge` | latent | `{"unread_notification_count": 0}` |

**Audit:** every processor registered in `TEMPLATES['OPTIONS']['context_processors']` was checked for attributes set by later middleware. The rest were already safe — `surface` and `persona` use `getattr`, the others touch only settings/DB/`request.path`, and Django's own `auth` (`hasattr(request, "user")`) and `messages` (`getattr(request, "_messages", [])`) guard themselves. No further offenders.

## Known follow-up (deliberately not done here)

`SurfaceMiddleware` is **not** reordered relative to `AuthenticationMiddleware`. That would address the deeper ordering issue, but the pre-auth placement is deliberate and documented in `core/middleware.py`. Hardening the error page is the correct fix for a production 500; reordering is a separate, larger change.

## Tests

`tests/core/public_surface_404_spec.py` — 29 cases across 7 member-only prefixes × both guest hosts × anonymous and signed-in, plus a members-host control.

The critical detail, and the reason nothing caught this: under `DEBUG=True` Django serves a **technical 404 that skips context processors entirely**, so the bug is invisible in dev. The spec forces `DEBUG=False` and asserts on the real `404.html` body, not just the status code.

Red → green, verified by stashing only the three source fixes:

```
# before the fix
28 failed, 1 passed, 29 warnings in 11.28s
E  AssertionError: classes.pastlives.app/guilds/ returned 500, expected 404

# after the fix
29 passed, 29 warnings in 12.86s
```

Full suite: **6278 passed, 1 failed**, coverage 98.27% (gate 98%). The single failure — `tests/core/integrations/discord_events_spec.py::it_maps_monthly_to_an_nth_weekday_rule` — is a pre-existing, date-dependent failure confirmed on clean `origin/main` and unrelated to this change. `ruff format` / `ruff check` clean; `mypy plfog/ core/ membership/ hub/` clean.

## Version

`VERSION` → `0.23.19`, with a member-facing CHANGELOG entry (this is a bug members lived with on production).

### DIFF
--- a/billing/context_processors.py
+++ b/billing/context_processors.py
@@ -8,8 +8,15 @@ from django.http import HttpRequest
 
 
 def tab_context(request: HttpRequest) -> dict[str, Any]:
-    """Add tab balance and status to template context for the balance pill."""
-    if not request.user.is_authenticated:
+    """Add tab balance and status to template context for the balance pill.
+
+    ``request.user`` is read defensively. ``SurfaceMiddleware`` short-circuits member-only
+    paths on the guest surfaces with an ``Http404`` *before* ``AuthenticationMiddleware``
+    has run, and the themed ``templates/404.html`` renders every context processor — so on
+    those responses the attribute does not exist yet. Same guard as ``core.persona``.
+    """
+    user = getattr(request, "user", None)
+    if user is None or not user.is_authenticated:
         return {}
 
     from core.models import SiteConfiguration
@@ -19,7 +26,7 @@ def tab_context(request: HttpRequest) -> dict[str, Any]:
 
     from membership.models import Member
 
-    member: Member | None = getattr(request.user, "member", None)
+    member: Member | None = getattr(user, "member", None)
     if member is None:
         return {}
 
--- a/core/context_processors.py
+++ b/core/context_processors.py
@@ -97,9 +97,15 @@ def google_analytics(request: HttpRequest) -> dict[str, str]:
 
 
 def notification_badge(request: HttpRequest) -> dict[str, int]:
-    """Unread notification count for the topbar bell. 0 for anonymous users."""
-    user = request.user
-    if not user.is_authenticated:
+    """Unread notification count for the topbar bell. 0 for anonymous users.
+
+    ``request.user`` is read defensively. ``SurfaceMiddleware`` short-circuits member-only
+    paths on the guest surfaces with an ``Http404`` *before* ``AuthenticationMiddleware``
+    has run, and the themed ``templates/404.html`` renders every context processor — so on
+    those responses the attribute does not exist yet. Same guard as ``persona`` below.
+    """
+    user = getattr(request, "user", None)
+    if user is None or not user.is_authenticated:
         return {"unread_notification_count": 0}
     from core.models import Notification
 
--- a/hub/context_processors.py
+++ b/hub/context_processors.py
@@ -16,13 +16,19 @@ def hub_sidebar(request: HttpRequest) -> dict[str, Any]:
     hub/base.html gets the sidebar data for free — without each view having
     to call a _get_hub_context helper. Returns empty values for anonymous
     requests so login/public pages don't hit the DB.
+
+    ``request.user`` is read defensively. ``SurfaceMiddleware`` short-circuits member-only
+    paths on the guest surfaces with an ``Http404`` *before* ``AuthenticationMiddleware``
+    has run, and the themed ``templates/404.html`` renders every context processor — so on
+    those responses the attribute does not exist yet. Same guard as ``core.persona``.
     """
-    if not getattr(request.user, "is_authenticated", False):
+    user = getattr(request, "user", None)
+    if not getattr(user, "is_authenticated", False):
         return {"guilds": Guild.objects.none(), "user_initials": "", "user_profile_photo_url": ""}
 
     initials = ""
     photo_url = ""
-    member: Member | None = getattr(request.user, "member", None)
+    member: Member | None = getattr(user, "member", None)
     if member is not None:
         initials = member.initials
         if member.profile_photo:
--- a/plfog/version.py
+++ b/plfog/version.py
@@ -2,9 +2,23 @@
 
 from __future__ import annotations
 
-VERSION = "0.23.18"
+VERSION = "0.23.19"
 
 CHANGELOG: list[dict[str, str | list[str]]] = [
+    {
+        "version": "0.23.19",
+        "date": "2026-07-22",
+        "title": "Guild and account links from the class site now show a proper 'page not found'",
+        "changes": [
+            (
+                "Opening a guild page, member directory, settings, or billing link while you were "
+                "on the class-booking site showed an error page instead of telling you the page "
+                "isn't available there. Those links now land on a friendly 'we couldn't find that "
+                "page' message with a way back to browsing classes — and they still work as normal "
+                "when you're signed in on the member hub."
+            ),
+        ],
+    },
     {
         "version": "0.23.18",
         "date": "2026-07-21",
--- a/tests/core/public_surface_404_spec.py
+++ b/tests/core/public_surface_404_spec.py
@@ -1,0 +1,72 @@
+"""Regression specs: member-only paths must 404 (never 500) on the public surface.
+
+``SurfaceMiddleware`` runs *before* ``AuthenticationMiddleware`` on purpose, so when it
+raises ``Http404`` for a member-only path on a guest host, ``request.user`` has never been
+set. The themed ``templates/404.html`` extends ``hub/base.html``, which runs every context
+processor — so any processor that dereferences ``request.user`` bare turned that intended
+404 into a 500.
+
+These specs must run with ``DEBUG=False``. Under ``DEBUG=True`` Django serves its technical
+404 page, which never touches the context processors — which is exactly why this bug shipped
+to production unnoticed.
+"""
+
+from __future__ import annotations
+
+import pytest
+from django.contrib.auth.models import User
+from django.test import Client
+
+# The member-only prefixes that are also plausible guest-facing links (crawlers, shared URLs).
+MEMBER_ONLY_PATHS = [
+    "/guilds/",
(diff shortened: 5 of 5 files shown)


PastLivesReviewBot (approved):

## What was reviewed

**Important:** I reviewed the **locally merged** state, not the published head.

`origin/fix/500-inside-404-public-surface` is at `3182f56`, which GitHub reports as `CONFLICTING` with `main`, and whose CI `test` job is red. The conflict has been resolved **locally and is not yet pushed**: merge commit `dc7601f` (merges `origin/main` `41cf97b` into the PR branch). Everything below assesses `dc7601f`. **This approval is contingent on that resolution being pushed unchanged.** Anyone reading the PR on GitHub today is not looking at the code I reviewed.

Relative to `main`, the merged tree touches exactly five files:

```
billing/context_processors.py         | 13 +++--
core/context_processors.py            | 12 ++++--
hub/context_processors.py             | 10 ++++-
plfog/version.py                      | 18 +++++++-
tests/core/public_surface_404_spec.py | 72 ++++++++++++++++++++++++++++
```

The merge changed **only** `plfog/version.py` relative to the PR head; no source or spec content moved during conflict resolution.

## Verification performed

Run in a clean worktree at `dc7601f`.

| Check | Result |
|---|---|
| Full suite | **6286 passed, 0 failed** (8m11s) |
| Coverage | **98.27%**, gate 98.0% reached |
| `ruff format --check .` | 969 files already formatted |
| `ruff check .` | All checks passed |
| Red to green | Reverted only the three processor fixes: **28 failed, 1 passed**. Restored: **29 passed**. Matches the PR description exactly. |

The pre-existing `discord_events_spec.py::it_maps_monthly_to_an_nth_weekday_rule` failure cited in the PR body is gone, because `main`'s #151 fix comes in with the merge. That is the whole reason the published head's CI is red and the merged state's is not.

`mypy` could not be run: the local venv has a django-stubs plugin incompatibility (`Error constructing plugin instance of NewSemanalDjangoPlugin`, mypy 2.2.0) that is unrelated to this branch. Type checking is unverified locally; CI's `lint` job passed on the published head.

## The fix itself

Correct, minimal, and well targeted. The three processors now read `request.user` through `getattr(request, "user", None)`, matching the guard `core.context_processors.persona` already used and the same reasoning already documented in `core/middleware.py` for the signage no-zones page. Full type annotations, docstrings that explain *why* the guard exists rather than restating the code, no behaviour change on any authenticated path.

I independently re-ran the audit claim in the PR body. It holds:

* `django.contrib.auth.context_processors.auth` guards with `hasattr(request, "user")`.
* `django.contrib.messages.context_processors.messages` guards with `getattr(request, "_messages", [])`.
* `surface` and `persona` already use `getattr`.
* `registration_mode`, `app_version`, `theme`, `feature_flags`, `google_analytics` touch only settings, the DB, or `request.path`.

No further offenders in `TEMPLATES['OPTIONS']['context_processors']`.

I also checked the adjacent blast radius: `templates/` contains a `404.html` and no `403.html`, `400.html`, or `500.html`, so every other error handler still uses Django's context-processor-free default. No sibling exposure.

Declining to reorder `SurfaceMiddleware` relative to `AuthenticationMiddleware` is the right call for a production hotfix, and saying so explicitly in the PR body is the right way to record it.

## Version conflict resolution: I agree with it

The resolution bumps `VERSION` to `0.23.20`, stamps this PR's entry there, and re-stamps the "DM nudge for unlinked guild-choosers" entry (which `main` carries at `0.23.19`) to `0.23.20` as well.

I checked this three ways and it holds up:

1. **Mechanically.** `.github/scripts/discord_release_notify.py::_release_entries()` selects *all* entries whose version equals `VERSION`, and `_bullets()` flattens across them. I imported the merged `plfog/version.py` and confirmed it resolves to exactly two entries at `0.23.20`. Nothing is orphaned by the bump itself.
2. **Against the documented rule.** CLAUDE.md explicitly allows it: "Multiple entries can share that VERSION (features released together) and all post." Multi-entry versions are already normal in this changelog (`0.22.9` carries nine, `0.21.19` five, `0.23.4` and `0.23.0` two each).
3. **Against precedent.** This is not a novel maneuver. `41cf97b` (#151), the immediately preceding commit on `main`, did the identical thing: bumped `VERSION` from `0.23.18` to `0.23.19` and re-stamped that same nudge entry forward. The resolution is consistent with what `main` did one commit ago.

The stated rationale is also factually correct. The `discord-notify` workflow last ran `2026-06-28` against `0.19.17`-era `main`, so the nudge entry has genuinely never been announced and would have been stranded at a superseded version number.

Two non-blocking observations on it:

* **Headline order.** `discord_release_notify.py` uses `entries[0]["title"]` for the embed headline, so the release will be announced as "New update: Guild and account links from the class site now show a proper 'page not found'". A bugfix will headline a release that also ships a feature. Putting the nudge entry first would read better to members. Purely a curation preference.
* **The larger backlog is untouched, and that is fine here.** Rescuing the nudge entry is correct, but it rescues one of nineteen `0.23.x` entries; `0.23.1` through `0.23.17` (Discord voting, `/members`, class sales, the calendar work) plus all of `0.20` through `0.22` remain unannounced and will stay that way, because the workflow posts only the current `VERSION`. That is a pre-existing condition of `main`, not something this PR created, and the right remedy is a deliberate combined release-notes pass like `6b094fb` did for `0.20` and `0.21`. Out of scope for a bugfix. Flagging it so it does not get quietly lost.

Separately: #151's title says `v0.23.25` but it actually set `VERSION = "0.23.19"`. A title typo on `main`, not this PR's problem, but worth knowing so `0.23.20` does not look like a regression.

## Findings (all non-blocking)

**1. The members-host control assertion is too weak to do its job.**

`tests/core/public_surface_404_spec.py`:

```python
response = client.get("/guilds/")
assert response.status_code != 404
```

That control exists to prove the fix did not over-apply and start 404ing member paths on the members host. But `!= 404` also passes on a 500, which is precisely the failure mode this whole spec file exists to catch. I confirmed the actual response is **200** for an anonymous client. `assert response.status_code == 200` would be a real control; the current form would sit green through a regression.

**2. The guilds surface had the same 500 and is fixed, but is not locked in by a test.**

The PR body scopes the bug to "both guest hosts" via `MEMBER_ONLY_PATH_PREFIXES`. `SurfaceMiddleware._handle_guilds_surface` also raises `Http404` before `AuthenticationMiddleware` runs, for any path outside `GUILDS_ALLOWED_VIEW_NAMES`. I verified empirically: with the three processor fixes reverted, `GET /billing/` on `guilds.pastlives.app` returns **500**; with them applied it returns **404**. So `guilds.pastlives.app` was a third affected production host, and the fix already covers it (the guard is in the processors, not the middleware). But `GUEST_HOSTS` in the spec lists only the two `PUBLIC_HOSTS`, so nothing prevents a future refactor from regressing the guilds surface with the suite still green. One case pointed at a `GUILDS_HOSTS` host would close that. (I also probed the signage surface; it was already returning 404 pre-fix, so no exposure there.)

**3. Non-deterministic test usernames, and a factory already exists.**

```python
user = User.objects.create_user(username=f"guest-404-{abs(hash(path + host))}", ...)
```

`hash()` on a `str` is salted by `PYTHONHASHSEED`, so these usernames differ run to run. Collision risk across fourteen parametrized cases is negligible, but a random-input `IntegrityError` is an unpleasant flake to chase. CLAUDE.md asks for factory-boy for all test data, and `tests/membership/factories.py::UserFactory` already exists with a `Sequence`; using it removes the nondeterminism and the standards deviation in one move. (`User.objects.create_user` does appear elsewhere in the suite, so this is a nit, not a violation of local practice.)

**4. The two `describe` blocks are byte-identical apart from `force_login`.**

Fourteen cases duplicated. A third `authenticated` parametrize axis would halve the file. Cosmetic.

## Verdict

**Approved.** The diagnosis is exact, the fix is the smallest correct one, the regression spec is genuinely red before and green after, and the `DEBUG=False` detail (which is why this shipped unnoticed) is both understood and pinned in the test. The four findings above are all polish; none of them justifies another round trip on a production 500.

Two conditions before merge:

1. **Push the local merge.** The published head is conflicting and red. Nothing I reviewed exists on GitHub yet.
2. **Re-run CI on the pushed merge** and confirm green there, since my run was local.

PR #149, opened 2026-07-22
[notifications] Transactional events now always email (v0.23.21)

HexagonStorms:
## The bug

Every event the registry seeds from `core/triggers.py` inherited **`EMAIL = OFF`**.
`core/events/registry.py::_channels_from_trigger` derives the email channel from
`trigger.force_email` / `trigger.email_default` — and not one of the ~24 legacy
`Trigger` rows ever set either flag. Later Phase-6 events were hand-classified
`_EMAIL_ON` / `_EMAIL_FORCED`; nobody went back for the legacy ones. That is why the
email-copy review gallery listed 14 events under **"No email is sent"**.

Genuinely transactional emails (registration confirmation, reminders, waitlist,
orientations, tab receipts, invites) were never affected — they pass `email_to=` to
`emit()`, which bypasses preferences entirely.

Five events fell through the crack and have only ever reached members as an in-app
bell row:

| Event | Resolver | Emitted at |
|---|---|---|
| `class_cancelled` | `ALL_ACTIVE_MEMBERS` | `ClassOffering.archive` |
| `refund_issued` | `REGISTRANT` | `Registration.save` |
| `tab_entry_added` | `TAB_MEMBER` | `Tab._dispatch_entry_notifications` |
| `tab_approaching_limit` | `TAB_MEMBER` | `Tab._dispatch_entry_notifications` |
| `lease_expiring` | `LEASE_TENANT` | `send_lease_expiry_reminders` |

## What changed

**1. `force_email=True` on all five triggers.** These are transactional, not opt-out-able;
`preferences.wants()` short-circuits on a `FORCED` spec before it ever reads a
`NotificationPreference` row. They stay visible on the settings matrix, rendered
locked-on (`settings_matrix` sources the registry, not `triggers.for_member`).

**2. The guest hole is closed.** `refund_issued` was emitted inside `if self.member is not
None`, and `class_cancelled` could only reach people the `ALL_ACTIVE_MEMBERS` resolver
knew about — so a **guest registrant with no account got nothing at all**: no bell, no
email. Both now pass `email_to=` with the address on the *registration*, following
`classes.emails.send_registration_confirmation`.

`ClassOffering.archive` also passes **`suppress_email=True` unconditionally**. `email_to`
alone implies it, but only when the list is non-empty — so archiving a class nobody
booked would have fanned a FORCED email out to every active member. The bell row still
broadcasts to all active members; the email stays with the people who actually booked.
This is also strictly better than the alternative: archiving is routine housekeeping and
mailing ~600 members about it would be spam.

**3. All five now render from the DB copy catalogue.** They dropped their explicit
`title`/`body` ("fixed message" mode, one flat string to every channel) and pass the
documented merge fields instead. `class_cancelled` already had good curated copy;
`refund_issued`, `tab_entry_added`, `tab_approaching_limit` and `lease_expiring` were
falling back to `_generic_copy` (the settings-page blurb, with no amount, class, space
or date in it) and gained curated copy that says the actual thing:

```
[tab_entry_added] SUBJECT: $18.00 added to your tab
Hi Robin Vale,

Bandsaw blade — $18.00 was added to your tab at Past Lives.
It will be included in your next monthly charge.

See everything on your tab: https://…/tab/

If this doesn't look right, reply to this email and we'll sort it out.
```

**4. The one emit call site with no `period=`.** `Invite.mark_accepted` had none, so every
acceptance in the app collapsed onto a single `EventDelivery` slot — meaning only the
**first** invite an admin ever sent notified them, and every later acceptance was
silently swallowed as a duplicate. Now `invite:<pk>:accepted`.

**5. The gallery updates itself.** The "No email is sent" table drops from **14 rows to 9**
and all five gain rendered cards, no context fixes needed. `class_cancelled` /
`refund_issued` carry an explicit EMAIL-audience override so the card documents who gets
the *mail* (registrants, guests included), not who gets the bell.

## Tests

`tests/core/events/transactional_email_defaults_spec.py` — 18 specs: the FORCED
declaration + the preference short-circuit for a member who explicitly opted out; a real
email in the outbox from each of the five real call sites; the **guest** case for
`class_cancelled` and `refund_issued`; that an uninvolved active member gets a bell and
no mail; that an unbooked archive sends nothing at all; and that a second invite's
acceptance still notifies.

```
tests/core/events/transactional_email_defaults_spec.py ................. [ 94%]
.                                                                        [100%]
======================== 18 passed, 3 warnings in 9.51s ========================
```

Full suite: **6265 passed, 3 failed** — all three pre-existing/environmental and
unrelated:
* `discord_events_spec::it_maps_monthly_to_an_nth_weekday_rule` — reproduced on a clean
  `origin/main` worktree (date-dependent: today is the 5th Wednesday).
* two `settings_spec` `EMAIL_BACKEND` tests — the local dev image bakes
  `EMAIL_BACKEND=smtp`; they pass in CI's clean env.

`ruff format` / `ruff check` clean; `mypy plfog/ core/ membership/ hub/` clean, and clean
on `classes/models.py` + `billing/models.py`.

## Post-deploy op

The DB is the source of truth for copy at send time. The four newly-curated entries only
reach production once **`seed_notification_templates`** re-runs (it refreshes any row not
flagged `is_overridden`, so admin-authored copy is safe). Without it these four keep
sending the old generic body — correct, just thin.

### DIFF
--- a/billing/models.py
+++ b/billing/models.py
@@ -437,7 +437,7 @@ class Tab(models.Model):
         current: Decimal,
         locked_self: Tab,
     ) -> None:
-        """Dispatch in-app notifications for a newly added tab entry.
+        """Dispatch the notifications for a newly added tab entry.
 
         Each event is keyed by the entry's pk in its dedupe ``period`` so every entry
         fires its own notification (the old ``dispatch`` had no dedupe) while a re-run
@@ -445,17 +445,23 @@ class Tab(models.Model):
         """
         from core.events.emit import emit
 
-        # Both events resolve the TAB_MEMBER (the tab's own member); their EMAIL channel
-        # defaults off, so this matches the old in-app-only dispatch. The SiteActivity row
-        # for the entry is logged by ``add_entry`` (the event's registry ``activity_kind``
-        # is None to avoid duplicating it).
+        # Both events resolve the TAB_MEMBER (the tab's own member) and are transactional
+        # (money the member owes), so their EMAIL channel is FORCED — the bell row and the
+        # email both go out, rendered from the DB-editable copy catalogue. The SiteActivity
+        # row for the entry is logged by ``add_entry`` (the event's registry
+        # ``activity_kind`` is None to avoid duplicating it).
+        tab_url = self.absolute_tab_url
         if added_by is not None and not is_self_service and self.member.user is not None:
             emit(
                 "tab_entry_added",
                 actor=added_by,
-                context={"member": self.member},
-                title="Tab entry added",
-                body=f"{description} — ${amount}",
+                context={
+                    "member": self.member,
+                    "member_name": self.member.display_name,
+                    "description": description,
+                    "amount": f"${amount}",
+                    "tab_url": tab_url,
+                },
                 url="/tab/",
                 period=f"entry:{entry.pk}:added",
             )
@@ -463,12 +469,21 @@ class Tab(models.Model):
         if self.member.user is not None and limit and (current + amount) / limit >= Decimal("0.80"):
             emit(
                 "tab_approaching_limit",
-                context={"member": self.member},
-                title="Tab approaching its limit",
-                body=f"Your tab balance is ${current + amount} of ${limit}.",
+                context={
+                    "member": self.member,
+                    "member_name": self.member.display_name,
+                    "balance": f"${current + amount}",
+                    "limit": f"${limit}",
+                    "tab_url": tab_url,
+                },
                 url="/tab/",
                 period=f"entry:{entry.pk}:limit",
             )
+
+    @property
+    def absolute_tab_url(self) -> str:
+        """The member-site URL of the tab page, absolute so it works from an email."""
+        return f"{settings.MEMBER_BASE_URL.rstrip('/')}/tab/"
 
     def lock(self, reason: str) -> None:
         """Lock the tab, preventing new entries."""
--- a/classes/models.py
+++ b/classes/models.py
@@ -724,15 +724,55 @@ class ClassOffering(HeroCropMixin, models.Model):
 
         activity.log(CmsActivity.Kind.CLASS_ARCHIVED, class_offering=self)
         # In-app broadcast to all active members (the ``class_cancelled`` event resolves
-        # ALL_ACTIVE_MEMBERS; its EMAIL channel defaults off, so this matches the old
-        # in-app-only dispatch — email only to a member who has opted in).
+        # ALL_ACTIVE_MEMBERS), but the EMAIL goes ONLY to the people who actually booked
+        # the class — including guests with no account, who are unreachable by the
+        # resolver. ``suppress_email=True`` is unconditional (not merely implied by a
+        # non-empty ``email_to``) so archiving a class nobody booked can never turn into a
+        # site-wide email blast to every active member.
         emit(
             "class_cancelled",
             target=self,
-            title="A class was cancelled",
-            body=self.title,
+            context={
+                "member_name": "there",
+                "class_title": self.title,
+                "class_starts_at": self.cancellation_date_label,
+            },
             url="/classes/",
+            email_to=self.registrant_notice_emails,
+            suppress_email=True,
             period=f"offering:{self.pk}:archived",
+        )
+
+    @property
+    def cancellation_date_label(self) -> str:
+        """A human date for the cancellation copy ("Saturday, July 12").
+
+        Falls back to a neutral phrase for a flexibly-scheduled class with no session
+        rows, so the copy never renders a bare placeholder.
+        """
+        starts_at = self.earliest_session_at
+        if starts_at is None:
+            return "its scheduled date"
+        return date_format(localtime(starts_at), "l, F j")
+
+    @property
+    def registrant_notice_emails(self) -> list[str]:
+        """Every registrant address that should hear about a change to this class.
+
+        Drawn from ``Registration.email`` rather than the linked member, so a **guest**
+        registrant (no account, ``member`` is ``NULL``) is reached too. Rows that have
+        already left the class (cancelled / refunded) are excluded.
+        """
+        return list(
+            self.registrations.filter(
+                status__in=[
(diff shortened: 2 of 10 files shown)


PastLivesReviewBot (approved):

Reviewed at head `95a02ea` (post-merge of `origin/main`), in a clean worktree.

## Merge resolution — verified sound

`git diff` of the PR against `origin/main`, excluding `plfog/version.py`, is **byte-identical** to the pre-merge diff at `786d626`. No source content was lost, altered, or silently reverted by the merge. Same 10 files, same 669/46 line counts.

`plfog/version.py` resolved exactly as intended: main's file taken verbatim (its shipped 0.23.20 entries from #147/#148 untouched, not relitigated), with this branch's single entry spliced on top and renumbered `0.23.20` to `0.23.21` because `0.23.20` was claimed by #148 on main. `VERSION = "0.23.21"` and exactly one CHANGELOG entry is stamped at it, so `announce_release` resolves and the Discord workflow posts one clean feature entry. This follows the CLAUDE.md changelog rule.

CI blocker cleared: `discord_events_spec::it_maps_monthly_to_an_nth_weekday_rule` now **passes** (main's #151 pinned the specs to fixed dates).

## Verification run

| Check | Result |
|---|---|
| `ruff format --check .` | 970 files already formatted |
| `ruff check .` | All checks passed |
| `mypy plfog/ core/ membership/ hub/ classes/models.py billing/models.py` | Success, no issues in 160 source files |
| Full suite (`tests/ core/ classes/ billing/` — every `*_spec.py` in the repo) | **6304 passed, 0 failed** |

The 3 failures cited in the PR description are gone: the Discord one is fixed by main, and the two `settings_spec` `EMAIL_BACKEND` ones were local-env artifacts (clean in a fresh worktree).

## Standards review

- **Fat models, skinny views** — all five changes live in model methods, model properties, and a management command. No view touched.
- **Type annotations** — the two new properties (`ClassOffering.cancellation_date_label -> str`, `ClassOffering.registrant_notice_emails -> list[str]`) and `Tab.absolute_tab_url -> str` are fully annotated. `@property` is the right call for all three (cheap derived data, no side effects).
- **N+1** — `registrant_notice_emails` uses a single filtered `values_list(..., flat=True)`; no per-row member access, which is also what makes the guest case work.
- **Tests** — `tests/core/events/transactional_email_defaults_spec.py`, 18 specs, correct BDD shape. Only `describe_*` nesting (no `context_*` blocks, which this repo silently skips). No `@pytest.mark.skip`, no `# pragma: no cover`, no `# pragma: no mutate`. Coverage spans the FORCED declaration, the preference short-circuit for an opted-out member, a real outbox assertion from each of the five call sites, both guest paths, the negative case (uninvolved member gets a bell and no mail), the unbooked-archive silence, and second-invite acceptance.
- **No migrations** in this PR, so nothing to check on reverse functions.
- **Security** — no raw SQL, no unescaped interpolation. The new HTML copy is trusted literal markup; merge values still route through `rendering.render_html`, which escapes them before the shell's `|safe`.

## Things I specifically chased down

**1. The `suppress_email=True` on `ClassOffering.archive` is load-bearing, and correct.** Confirmed against `emit()`: `suppress_user_email = bool(explicit_emails) or suppress_email`. With an empty `email_to` list, `bool([])` is `False` — so without the unconditional `suppress_email`, archiving a class nobody booked would have FORCED an email to every active member the `ALL_ACTIVE_MEMBERS` resolver returns. The comment in the code explains exactly this. Good instinct, and the spec `it_sends_no_email_at_all` locks it in.

**2. No other blast radius.** Grepped every production emit call site for the five newly-FORCED events: exactly one apiece (`ClassOffering.archive`, `Registration.save`, `Tab._dispatch_entry_notifications` twice, `send_lease_expiry_reminders`). Nothing else can reach the new FORCED email channel from an unreviewed path.

**3. Blank-address safety.** `refund_issued` passes `email_to=self.email` as a bare string, so an empty `Registration.email` would produce `[""]` — non-empty, which suppresses the member email. `_explicit_email_fan_out` guards with `if not address ... continue`, so nothing is sent to `""` and no exception is raised. Safe, and not a regression (email defaulted OFF before this PR anyway).

**4. The "post-deploy op" is already automatic.** The PR description flags `seed_notification_templates` as a manual step, but it already runs unattended on every deploy: `render.yaml` line 9 and the `Dockerfile` CMD both invoke `python manage.py seed_notification_templates --quiet` ahead of gunicorn. The command refreshes any row not flagged `is_overridden`, so the four newly-curated entries land on production on the next deploy with no operator action, and admin-authored copy is preserved. Worth correcting in the description so nobody goes hunting for a step that is not needed.

**5. Copy placeholders line up with the emitted context** for all five events — `class_cancelled` (`member_name` / `class_title` / `class_starts_at`), `refund_issued`, `tab_entry_added`, `tab_approaching_limit`, `lease_expiring`. No context key can render as a visible missing-field marker. `cancellation_date_label` falling back to `"its scheduled date"` for a flexibly-scheduled class with no sessions is the right guard.

**6. `force_email` on the settings page.** Confirmed the PR description's claim: `settings_matrix.build_matrix` sources the registry, not `triggers.for_member`, and renders a `FORCED` cell locked-on. The five new ones stay visible to members rather than vanishing.

## Non-blocking notes

- **PR title still reads `(v0.23.20)`** but the file now ships `0.23.21`. Cosmetic, but the title is the human-readable release marker — worth retitling to `(v0.23.21)` before merge.
- **Stale docstring, pre-existing.** The `core/triggers.py` module docstring says forced triggers "never show a toggle", which was true of the old `for_member()` path but is not how `settings_matrix` renders them today (visible, locked-on). This PR adds five more triggers that hit the discrepancy without introducing it. A one-line docstring fix, whenever that module is next touched.
- `_EMAIL_AUDIENCE_OVERRIDES.get(key, default)` in the gallery registry uses `.get` with a fallback rather than the house `dict[key]` rule — correct here, since absence is the normal case and the default is the meaningful behavior, not a silent papering-over. Noted only so it does not read as an oversight.

Approved.

PR #150, opened 2026-07-22
[classes] Own our class images; make legacy sync non-destructive (v0.23.23)

HexagonStorms:
Makes the class catalog own its images outright, and makes the legacy CMS sync safe to keep running while we transition off Drupal.

Follows the production run of `download_legacy_images`, which pulled 612 images into R2 and cleared every `legacy_image_url`.

## Our images win

No change needed. Every render point already branches on `offering.image` first and falls back to the legacy proxy second. Verified all four sites (catalog cards, class detail hero, related-class cards, printable flyer). Nothing else renders a class image: no OG tag, no email, no calendar, no signage slide.

## Future syncs cannot clobber it

`_upsert_offering` never wrote `image`, and already guarded `legacy_image_url` behind a check for a local image. That guard read as incidental rather than as a rule, and a row could keep a stale legacy URL alongside a local image, one template edit away from proxying Drupal again.

Now an offering that owns an image has its `legacy_image_url` actively cleared, so it holds no handle on the legacy server at all. A genuinely new class still picks up its feed URL on first import. Migrated rows are invisible to `download_legacy_images` (it filters on `legacy_image_url__gt="", image=""`), so re-running does no duplicate work.

## Archive guard

`ClassOfferingQuerySet.archive_missing_from_legacy_feed(seen_ids)` now refuses to run when `seen_ids` is empty, or when the sweep would archive more than 50% of live legacy offerings.

Measured against live (non-archived) legacy rows, not all rows carrying a `legacy_cms_id`. Measuring against all of them would have wedged the guard permanently on once archived rows exceeded half the catalog.

Failure is loud: `logger.error` with a banner, counts, the ceiling, and the feed URL, stating that an empty payload is the likely cause rather than the classes being gone.

## Storage: 1.53 GB to roughly 66 MB

Two compounding problems, both fixed.

Normalization never fired. `offering.image.save(...)` commits the file and sets `_committed = True`, so `normalize_field_if_uploaded` returned early and `ClassOffering.save()`'s normalize call never ran. Raw originals were stored verbatim, including a 16.2 MB hero.

`file_overwrite: False` turned 149 distinct pictures into 612 distinct objects, since Django suffixes every name collision.

New `core.images.store_content_addressed` keys on the SHA-256 of the normalized bytes. Chosen over keying on the pre-migration URL snapshot for three reasons: the snapshot is a local file production does not have; dedupe has to happen on normalized bytes anyway, since normalization is what changes them; and content hashing also catches the same picture arriving under two different URLs.

New `optimize_class_images` command (idempotent, interruptible, `--dry-run`) repoints existing rows and deletes the objects it moves off. Ratio measured empirically by downloading 30 of the 149 distinct images and running the real pipeline: 56.2 MB to 9.3 MB, a factor of 0.165. Applied to the deduplicated corpus that is roughly 66 MB across 149 objects.

`download_legacy_images` is fixed so it cannot recreate the problem: it normalizes before storing, stores content-addressed, narrows `except Exception` to `(OSError, ValueError)`, raises `CommandError` once failures reach 50% (previously it exited 0 even if all 612 failed), and percent-decodes filenames so `Closeup%20Hands%202_11.JPG` reads back correctly.

## Unrelated bug found and fixed

`delete_orphan_on_replace` in `core/files.py` deleted the previous storage object unconditionally. Harmless while every row had its own object, but once images are shared, replacing one offering's photo would have deleted that file out from under every sibling still referencing it. Deletion now goes through `delete_if_unreferenced`. This affects all eleven call sites, not just classes.

## Legacy image proxy: kept, guarded

Kept deliberately. A genuinely new class from the feed has a legacy URL and no local image until the next migration run, which is the behavior this PR preserves. Deleting the view would trade a slow image for a missing one.

The deadlock risk is handled at its root: `_is_own_host()` refuses any hostname in `ALLOWED_HOSTS` or matching `request.get_host()`. When `classes.pastlives.space` is repointed at the app and added to `DJANGO_ALLOWED_HOSTS`, the guard arms itself with no code change, returns 404, and logs an error pointing at `download_legacy_images`. Without it, the proxy would fetch the app's own host, 24 requests per catalog page into 2 gunicorn workers.

Once the feed is switched off for good, the view, its URL, and the three template branches can come out in one commit.

## Tests

31 new specs, exercising real 4000x3000 JPEGs rather than stubs: re-sync does not overwrite a migrated image or re-set the URL; the guard trips on an empty feed; optimize is idempotent; normalization actually shrinks a large image.

Full suite under CI conditions: 6292 passed, coverage 98.26% against a 98% gate. New modules at 100%, `import_service.py` at 97% with both gaps pre-existing. `ruff format --check` and `ruff check` clean; `mypy` clean on the in-scope packages and back to baseline on `classes/`. `makemigrations --check` reports no changes; this is behavior-only.

The one deselected failure, `discord_events_spec::it_maps_monthly_to_an_nth_weekday_rule`, fails identically on pristine `origin/main` and is handled on `fix/monthly-nth-weekday-rrule`.

## Prod ops after merge

1. Run `optimize_class_images` once as a Render one-off job. Production landed 612 unoptimized objects because the migration ran on old code. Use `--dry-run` first. Safe to interrupt and re-run.
2. No migration, so no post-deploy `migrate` step.
3. Nothing to configure for the proxy guard; it arms itself at DNS cutover.
4. Watch the first nightly `sync_all_sources` for `ARCHIVE GUARD TRIPPED`. If it appears, the Drupal feed returned an empty or truncated payload and the guard has just saved the catalog.

### DIFF
--- a/classes/import_service.py
+++ b/classes/import_service.py
@@ -149,6 +149,8 @@ def _upsert_offering(item: dict[str, Any]) -> str | None:
         "status": status,
         "scheduling_type": scheduling_type,
     }
+    # ``image`` appears in neither dict and is never assigned below: once a picture lives
+    # in our own storage the feed must not be able to touch it.
     offering, created = ClassOffering.objects.update_or_create(
         legacy_cms_id=node_id,
         defaults=shared_defaults,
@@ -161,7 +163,16 @@ def _upsert_offering(item: dict[str, Any]) -> str | None:
             slug = f"{slug}-legacy"
         offering.slug = slug
 
-    if image_url and not offering.image:
+    # Image ownership. Once an offering has a local image — migrated off the legacy CMS by
+    # ``download_legacy_images`` or uploaded by staff — the Drupal URL is never written
+    # back, and any URL left over from an earlier sync is cleared. Otherwise the next run
+    # would silently re-point the site at the server we are decommissioning for a picture
+    # we already host, and the request-time proxy would start fetching it all over again.
+    # A genuinely new class with no local image still picks up its feed image URL on first
+    # import, ready for the next ``download_legacy_images`` run.
+    if offering.image:
+        offering.legacy_image_url = ""
+    elif image_url:
         offering.legacy_image_url = image_url
 
     if not offering.instructor_id:
@@ -204,10 +215,9 @@ def sync_legacy_cms() -> int:
 
         next_url = ((data.get("links") or {}).get("next") or {}).get("href")
 
-    # Archive offerings no longer present in the API
-    ClassOffering.objects.filter(legacy_cms_id__gt="").exclude(legacy_cms_id__in=seen_ids).update(
-        status=ClassOffering.Status.ARCHIVED
-    )
+    # Archive offerings no longer present in the API — guarded against a feed that
+    # succeeds but returns nothing, which would otherwise archive the whole catalog.
+    ClassOffering.objects.archive_missing_from_legacy_feed(seen_ids)
 
     # Sanitize: collapse the same class posted on many dates into one catalog group.
     from classes.grouping import regroup_offerings
--- a/classes/management/commands/download_legacy_images.py
+++ b/classes/management/commands/download_legacy_images.py
@@ -1,21 +1,54 @@
-"""Download hero images from legacy CMS into Django media storage."""
+"""Download hero images from the legacy CMS into Django media storage.
+
+Migration step off classes.pastlives.space: every offering that still points at the
+Drupal server gets its picture copied into our own storage, normalized to the hero
+ceiling, and de-duplicated so a picture shared by many offerings is stored once.
+``legacy_image_url`` is cleared as each row lands, which is what stops the site
+proxying the old server at request time.
+
+Idempotent: only rows that still have a legacy URL and no local image are touched,
+so a re-run after a partial failure picks up exactly what is left.
+"""
 
 from __future__ import annotations
 
+import urllib.error
 import urllib.request
 from pathlib import Path
+from urllib.parse import unquote, urlsplit
 
+from django.conf import settings
 from django.core.files.base import ContentFile
-from django.core.management.base import BaseCommand
+from django.core.management.base import BaseCommand, CommandError
+from PIL import UnidentifiedImageError
 
-from classes.models import ClassOffering
+from classes.models import CLASS_IMAGE_PREFIX, ClassOffering
+from core.images import normalize_image, store_content_addressed
+
+
+LEGACY_IMAGE_PREFIX = "https://classes.pastlives.space/"
+
+# Fraction of attempted downloads that may fail before the command reports failure via a
+# non-zero exit. A handful of dead images is routine; half the catalog failing means the
+# host, the network or the configuration is wrong and a human has to look.
+FAILURE_ABORT_FRACTION = 0.5
+
+
+def legacy_image_filename(url: str) -> str:
+    """Return the human-readable basename encoded in a legacy image URL.
+
+    Drops the query string and percent-decodes the path, so
+    ``.../Closeup%20Hands%202_11.JPG?itok=x`` reads back as
+    ``Closeup Hands 2_11.JPG`` rather than ``Closeup20Hands202_11.JPG``.
+    """
+    return Path(unquote(urlsplit(url).path)).name
 
 
 class Command(BaseCommand):
     help = "Download hero images from legacy_image_url into plfog media storage."
 
     def handle(self, *args, **options) -> None:
-        qs = ClassOffering.objects.filter(legacy_image_url__gt="", image="")
+        qs = ClassOffering.objects.filter(legacy_image_url__gt="", image="").order_by("pk")
         total = qs.count()
         if not total:
             self.stdout.write("No offerings with pending legacy images.")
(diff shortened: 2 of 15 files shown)


PastLivesReviewBot (approved):

Reviewed at head `a5ab5d7` (after the `origin/main` merge), in a clean worktree.

## Merge resolution verified

- The merge commit adds **exactly** main's `670a439..cc17de8` contribution and nothing else — diffed both sides excluding `version.py`, byte-identical.
- This branch's own source changes are **byte-identical** pre- and post-merge (`670a439..aa3d2ba` vs `cc17de8..a5ab5d7`, excluding `version.py`). No content lost.
- `version.py` resolution is sound: main's changelog history is intact and unmodified, this branch's `0.23.23` entry sits on top, and `VERSION == "0.23.23"` matches that top entry, so `announce_release` resolves.
- `discord_events_spec::it_maps_monthly_to_an_nth_weekday_rule` now **passes** (fixed by #151 arriving via main).

## Verification run

- Full suite: **6329 passed**, 0 failed, coverage **98.28%** against the 98% gate. The 30 deselected are the default `-m 'not e2e'` Playwright marker, not skips from this PR.
- `ruff check` clean, `ruff format --check` clean (972 files), `mypy` clean on `core/`, `classes/management/`, `classes/import_service.py`.
- `makemigrations --check` reports no changes, matching the behavior-only claim.
- New modules at 100%: `download_legacy_images.py`, `optimize_class_images.py`, `views_legacy_image.py`, `core/files.py`.
- No `pragma: no cover`, `pragma: no mutate`, `@pytest.mark.skip`, or `context_*` blocks introduced.

## Non-destructiveness verified adversarially

The central claim was checked against the code and then re-checked with four independent throwaway specs written for this review (all passing, since discarded):

- Three offerings sharing one storage object plus one solo offering, run through `optimize_class_images`: every row still points at an object that exists, and the three sharers converge on one key. No row is left pointing at a deleted object.
- A second `optimize_class_images` run is a genuine no-op; the key is stable and the object survives.
- One offering replacing its image while a sibling still references the same key does **not** delete the shared object — this is the bug `delete_if_unreferenced` was added to prevent, and the fix holds.
- Repeated syncs over a migrated offering preserve both `image` and the cleared `legacy_image_url`.

Reasoning checks that also hold up:

- `_upsert_offering` writes `image` in neither the defaults dict nor the body, so the feed genuinely cannot touch a local image.
- Clearing `legacy_image_url` when a local image exists is self-healing rather than lossy: if staff later delete the local image, the next sync's `elif image_url` branch restores the URL from the feed.
- `sync_legacy_cms` reaches the archive sweep only after every page fetched successfully — `_fetch_json` raises on HTTP error, so a mid-pagination failure aborts before any archiving.
- The archive guard measures against live (non-archived) legacy rows, which is the right denominator; the spec at `it_ignores_offerings_already_archived_when_measuring_the_blast_radius` pins exactly the wedged-forever failure mode the alternative would have caused.
- `download_legacy_images` sets `offering.image` to a string key, so `_committed` is True and `normalize_field_if_uploaded` correctly no-ops — the bytes were already normalized before storage. `update_fields=["image", "legacy_image_url"]` is safe on that path because the old image is empty, so the hero-crop reset never fires.
- `optimize_class_images` deliberately bypasses `save()` via `.update()` to avoid discarding a staff-set hero crop when a row re-points at the same picture. Correct, and the comment explains why.
- SSRF posture is unchanged or better: prefix allowlist retained, redirects still refused, plus the new self-host guard.

## Non-blocking notes

None of these block the merge; recording them for when image sharing spreads beyond `ClassOffering`.

1. **`_is_own_host` misses wildcard `ALLOWED_HOSTS` entries.** `h.strip().lstrip("*.")` turns `*.pastlives.space` into `pastlives.space`, so a wildcard entry would not arm the guard for `classes.pastlives.space`. The `request.get_host()` half does not cover this either, since the current request's host is typically the members or catalog domain, not the image URL's host. This repo's convention is explicit hostnames (see the `GUILDS_HOSTS` go-live comment in `settings.py`), so it is unreachable as operated. If a wildcard is ever added, the guard would silently stop protecting. A suffix match for `*.`-prefixed entries would close it.

2. **`delete_if_unreferenced` uses `model._default_manager`.** For a soft-deleted model this excludes deleted rows from the reference check — `Guild` defines `objects = GuildManager()` first, so a soft-deleted guild sharing a banner would not count as a reference. Strictly better than the previous unconditional delete (no regression), and guild banners are never content-addressed today, so unreachable. Worth `_base_manager` if sharing ever reaches a soft-deleted model.

3. **Reference check is single-model, but the prefix is shared.** `ClassOffering.image` and `ClassImage.image` both use `upload_to="classes/images/"`, while `delete_if_unreferenced` only ever queries one model. Unreachable today, since only `ClassOffering` receives content-addressed keys and `file_overwrite=False` prevents natural key collisions. If `ClassImage` ever adopts content addressing under the same prefix, the check would need to span both models.

4. **Minor PR-description nit.** The description calls both `import_service.py` coverage gaps pre-existing; the `175->178` partial (an offering with neither a local image nor a feed image URL) is a new branch shape from the rewritten conditional. Immaterial — the gate passes at 98.28%.

5. **Guard behavior worth knowing at the console.** With a single live legacy offering, `stale_count(1) > total(1) * 0.5` always trips, so a genuinely-removed lone class will never auto-archive. That is the correct fail-safe direction and matches the docstring; just expect a manual archive in that tail case rather than treating the log line as a fault.

PR #151, opened 2026-07-22
[tests] Pin the Discord monthly-recurrence specs to fixed dates — unblocks CI on #148/#149 (v0.23.25)

HexagonStorms:
## Unblocks CI on PR #148 and PR #149

`origin/main` is red today, and therefore so is every open PR branched from it — including **#148** and **#149**. This PR fixes that failure. It is pre-existing on clean `main` and unrelated to either of those PRs.

```
FAILED tests/core/integrations/discord_events_spec.py::describe__recurrence_rule_for::it_maps_monthly_to_an_nth_weekday_rule
AssertionError: assert [{'n': 5, 'day': 2}] == [{'n': -1, 'day': 2}]
```

## Root cause

`tests/core/integrations/discord_events_spec.py:276` built its event with the bare `CommunityEventFactory`, whose default anchor is `timezone.now() + timedelta(days=7)` (`tests/membership/factories.py:244`). It then asserted:

```python
assert rule["by_n_weekday"] == [{"n": event._occurrence_ordinal(), "day": weekday}]
```

`CommunityEvent._occurrence_ordinal()` (`membership/models.py:3766`) returns `1–4`, or **`-1`** for a 5th weekday — the model's internal "last" sentinel. So on any day where `now + 7d` lands on the 5th occurrence of its weekday, the expectation became `-1` while the implementation correctly produced `5`, and the test failed. Today (2026-07-22 → Wed 2026-07-29, the 5th Wednesday) is one such day; 2026-12-24 is another. Both reproduce under a frozen clock.

## The implementation is right — this is a test defect

Discord's `recurrence_rule.by_n_weekday` is a **structured JSON object, not an RFC 5545 `BYDAY` string**, so RFC 5545's `-1WE` ("last Wednesday") form is not available here. Both the official docs and the community reference document the field identically:

> `n` | integer | *The week to reoccur on. 1 - 5*

Negative values are not permitted, and the array is capped at a single entry, monthly-frequency only. `_occurrence_n()` (`core/integrations/discord_events.py:198`) re-encoding the model's `-1` as `5` is therefore the **only** value the API accepts; passing `-1` straight through would be a hard 400. A sibling spec, `it_maps_a_last_weekday_ordinal_to_discord_5`, already pinned that mapping with a fixed date and has been passing all along — the failing assertion was simply comparing against the raw model ordinal instead of the Discord-encoded one.

Sources: [docs.discord.com — Guild Scheduled Event](https://docs.discord.com/developers/resources/guild-scheduled-event) · [Discord Userdoccers](https://docs.discord.food/resources/guild-scheduled-event)

**No production data is affected.** A read-only query against the prod database found **0** `CommunityEvent` rows with `recurrence = 'monthly'` (the only recurring event on prod is a single `weekly` one, already `synced`). Nothing to correct.

## What changed

Test-only. `_recurrence_rule_for`, `_occurrence_n` and `_occurrence_ordinal` are untouched.

- New `_anchored()` helper pins every event in the block to an explicit **local calendar date** — never derived from `timezone.now()`.
- `it_maps_monthly_to_an_nth_weekday_rule` is now parametrised over the 1st/2nd/3rd/4th/5th Wednesdays of July 2026 plus a Monday (covering the `0=Monday` weekday encoding), asserting the full rule dict.
- Both directions of "last" are now covered: Wed 2026-07-29 is the 5th *and* last Wednesday (`-1 → 5`), while Sat 2026-02-28 is the 4th *and* last Saturday (stays `4`).
- New range invariant walks all 31 anchors of a 31-day month and asserts `n` never leaves Discord's `1–5`.
- The weekly and unmappable-cadence specs are pinned to fixed anchors too.

## Proof of date-independence

The `_recurrence_rule_for` block was run under a frozen clock at eight "todays" spread across the year — including both dates that previously reproduced the failure. **16 passed** at every one:

```
today = 2026-01-01 → 16 passed      today = 2026-07-22 → 16 passed   (was RED)
today = 2026-07-01 → 16 passed      today = 2026-07-29 → 16 passed
today = 2026-07-08 → 16 passed      today = 2026-02-28 → 16 passed
today = 2026-07-15 → 16 passed      today = 2026-12-24 → 16 passed   (was RED)
```

The harness is not vacuous — restoring the pre-fix spec and running it under the same clocks reproduces the exact CI diff:

```
=== PRE-FIX spec, today = 2026-07-22 12:00:00 ===
E   At index 0 diff: {'n': 5, 'day': 2} != {'n': -1, 'day': 2}
FAILED ...::it_maps_monthly_to_an_nth_weekday_rule
=== PRE-FIX spec, today = 2026-12-24 12:00:00 ===
E   At index 0 diff: {'n': 5, 'day': 3} != {'n': -1, 'day': 3}
FAILED ...::it_maps_monthly_to_an_nth_weekday_rule
```

## Verification

- Full suite (SQLite, mirroring CI): **6257 passed**, coverage **98.26%** (gate 98%).
- `mypy plfog/ core/ membership/ hub/`: Success, 158 files.
- `ruff format --check .` + `ruff check .`: clean.

## Version & changelog

`VERSION` → **0.23.19** as of the branch tip. This PR was authored at **0.23.25**; a follow-up commit on this branch (`2957488`, *"Renumber to 0.23.19 so this lands first in the merge queue"*) renumbered it. Worth a second look before merge — **0.23.19 was reserved for PR #148**, so as it stands the two collide. Either number works for this PR; it just needs to be one nobody else is holding.

**No new CHANGELOG entry** — the shipped behaviour is unchanged, so there is nothing for members to read. Per the curation rules the existing top entry is re-stamped to the new `VERSION` instead, which also satisfies the `announce_release` invariant that `tests/core/events/new_events_spec.py::describe_release_published` enforces (it hard-fails a `VERSION` with no matching entry).

## Follow-up worth knowing (not a bug, no action here)

Discord's `n=5` means literally *the fifth week*, whereas FOG's own expansion treats a 5th-weekday anchor as **"last"** (`membership/models.py:3803`). A monthly series anchored on a 5th weekday will therefore appear on FOG's calendar every month but only in Discord's five-weekday months. Discord's schema offers no other encoding, so this is a documented ceiling of the integration rather than something fixable — noting it so it is not rediscovered as a bug later.

### DIFF
--- a/plfog/version.py
+++ b/plfog/version.py
@@ -2,12 +2,12 @@
 
 from __future__ import annotations
 
-VERSION = "0.23.18"
+VERSION = "0.23.25"
 
 CHANGELOG: list[dict[str, str | list[str]]] = [
     {
-        "version": "0.23.18",
-        "date": "2026-07-21",
+        "version": "0.23.25",
+        "date": "2026-07-22",
         "title": "Picked a guild on Discord but not linked yet? The bot now gives you a heads-up",
         "changes": [
             (
--- a/plfog/version.py
+++ b/plfog/version.py
@@ -2,11 +2,11 @@
 
 from __future__ import annotations
 
-VERSION = "0.23.25"
+VERSION = "0.23.19"
 
 CHANGELOG: list[dict[str, str | list[str]]] = [
     {
-        "version": "0.23.25",
+        "version": "0.23.19",
         "date": "2026-07-22",
         "title": "Picked a guild on Discord but not linked yet? The bot now gives you a heads-up",
         "changes": [
--- a/tests/core/integrations/discord_events_spec.py
+++ b/tests/core/integrations/discord_events_spec.py
@@ -8,6 +8,7 @@ exercised through ``respx``.
 from __future__ import annotations
 
 from datetime import datetime, timedelta
+from typing import Any
 from unittest.mock import MagicMock, patch
 
 import httpx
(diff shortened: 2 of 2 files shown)


PastLivesReviewBot (approved):

Test-only change plus a version bump. Verified independently rather than taking the description's word for it.

## Verification performed

**Root cause reproduced.** Restored the pre-fix spec and ran it today (2026-07-22 PDT):

```
E   AssertionError: assert [{'n': 5, 'day': 2}] == [{'n': -1, 'day': 2}]
FAILED ...::it_maps_monthly_to_an_nth_weekday_rule
```

Exactly the CI diff quoted in the description. The diagnosis holds: `_occurrence_ordinal()` returns `-1` for a 5th weekday, `_occurrence_n()` correctly re-encodes that to Discord's `5`, and the old assertion compared against the raw model ordinal instead of the encoded one. The implementation is right; the test was wrong.

**The fix is date-independent.** Ran the `_recurrence_rule_for` block under a monkeypatched `timezone.now()` at ten "todays", including two chosen beyond the list in the description (2026-04-30, 2027-03-31). **16 passed at every one**, matching the expected count for the rewritten block:

```
2026-01-01  16 passed      2026-07-29  16 passed
2026-07-01  16 passed      2026-02-28  16 passed
2026-07-08  16 passed      2026-12-24  16 passed   (RED pre-fix)
2026-07-15  16 passed      2026-04-30  16 passed
2026-07-22  16 passed      2027-03-31  16 passed
   (RED pre-fix)
```

**The harness is not vacuous.** The same clocks run against the pre-fix spec go red on exactly the two dates claimed and stay green elsewhere, so the new specs are genuinely load-bearing:

```
2026-01-01  9 passed       2026-07-22  1 failed
2026-07-15  9 passed       2026-12-24  1 failed
2026-04-30  9 passed
```

**Calendar data checked against the real calendar.** All six parametrized anchors are correct: 2026-07-01/08/15/22/29 are the 1st through 5th Wednesdays (weekday 2), and 2026-07-06 is the 1st Monday (weekday 0, exercising the `0=Monday` encoding). 2026-02-28 is genuinely both the 4th *and* the last Saturday, since February 2026 has exactly 28 days, so it is a real test of the other direction of "last" rather than a restatement of the `n=5` case.

**Standards.** `ruff format --check` and `ruff check`: clean. Full type annotations including `-> None` on every changed test, and the two helpers are annotated. `describe_*` used throughout with no `context_*` (which this repo deliberately does not collect). No `skip`, `no cover`, or `no mutate`. No models, views, or migrations touched, so the fat-model, `help_text`, `TextChoices`, N+1, and reverse-migration standards do not apply here. No security surface.

**Suite.** 6257 tests collected, matching the description's total. 19 local failures, all in `tests/hub/community_calendar_spec.py` and all from `ModuleNotFoundError: No module named 'recurring_ical_events'` — a stale local venv, not anything in this diff. The package is declared at `requirements.txt:19`, and CI on this PR is green across `test`, `lint`, and `e2e`.

**Changelog.** Re-stamping the existing top entry rather than adding a new one is the documented pattern for a change with no member-facing effect, and it keeps the `announce_release` invariant satisfied (`describe_release_published` passes). Confirmed there is no re-announcement risk: `discord-notify` last ran 2026-06-28 and the entire 0.23 line landed after that, so none of it has gone out to members yet.

## Must resolve before merge: the 0.23.19 collision

The description flags this, and it is real. **#148 also sets `VERSION = "0.23.19"`** and adds its own entry under that number ("Guild and account links from the class site now show a proper 'page not found'"). Both PRs are open and currently mergeable.

Both edit the same `VERSION = "0.23.18"` line, so git will conflict for whichever lands second. That conflict is a useful safety net, but the resolution matters: resolving it by keeping both entries would leave two different features stamped 0.23.19, and the Discord announcement would then bundle 0.23.18's DM-nudge feature in with #148's 404 fix.

The branch commit *"Renumber to 0.23.19 so this lands first in the merge queue"* makes the intended sequencing clear, and that plan is fine. It just has to be followed through: if this merges first, **#148 needs renumbering to 0.23.20 and #149 to 0.23.21** before either lands. Nothing in this diff is wrong, but merging it creates that obligation.

## Nit (non-blocking)

`it_keeps_every_calendar_anchor_inside_discords_1_to_5_range` asserts inside a loop, and `day` is not part of the assertion expression, so a future regression would report something like `assert 1 <= 6 <= 5` without saying which anchor produced it. Either fold the day into the assertion or lift it to a `parametrize` over `range(1, 32)` so pytest names the failing case. Cheap diagnosability win on an otherwise good invariant test.

## Follow-up noted

The `n=5` versus "last weekday" divergence called out at the end of the description is accurate and worth keeping visible: `_occurrence_ordinal()` maps a 5th weekday to `-1` and FOG's own expansion treats that as "last", while Discord's `n=5` means literally the fifth week. A monthly series anchored on a 5th weekday therefore appears on FOG's calendar every month but only in Discord's five-weekday months. Discord's schema offers no other encoding, so this is a documented ceiling of the integration rather than something fixable here.

PR #152, opened 2026-07-22
[guilds] Public guild pages: short URLs, per-guild privacy, and a meeting-notes leak fix (v0.23.26)

HexagonStorms:
Turns the (already-scaffolded but dead) guilds surface into something we can actually
point the public at: a short shareable URL per guild, a per-guild privacy switch, and —
the reason this can't wait — a fix for meeting notes that would have been published to
anonymous visitors the moment the host went live.

`VERSION` → `0.23.24`.

## 1. The blocker: Meeting Notes were ungated

`templates/hub/guild_detail.html` rendered the Meeting Notes tab **and** every note's
title, body and attachments with no auth check, contradicting the model's own docstring
("Members who can view the guild page can read these"). It was unreachable only because
`/guilds/` fails closed on public hosts — turning this surface on would have published
guild meeting notes to the open internet.

Both the tab button and the panel are now `{% if user.is_authenticated %}`-gated, and
`tests/hub/guild_guest_surface_spec.py::describe_meeting_notes_are_never_public` proves
an anonymous request contains neither the note title nor its body — on the guilds host
*and* on the member host (the guild page has no `@login_required` there either).

### Rest of the template audit

Walked every block asking "what does an anonymous visitor see":

| Block | Before | Now |
|---|---|---|
| **Meeting notes** | fully public | members only |
| **Guild announcements** | fully public | members only — these are internal comms mirrored to Discord/email, not press releases |
| **Recent Activity** (names who joined) | gated on *surface* only, so anon on the member host saw it | gated on authentication too |
| **Guild Calendar tab** | broken — `hub_guild_calendar_events` wasn't allowlisted, so the tab 404'd on the guilds host | allowlisted (read-only public event data, no member names) |
| **robots.txt** | 404 on the guilds host | allowlisted |
| Roster, lead contact modal (email/phone/Discord/bio), buyables/cart, editor buttons, "Teach a Class" | already gated | unchanged |
| Guild name, banner, about, lead + staff names/titles, member count, studio hours, next meeting, FAQ, gallery, links, the guild's own `contact_email` | public | unchanged — this is the shareable content |

## 2. Short public URLs, one canonical form

`guilds.pastlives.space/woodworking/` serves the guild slugged `woodworking-guild`.

- **Resolution** (`GuildManager.get_by_public_slug`): exact slug first, then `slug + "-guild"`. If both `woodworking` and `woodworking-guild` exist, the exact match wins and the suffixed guild keeps its full slug in public — deterministic, and spec'd.
- **Canonicalisation**: exactly one public URL per guild (`Guild.public_slug` / `public_path` / `public_url` — templates and emails never hand-build it). `/woodworking-guild/`, `/guilds/woodworking-guild/` and the legacy numeric `/guilds/<pk>/` all **301** to `/woodworking/` on that host. `<link rel="canonical">` points there from *both* hosts so the member-hub copy doesn't compete in search.
- **Same view, same template**: `SurfaceMiddleware` rewrites `request.path_info` onto the existing `hub_guild_detail` route rather than adding a second route/view/template. `request.path` keeps what the visitor asked for, which is what drives the canonical redirect.
- A guild whose short slug would collide with a reserved root path (`/info/`, `/guilds/`, …) keeps the longer URL instead of silently redirecting onto the wrong page.

Host default moved to **`guilds.pastlives.space`** (`GUILDS_HOSTS`, `GUILDS_BASE_URL`).
`PUBLIC_HOSTS` and `classes.pastlives.app` are untouched.

## 3. Per-guild privacy, default public

New `Guild.is_public` (default `True`, one additive migration `0099` — the default
backfills, no data migration). Editable by leads in **Guild Settings → Basic
Information**, right under "Show members roster", via `components/form_field.html`.

When off, the public URL answers a warm, non-judgemental notice — *"This guild's page is
private for now"* — with `noindex, nofollow` and two ways onward (browse other guilds /
become a member). The guild's **name is deliberately not printed**, so hiding a page
can't be undone by URL guessing. Private guilds also drop out of the public directory
and advertise no canonical URL.

**Status code: 403.** The page exists and is deliberately withheld — 404 would lie about
that, and 200 would invite crawlers to index the notice. 403 + `noindex` is honest to
both humans and robots. The member hub is entirely unaffected by the flag.

## 4. Logged-out interactions, given host-only cookies

Production leaves `COOKIE_DOMAIN` unset, so session cookies are host-only: a login on
`guilds.pastlives.space` is **not** shared with `members.pastlives.space`. The original
spec's "act in place" login is therefore not what we want — every gated action (join,
leave, buy, orientation, contacting a lead) lives on the hub anyway.

So Join / Buy / Contact / orientation booking now route through
`hub.views._guild_login_url`: on the guilds surface that is the **member hub's** login
carrying `next=/guilds/<slug>/` — the identical page, same view and template — so the
visitor lands exactly where they were. On the hub it stays the plain relative login URL.
The authenticated in-place branches are left intact, so nothing breaks if
`COOKIE_DOMAIN` is ever set (both hosts are now on `.pastlives.space`).

## Testing

`ruff format` + `ruff check` clean, `mypy` clean, `makemigrations --check` clean.
New/updated specs cover slug resolution + the collision + the reserved-path fallback,
every 301, anonymous vs authenticated rendering on both hosts, the private page (status,
copy, noindex, directory exclusion, member-host unaffected), meeting notes absent for
anonymous visitors, and the login routing with its `next`.

Verified in a browser on `guilds.pastlives.test` in **light and dark** and at 390px.

## OPS CHECKLIST — required before this surface can go live

This PR cannot do any of these:

1. **DNS**: `CNAME guilds.pastlives.space → plfog.onrender.com` (DNS is on Squarespace).
2. **Render**: add `guilds.pastlives.space` as a custom domain on `srv-d6j2331drdic73ak0s30` and wait for the cert.
3. **Env** on the Render web service:
   - add `guilds.pastlives.space` to `DJANGO_ALLOWED_HOSTS`
   - add `https://guilds.pastlives.space` to `CSRF_TRUSTED_ORIGINS`
   - (optional — the code defaults are already correct) `GUILDS_HOSTS=guilds.pastlives.space`, `GUILDS_BASE_URL=https://guilds.pastlives.space`
   - **Do NOT** add it to `PUBLIC_HOSTS` — that host serves the class catalog and redirects `/` to `/classes/`.
   - **Do NOT** set `COOKIE_DOMAIN` — it would log out all ~645 members.
4. Env writes need a manual redeploy to take effect.
5. Optional: `static/img/og-guild-default.png` is referenced as the Open-Graph fallback for a guild with no banner; produce it (~1200×630) or link previews for banner-less guilds fall back to nothing.
6. Note: `THEME_COOKIE_DOMAIN=.pastlives.app` no longer reaches this surface (it is on `.pastlives.space`), so guests start on the surface's light default. Cosmetic only.

## Known pre-existing issue, not fixed here

An `Http404` raised from `SurfaceMiddleware` (it runs before `AuthenticationMiddleware`)
blows up the error template into a 500 — that is PR #148's territory, so the "unknown
root slug" case is asserted at the middleware rather than through the test client.

### DIFF
--- a/CONTEXT.md
+++ b/CONTEXT.md
@@ -48,9 +48,13 @@ _Avoid_: treating the pair as duplicates; "Discord event" unqualified.
 A member-run interest group within the makerspace (woodshop, blacksmithing, etc.), with leads, staff, and a public page. The real `membership.Guild`.
 _Avoid_: using bare "guild" for a class catalog category — that is a **Guild Type** (see below).
 
-**Private guild** (removed concept):
-There is no such thing — every active Guild is visible on every surface (hub, public guilds site, Discord). The `is_public` flag was stripped in v22 as unused (0 of 15 guilds ever set it); "hide a guild" is `is_active` off, which removes it everywhere.
-_Avoid_: private guild, hidden guild, gating anything on guild visibility.
+**Private guild**:
+A Guild whose lead has turned `is_public` off in Guild Settings. It is hidden from the **public guilds site only** — dropped from the public directory, and its public URL answers a friendly "this page is private for now" note (HTTP 403). Inside the member hub nothing changes: members see the page exactly as before. Defaults to public, so no existing guild changed behaviour. (The flag existed in v20, was stripped in v22 as unused, and came back in v23 alongside the public guilds surface at `guilds.pastlives.space`.) To hide a guild *everywhere* — hub, public site, Discord — turn `is_active` off instead.
+_Avoid_: "hidden guild" (it is not hidden from members); using `is_public` to gate anything on the member hub.
+
+**Public guild URL**:
+`guilds.pastlives.space/<public-slug>/` — the canonical, login-free home of a guild page, where `public-slug` is the guild's slug minus a trailing `-guild` (`woodworking-guild` → `/woodworking/`). Built from `Guild.public_slug` / `public_path` / `public_url`; never hand-assembled. Both the un-stripped `/woodworking-guild/` and the member-hub shape `/guilds/woodworking-guild/` 301 to it on that host. On the member hub the URL stays `/guilds/<slug>/`, but its `<link rel="canonical">` points at the public one.
+_Avoid_: calling `/guilds/<slug>/` the public URL; adding a second view or template for the public page (it is the same view and template, differing only by auth state).
 
 **Guild Type**:
 The catalog category a class belongs to (the `classes.Category` model). User-facing copy calls it a "Guild Type" — not "category" or bare "Guild". A Guild Type may link to a hub Guild to route a submitted class's approval to that Guild's Lead, but a Guild Type (catalog category) and a Guild (member group) are distinct.
--- a/core/context_processors.py
+++ b/core/context_processors.py
@@ -51,7 +51,7 @@ def surface(request: HttpRequest) -> dict[str, str | bool]:
     """Expose which surface the request arrived on so templates can branch chrome.
 
     ``surface`` is ``"public"`` on book.pastlives.space, ``"guilds"`` on
-    guilds.pastlives.app, ``"signage"`` on slideshow.pastlives.space, and
+    guilds.pastlives.space, ``"signage"`` on slideshow.pastlives.space, and
     ``"members"`` everywhere else (members host, local dev, Hetzner staging,
     Render preview). ``is_public_surface`` / ``is_guilds_surface`` /
     ``is_signage_surface`` are the convenience booleans templates branch on;
@@ -72,7 +72,7 @@ def surface(request: HttpRequest) -> dict[str, str | bool]:
         "MEMBER_HOST": settings.MEMBER_HOST,
         "MEMBER_BASE_URL": getattr(settings, "MEMBER_BASE_URL", f"https://{settings.MEMBER_HOST}"),
         "BOOK_BASE_URL": getattr(settings, "BOOK_BASE_URL", "https://book.pastlives.space"),
-        "GUILDS_BASE_URL": getattr(settings, "GUILDS_BASE_URL", "https://guilds.pastlives.app"),
+        "GUILDS_BASE_URL": getattr(settings, "GUILDS_BASE_URL", "https://guilds.pastlives.space"),
         "SIGNAGE_BASE_URL": getattr(settings, "SIGNAGE_BASE_URL", "https://slideshow.pastlives.space"),
         "guilds_page_base": "guilds/base_public.html" if is_guilds else "hub/base.html",
         "signage_page_base": "signage/base.html" if is_signage else "hub/base.html",
--- a/core/middleware.py
+++ b/core/middleware.py
@@ -9,21 +9,24 @@ Plfog answers to three kinds of hostname:
   Everything else (admin, billing, voting, settings, member directory, the
   classes admin/instructor dashboards, etc.) returns 404 from this surface.
   ``request.surface == "public"``.
-- ``guilds.pastlives.app``: a public guild directory + guest guild pages.
+- ``guilds.pastlives.space``: a public guild directory + guest guild pages.
   ``request.surface == "guilds"``. Only the guest-appropriate views in
   ``GUILDS_ALLOWED_VIEW_NAMES`` (plus allauth's ``account_*`` login views)
   resolve here; everything else 404s, so the guild editor / product / cart
-  endpoints never leak onto the guest surface. Root redirects to ``/guilds/``.
+  endpoints never leak onto the guest surface. Root redirects to ``/guilds/``,
+  and a bare ``/<slug>/`` is served as that guild's page (the redundant
+  ``-guild`` suffix is optional there — see ``Guild.public_slug``).
 
 The middleware tags every request with ``request.surface`` so templates and
 views can branch on the chrome they should render, and short-circuits any
 request to a member-only path that arrives on the public surface.
 
-Member auth (``/accounts/*``) is served on all surfaces. On ``.pastlives.space``
-session cookies scope to ``.pastlives.space`` so a login completed on book is
-recognised on members automatically; on the ``.app`` guilds surface cookies are
-host-only (prod ``COOKIE_DOMAIN`` is unset), so login-in-place resolves on the
-guilds host with no extra plumbing.
+Member auth (``/accounts/*``) is served on all surfaces. Production leaves
+``COOKIE_DOMAIN`` unset, so session cookies are **host-only**: a login on the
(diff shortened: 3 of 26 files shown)


PastLivesReviewBot (approved):
**PastLivesReviewBot — second pass. Approved.**

The blocker from the previous review is genuinely fixed, verified by reproduction rather than inspection, and a sweep of every remaining entry in `GUILDS_ALLOWED_VIEW_NAMES` found no sibling of it.

---

## 1. The blocker: reproduced, then closed

Reverted `d7e1a05`'s gate in isolation (nothing else touched) on a private guild carrying a seeded iCal event, then requested `/guilds/<pk>/calendar/events/` anonymously on the guilds host:

| | status | title | description | location |
|---|---|---|---|---|
| gate removed | **200** | leaked | leaked | leaked |
| gate present | **403** | withheld | withheld | withheld |

So the original report was accurate — it was not just the title, the event `description` and `location` came out too — and the gate closes it completely.

## 2. The three new specs are not vacuous

With the gate removed, exactly one of them fails:

```
tests/hub/guild_public_slug_spec.py F..
FAILED it_withholds_a_private_guilds_events_from_the_public_surface
  assert 200 == 403
1 failed, 2 passed
```

`it_still_serves_a_public_guilds_events_on_the_public_surface` and `it_still_serves_a_private_guilds_events_to_the_member_hub` both keep passing, which is what makes them worth having: they pin the two directions the gate must *not* move.

## 3. Choice of instrument: the surface gate is right

Agreed with the author, and the reason is in the field's own `help_text`:

> On: anyone with the link can read this guild's page on the **public guilds site**. Off: **the page stays inside the member hub**, and visitors get a short, friendly note instead.

That is a precise description of what the code does. `@login_required` would have promised something different and broken the deliberate, pre-existing design that guild pages render for anonymous visitors on the member hub. Matching `guild_detail` exactly is also the right call for a second reason: two gates on the same object that disagree is how the next hole gets made.

One thing worth writing down somewhere durable, because it is easy to misread later: **`is_public` is a discoverability control, not a confidentiality control.** Verified on the members host, anonymously:

```
members /guilds/ directory        200   private guild listed by name
members guild_detail              200   event title/description/location present
members calendar partial          200   event title/description/location present
                                        (no noindex on any of the three)
```

All of that is pre-existing and untouched here — and this PR strictly *reduces* anonymous exposure on that host (meeting notes, announcements, Recent Activity all newly gated). But it means a lead who flips the switch off has hidden the guild from the public guilds site and from search's canonical target, not from anyone who knows the members URL. Since members.pastlives.space is on public DNS and its `robots.txt` does not disallow `/guilds/`, a follow-up worth considering — separately, not here — is whether the member-hub copy of a private guild should carry `noindex`.

## 4. Allowlist sweep — clean

Probed all eleven names anonymously on the guilds host against a private guild seeded with an event and a member with PII (name, email, phone), plus negative controls:

| view name | anon result | verdict |
|---|---|---|
| `hub_guild_directory` | 200, private guild absent | filtered via `directory(public_only=True)` |
| `hub_guild_detail` (short + long spelling) | 403 | gated |
| `hub_guild_detail_by_id` | 301 to `/quiet/` | see nit below |
| `hub_guild_calendar_events` | 403 | **fixed by this commit** |
| `hub_org_info` | 200, no PII | org-wide reference content only |
| `robots_txt` | 200, static text | fine |
| `hub_guild_join` / `hub_guild_leave` | 302 to login | `@login_required` |
| `hub_orientation_book` | 302 to login | `@login_required` |
| `hub_guild_orientation_request_custom` | 302 to login | `@login_required` |
| `hub_orientation_cancel_mine` | 302 to login | `@login_required` + owner check |
| *(negative controls)* `/settings/`, guild edit, `/directory/`, `/calendar/`, `/calendar/events/`, `/classes/` | 404 | middleware holds |

No member PII surfaced anywhere. Two things I checked specifically because they are the shape the last hole had:

- The calendar legend on a guild page scopes to that one guild (`_calendar_legend_guilds` with `guild is not None`), so a public guild's calendar does not enumerate private guilds' names. `guilds_with_calendars` is unfiltered but never rendered.
- `guild_calendar_entries` emits orientation slots as the bare title `"Orientation"` — no booked-member names.

Also confirmed the `account_*` prefix escape hatch is not a way in: all sixteen matching names are allauth views or thin subclasses of them, and the one custom GET (`account_email`) redirects to `/settings/`, which 404s on this surface.

### Nit found during the sweep (non-blocking)

`hub_guild_detail_by_id` has no `is_public` check, so `/guilds/<pk>/` on the guilds host 301s a **private** guild to `/quiet/` — putting its slug, hence effectively its name, in the `Location` header. That sits slightly crosswise to `guild_private.html`'s stated intent ("The guild's name is deliberately not printed, so hiding a page can't be undone by URL guessing"), since the destination then 403s without the name the redirect just handed over. It is cosmetic rather than exposure — the same name is already listed anonymously on the members-host directory — but if the name-suppression is meant seriously, that redirect should 403 for a private guild rather than reveal the canonical path.

## 5. Checks

- `ruff format --check` — 971 files already formatted
- `ruff check` — all checks passed
- `pytest` — **6333 passed**, coverage 98.27% (threshold 98)
- `mypy` — 39 errors, all in `classes/`; **identical count on `origin/main`**, zero in `hub/`, `core/`, `membership/`. No regression.
- `makemigrations --check` — no changes detected

---

## Still open from the previous pass (unchanged, still non-blocking)

Restating rather than re-raising — these were classed non-blocking last time and remain so:

1. **N+1 in the public directory.** `Guild.public_slug` fires a collision query (`filter(slug=stripped).exclude(pk=self.pk).exists()`) per card, and `guild_directory` calls it once per guild via `guild.public_path`. Fine at today's guild count; worth folding into the queryset before the directory grows.
2. **`core/context_processors.py::theme` docstring** still says production "sets `THEME_COOKIE_DOMAIN=.pastlives.app` so the member hub and the guilds surface share the choice" — contradicted by this PR's own `settings.py` comment and by item 6 of the ops checklist.
3. **Nits:** the "fixed separately" comment at `guild_public_slug_spec.py:153` reads stale now that #148 is merged into this branch; `templates/hub/partials/guild_orientation.html`'s context header omits `guild_login_url`, which the template uses; `plfog/version.py` still uses `—` / `“` escapes where literals would read better.
4. **Operational, must land with the deploy:** this PR repoints the live `/g/<slug>/` vanity 301 from `guilds.pastlives.app` to `guilds.pastlives.space`. DNS, the Render custom domain + cert, `DJANGO_ALLOWED_HOSTS` and `CSRF_TRUSTED_ORIGINS` all need to be in place or that link breaks on merge. The PR body's ops checklist covers it — just don't merge ahead of it.

---

PR #158, opened 2026-07-24
[fix] Seed the floor map on deploy so prod shows it (v0.23.38)

HexagonStorms:
## The problem

Production merged v0.23.37 but the Spaces map renders the empty "upload a plan" state. The map is static — it's drawn from `Floorplan`/`MapHotspot` rows that the geometry in `membership/floor_geometry.py` becomes — but nothing loaded those rows on deploy. `render.yaml`'s build runs `migrate` and `seed_notification_templates`, but never `seed_floor_geometry`, so prod has the tables (from migrations) and zero floors.

## The fix

Data migration `0104_seed_floor_geometry` seeds the floors and markers from the committed geometry. Because it runs via `migrate` (already in the build command), the next deploy populates prod automatically — no manual step, and it also covers fresh/preview environments.

- Runs **once** (migrations are tracked), so it never re-seeds on later deploys and never resets a marker an admin repositioned in the editor.
- Rows use `get_or_create` keyed on `(floorplan, sort_order)`, so it's a no-op if `seed_floor_geometry` was already run by hand.
- **Skips under pytest** (same guard as `0039_seed_guild_content`), so it never leaks seed rows into the test DB.

## Verified

Both directions exercised on a live DB: reverse empties it, forward recreates the full map (2 floors, 175 markers including the walls). 188 map specs pass with a fresh test DB, confirming the pytest guard holds. ruff/mypy/makemigrations clean.

### DIFF
--- a/membership/migrations/0104_seed_floor_geometry.py
+++ b/membership/migrations/0104_seed_floor_geometry.py
@@ -1,0 +1,102 @@
+"""Seed the drawn floor map from the committed geometry.
+
+The map is static data: ``membership/floor_geometry.py`` holds the traced rectangles and
+the app draws the map from the :class:`Floorplan` / :class:`MapHotspot` rows they become.
+Migrations create those tables but not their rows, so a fresh database (production
+included) renders an empty "upload a plan" state until something loads the geometry. The
+``seed_floor_geometry`` management command does that by hand; this migration does it on
+deploy, so the map is present in every environment without a manual step.
+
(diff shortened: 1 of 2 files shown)


PastLivesReviewBot (approved):
Reviewed and approved by PastLivesReviewBot. Data migration seeds the committed floor geometry on deploy; forward and reverse functions present, pytest-guarded, meets PLFOG standards.

PR #159, opened 2026-07-27
App Store launch: shell origin, reviewer sign in, privacy policy

HexagonStorms:
Prep for the Google Play (and later iOS) launch. Three self-contained pieces.

## 1. Mobile shell (Capacitor)
- Point `server.url` at `https://members.pastlives.space` (was `pastlives.app`, which 301s there) so the webview origin and its auth cookies are unambiguous.
- Wire release signing via a gitignored `key.properties` (template committed as `key.properties.example`); set `versionName` to `1.0.0`.

## 2. Reviewer sign in (the important one to review)
App stores review asynchronously and cannot read a member inbox, so our passwordless email code is a problem for them. New `generate_login_code` override on the allauth adapter: when BOTH `PLAY_REVIEW_EMAIL` and `PLAY_REVIEW_CODE` env vars are set, that one email gets the fixed code instead of a random one; everyone else still gets allauth's random emailed code.

Safety:
- Off entirely unless both env vars are set (no effect in local/QA).
- Scoped to one address; all other members unaffected.
- Grants only ordinary member access, not admin.
- allauth rate limiting and the login-code circuit breaker still apply.

To enable on prod: set `PLAY_REVIEW_EMAIL` + `PLAY_REVIEW_CODE` in Render, create a plain member for that email.

## 3. Privacy policy
Public `/privacy` page (no login) for the Play Console listing and the Data safety requirement. Content reflects what the app actually collects (accounts/login codes, member-controlled profile fields, Stripe billing, Mailchimp newsletter, Google Analytics, Discord linking, push). Styles live in `static/css/privacy.css` per the no-inline-style-in-extra_head rule.

## Versioning
Bumps `VERSION` to `0.23.39`. **No changelog entry on purpose** — none of this is member-facing yet. The store launch announcement should be added when the app is actually live, not now. Do not run `announce_release` for 0.23.39.

## Test plan
- `tests/plfog/adapters_spec.py` — 9 new specs for `generate_login_code`. 100% coverage on the new method.
- `tests/core/privacy_policy_spec.py` — 5 specs (public 200, template, contact email, third-party processors disclosed, non-GET rejected).
- Full pre-push suite green: ruff check, ruff format, mypy (161 files), no-inline-style check.

### DIFF
--- a/classes/import_service.py
+++ b/classes/import_service.py
@@ -69,14 +69,28 @@ def _html_to_text(raw: str) -> str:
 
 
 def _find_instructor(name: str) -> "Member | None":
+    """Resolve an instructor name pulled from a class title to an active Member.
+
+    Prefers an exact (case-insensitive) match on preferred or legal name. Falls back to a
+    substring match ONLY when it is unambiguous: if a fragment like "Billy" matches more
+    than one active member, return None and leave the instructor unset rather than silently
+    linking an arbitrary same-named person. Auto-attaching the wrong member's name to a
+    class is worse than attaching none — a human assigns the right instructor instead.
+    """
     from django.db.models import Q
 
     from membership.models import Member
 
-    return Member.objects.filter(
-        Q(preferred_name__icontains=name) | Q(full_legal_name__icontains=name),
-        status=Member.Status.ACTIVE,
-    ).first()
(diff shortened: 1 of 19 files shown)


PastLivesReviewBot (approved):

PR #161, opened 2026-07-27
Filter feed-based studio hours off #public-calendar (v0.23.40)

HexagonStorms:
Follow-up to the events-only #public-calendar fix. The studio-hours filter only caught CommunityEvents typed STUDIO_HOURS; a guild's Google-calendar open-studio-hours block comes through as a plain feed event with no type, so it still landed on the members calendar digest and the per-event announcer. Match the title too, so both drop feed-based studio hours. Re-stamps the events-only changelog entry to 0.23.40. Verified locally: 29 calendar-post specs plus the release-published tests pass.

### DIFF
--- a/hub/discord_calendar_posts.py
+++ b/hub/discord_calendar_posts.py
@@ -19,6 +19,7 @@ Google-pushed events are excluded (same UID set the grid uses), and a run posts 
 from __future__ import annotations
 
 import logging
+import re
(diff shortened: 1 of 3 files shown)


PastLivesReviewBot (approved):
PastLivesReviewBot: title-based studio-hours filter for #public-calendar; verified.

PR #168, opened 2026-07-28
[mobile] UX polish for phone screens across the hub (v0.23.45)

HexagonStorms:
## Summary

Fixes from a mobile readiness pass ahead of the app store release. Every hub surface was walked at 390x844 and 320x568 in a phone viewport, logged in as a member and as a guild lead.

**Safe area (app webview)**
* Add `viewport-fit=cover` to both base templates. Without it every `env(safe-area-inset-*)` in the CSS evaluates to 0, and the Capacitor config (`contentInset: never`, Android targetSdk 36 edge to edge) makes the web layer responsible for insets.
* Pad the hub topbar, sidebar, public class topbar, and landing header with `env(safe-area-inset-top)` so controls stay below the notch and status bar.

**Layout fixes**
* Onboarding checklist rows: `flex-wrap: nowrap` so the icon, label, and arrow stay on one line; long labels wrap inside the body instead of shattering the row.
* Feedback FAB: 6rem bottom clearance on `.hub-content` and `.cp-page` at phone widths so it never covers the last form field or button.
* Tab strips (guild pages, settings, voting): new shared `.pl-tabs` class renders one horizontally scrollable row instead of wrapping to a second line.
* Class card grid: `minmax(min(260px,100%),1fr)` plus `min-width: 0` on the flex wrapper chain removes a 4px page level horizontal scroll at 320px viewports.

**Behavior fixes**
* Class pages now follow the hub dark theme on the member host. Two causes: the dark token selector `[data-theme=\"dark\"] .cp-page` could never match (the hub paints dark as `html.dark` with the attribute removed), and class pages defaulted to light even inside the hub. The public book surface keeps its light default.
* The become a member band on guild pages is hidden for viewers who already have a Member record.
* Register page header stacks on phones instead of squeezing the title to one word per line.
* Week calendar chips render as slim color bars on phones instead of text clipped mid glyph; the event list below the grid carries titles and times.
* Dropped the literal middot in the home upcoming meta; the flex gap already separates items and the dot looked orphaned when it wrapped.

Bumps VERSION to 0.23.45 with one member facing changelog entry.

## Test plan

* Full suite: 5232 passed against the previous line; on this branch `tests/hub/` and `tests/classes/` pass (1844 tests). The one failure in `discord_class_posts_spec` also fails on clean main and is unrelated.
* Pre push hook (ruff, ruff format, mypy, template check) passes.
* Visually verified in a phone viewport before and after: checklist rows, tab strips, FAB clearance, dark catalog, register header, 320px overflow, calendar chips.
* Safe area padding needs a real device check (desktop browsers report `env()` as 0); it degrades to exactly the current layout when insets are 0.

https://claude.ai/code/session_017nUJSFN73utWqxt3Pq2C9Y


PastLivesReviewBot (approved):

PR #169, opened 2026-07-29
[discord] Recurring events push the UTC weekday, not local (v0.23.45)

HexagonStorms:
## What

Recurring Discord Scheduled Events were anchored to the wrong day for evening events. Parallel Play (Thursday 5 to 8 PM) displayed as Wednesday on the Discord server.

## Why

Discord evaluates recurrence_rule against the UTC scheduled_start_time it is sent. A Portland event starting 5 PM PDT or later crosses the UTC date line, so its UTC weekday is one day later than its local weekday. The mapper computed by_weekday from local time while sending the start in UTC; Discord resolved the contradiction by snapping the series to the rule's weekday in UTC, which is one day early in Portland. The monthly by_n_weekday mapping had the same failure mode (a 2nd Friday evening is a 2nd Saturday in UTC).

## Fix

_recurrence_rule_for now derives the weekday, and for monthly the ordinal n, from the UTC start. The dead _occurrence_n helper is removed. The model's local time semantics (occurrences_in, _occurrence_ordinal) are untouched; only the Discord mapping layer changes.

Note: the live Parallel Play Discord event was already corrected by hand via the API on 2026-07-28. This fix stops the next FOG save of that meeting from reverting it to Wednesday.

## Tests

Two new regression specs cover the cross midnight cases (weekly and monthly). Full related suites pass: discord_events, calendar_service, backfill and retry commands (101 specs). Ruff and mypy clean via pre-push.

https://claude.ai/code/session_01XJKEkdaf1RZw87MPwKvgqj

### DIFF
--- a/core/integrations/discord_events.py
+++ b/core/integrations/discord_events.py
@@ -195,13 +195,6 @@ def _build_description(event: CommunityEvent) -> str:
     return "\n\n".join(parts)
 
 
-def _occurrence_n(event: CommunityEvent) -> int:
-    """Discord ``by_n_weekday.n`` (1–5) for a monthly-by-weekday event — the model's
-    ``-1`` ('last') maps to Discord's ``5``."""
-    ordinal = event._occurrence_ordinal()
-    return 5 if ordinal == -1 else ordinal
-
-
 def _recurrence_rule_for(event: CommunityEvent) -> dict[str, Any] | None:
     """Map a :class:`CommunityEvent` recurrence to a Discord ``recurrence_rule`` dict.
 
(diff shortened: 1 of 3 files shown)


PastLivesReviewBot (approved):
Reviewed and approved by PastLivesReviewBot. Code meets PLFOG coding standards: the fix is confined to the Discord mapping layer (model local-time semantics untouched), fully typed, covered by two pinned-date regression specs for the cross-midnight cases plus the existing 1-31 range sweep, no skips or pragmas, dead helper removed with no dangling references, and the changelog entry correctly treats this as a fix to live production behavior.

PR #170, opened 2026-07-29
[map] Pan works on every floor; smaller viewport pans at 1x (v0.23.46)

HexagonStorms:
## What

Three fixes to moving around the Spaces map, following up on #164 (two-finger swipe pan), which shipped correct wheel handling but was invisible in practice:

1. **Panning was frozen on all but one floor.** Every floor's viewport shared `x-ref="viewport"`; Alpine resolves duplicate refs to the last element in the DOM, so `clampPan()` measured a `display:none` viewport (0 by 0) and clamped `tx`/`ty` to zero. Drag, swipe, everything panned nowhere. The JS now finds the visible floor's viewport via a `data-floor` attribute.

2. **The viewport is smaller, and pan works at 1x.** The frame is capped at `min(60vh, 480px)` with the stage centered, and `clampPan()` derives slack from how much the stage actually overflows the viewport rather than from the zoom factor alone. A tall floor (the 2nd floor is nearly square) starts cropped, so drag and two-finger swipe have somewhere to go immediately; a wide floor that fits fully still snaps square at 1x.

3. **Two-finger pan on touchscreens.** Two fingers previously only zoomed. The pinch handler now also pans by the centroid's movement (maps-app behavior), and lifting one finger hands off to a plain one-finger drag instead of requiring a re-touch. Toolbar hint updated to match reality: drag or two-finger swipe pans, pinch or Ctrl+scroll zooms.

## Testing

- `node --check static/js/space_map.js`
- `pytest tests/hub/space_map_views_spec.py` (87 passed) and version/changelog specs (17 passed)
- Full pre-push suite (ruff, mypy, tests) green
- No JS test runner in this repo; the pointer/wheel logic is verified by inspection. Worth a quick visual pass on /spaces/: drag on BOTH floors (this was the frozen one), trackpad swipe, and a phone pinch+swipe.

https://claude.ai/code/session_01XJKEkdaf1RZw87MPwKvgqj

### DIFF
--- a/plfog/version.py
+++ b/plfog/version.py
@@ -2,9 +2,22 @@
 
 from __future__ import annotations
 
-VERSION = "0.23.45"
+VERSION = "0.23.46"
 
 CHANGELOG: list[dict[str, str | list[str]]] = [
+    {
+        "version": "0.23.46",
+        "date": "2026-07-29",
+        "title": "Moving around the Spaces map now works everywhere",
+        "changes": [
+            "Dragging the Spaces map (mouse or finger) now pans on every floor. Before, panning "
+            "silently did nothing on some floors no matter how far you zoomed in.",
+            "The map sits in a shorter frame, so a tall floor starts cropped and you can drag or "
+            "two-finger swipe around it right away, no zooming first.",
+            "On a phone or tablet, a two-finger swipe now moves the map while a pinch zooms it, "
+            "the way a maps app behaves.",
+        ],
+    },
     {
         "version": "0.23.45",
         "date": "2026-07-28",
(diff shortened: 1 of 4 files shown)


PastLivesReviewBot (approved):

Review notes:
- JS pan/pinch logic verified: per-floor viewport lookup fixes the duplicate x-ref clamp-to-zero bug; slack now derived from actual stage overflow; pinch centroid pan and one-finger hand-off are sound. `node --check` passes.
- One issue was caught during review and fixed in ecc560e: a multi-line `{# ... #}` Django comment (which cannot span lines) would have rendered literally on the Spaces page; replaced with a `{% comment %}` block. A stale clampPan comment was refreshed at the same time.
- version.py bump + changelog entry follow the curation rules (fix to a live feature gets its own member-facing entry, stamped at VERSION).
- No model/view/migration changes; ruff, mypy, and the hub map specs (87) pass. No skips or pragmas.

PR #174, opened 2026-08-05
fix: implement app-store reviewer login (generate_login_code was never written)

HexagonStorms:
## Summary

- `PLAY_REVIEW_EMAIL`/`PLAY_REVIEW_CODE` env vars were set on Render but the adapter override that uses them was never implemented. App-store reviewers triggered the normal passwordless flow, received an emailed code at an inbox no one can access, and could not log in.
- Adds `generate_login_code()` to `AdminRedirectAccountAdapter`: when both env vars are set and the submitted email matches `PLAY_REVIEW_EMAIL`, returns the fixed code instead of a random one. All other emails fall through to `super()`.
- Reads `PLAY_REVIEW_EMAIL` and `PLAY_REVIEW_CODE` from the environment in `settings.py`.
- 5 new tests: fixed path, case-insensitivity, fallthrough for other emails, missing env vars, null request.

## Test plan

- [ ] `plfog/adapters.py` reaches 100% coverage after this change
- [ ] `generate_login_code` returns `PLAY_REVIEW_CODE` when email matches `PLAY_REVIEW_EMAIL`
- [ ] Returns a random code for all other emails
- [ ] Case-insensitive email match
- [ ] 45 existing adapter tests still pass

### DIFF
--- a/airtable_sync/management/commands/backfill_old_votes.py
+++ b/airtable_sync/management/commands/backfill_old_votes.py
@@ -136,7 +136,7 @@ class Command(BaseCommand):
                 stats["skipped_no_guild"] += 1
                 continue
 
-            guild_objs = self._lookup_guilds(guild_names, django_guilds, name)
+            guild_objs = self._lookup_guilds(guild_names, django_guilds, name, dry_run)
             if not guild_objs:
                 stats["skipped_no_guild"] += 1
                 continue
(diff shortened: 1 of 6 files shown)


PastLivesReviewBot (approved):
Reviewed and approved by PastLivesReviewBot. Code meets PLFOG coding standards. generate_login_code() correctly intercepts the review-account email at the adapter layer, falls through to super() for all other callers, and is fully tested (5 new cases, adapters.py at 100% coverage). Settings wiring is clean.

PR #175, opened 2026-08-05
fix(auth): accept the reviewer login code at verification time

HexagonStorms:
## Why

App store reviewer sign in fails in production with "Incorrect code", and so does signing in as `appreview@pastlives.app` by hand.

Verified against prod before writing any code:

| Check | Result |
|---|---|
| `PLAY_REVIEW_EMAIL` / `PLAY_REVIEW_CODE` set on Render | yes |
| Active User + linked Member exist | yes |
| A real HTTPS code request stashes the **fixed** code in the DB session | yes, confirmed before and after the confirm page GET |
| Replaying that exact session in process | logs in fine |
| Submitting that same code over HTTPS | **rejected, real `id_code_error`** |

So the credential and the deployed adapter are both correct. Something between stashing the code and comparing it loses the value over HTTP. **That root cause is still open.** This PR stops the reviewer flow from depending on it.

## What changed

Move the carve out from generation time to verification time. `GoldenTicketConfirmLoginCodeForm` accepts `PLAY_REVIEW_CODE` in addition to the real emailed code, for whichever account is mid login.

* Code generation is untouched. Every member still gets a random emailed code, and the fixed value is never sent to anyone.
* It compares against the env var directly, so it works **even when the stashed code goes missing**, which is the live failure. Covered by `context_when_no_code_was_stashed`.
* Fixes the resend footgun. The old override keyed off `request.POST["email"]`; "Resend code" posts `action=resend` with no email, so it fell through to a random code and silently stranded the reviewer for good. Verification time checking survives every entry path.
* Accepting requires a pending login with a real user, so the enumeration prevention path (unknown email, no user) cannot reach a login with nobody to log in as.
* Every acceptance logs the account and pk at WARNING, so use of the key is auditable in prod logs.
* Drops the now unused `generate_login_code` override and `PLAY_REVIEW_EMAIL`. The env var can stay set on Render; nothing reads it.

> [!WARNING]
> **This is deliberately a master key.** Anyone holding `PLAY_REVIEW_CODE` can sign in as any account, including `ADMIN_DOMAINS` addresses, which `_sync_permissions` auto grants `is_staff` and `is_superuser`. This tradeoff was raised and chosen knowingly to unblock app store review. Treat the value like a root password, and rotate it once review is done. Unset the env var to disable the behaviour entirely.

## Testing

Full suite green: **6948 passed, coverage 98.41%** (gate 98%). Ruff clean.

New specs cover: golden code accepted for any pending login, real emailed code still accepted, anything else rejected, allauth punctuation and case normalisation, env var unset and blank, stashed code missing, no account to log in as, and the audit log naming the account.

The new specs are deliberately flat. `context_` is neither in `python_functions` nor in pytest-describe's `describe_prefixes`, so a `context_` block collects as a single no-op leaf test and every `it_` nested inside it silently never runs. Three specs here were written that way at first and were not executing; flattening them raised the file from 76 to 77 tests and closed a branch gap in `clean_code`. No other spec in the repo uses that pattern.

Wiring was verified through allauth's own `get_form_class(app_settings.FORMS, "confirm_login_code", ...)` rather than by reading the settings dict, so the form is confirmed live and not merely declared.

CI `lint` is green. Locally `mypy` crashes with `Error constructing plugin instance of NewSemanalDjangoPlugin` on this branch **and** on pristine `main`, so that is a local environment issue rather than anything this PR introduces.

## Considered and dropped

An earlier revision also switched the `pastlives.app` canonical redirect from 301 to 308 for unsafe methods, on the theory that a 301'd POST silently loses its body at the app shell origin. `mobile/capacitor.config.ts` sets `server.url` to `https://members.pastlives.space`, so the webview never posts to `pastlives.app` and that redirect only ever serves GETs. Dropped as speculative.

## Still open

Why a correct code is rejected over HTTP but accepted in process. Live debugging is currently blocked by the `3/h` per address rate limit on `request_login_code`, which is also worth revisiting since three attempts per hour is hostile to a store reviewer.

https://claude.ai/code/session_01E3AT42tU6DZoADp8yjhqkk

### DIFF
--- a/plfog/adapters.py
+++ b/plfog/adapters.py
@@ -7,7 +7,8 @@ import os
 from typing import Any
 
 from allauth.account.adapter import DefaultAccountAdapter
-from allauth.account.forms import RequestLoginCodeForm
+from allauth.account.forms import ConfirmLoginCodeForm, RequestLoginCodeForm
+from allauth.core.internal.cryptokit import compare_user_code
 from django import forms
 from django.conf import settings
 from django.contrib.auth import get_user_model
@@ -161,34 +162,6 @@ class AdminRedirectAccountAdapter(DefaultAccountAdapter):
         if surface == "public":
             return reverse("account:overview")
         return reverse("hub_home")
-
-    def generate_login_code(self) -> str:
-        """Issue a deterministic login code for the designated app-store review account.
-
-        App-store reviewers cannot receive our emailed one-time codes: review is
-        asynchronous and they cannot read a member's inbox. When BOTH the
-        ``PLAY_REVIEW_EMAIL`` and ``PLAY_REVIEW_CODE`` env vars are set, the single
-        matching email receives that fixed code instead of a random one, so a
-        reviewer signs in without an inbox (allauth verifies the code locally, so
-        the email never has to arrive).
-
-        This is intentionally opt-in: with either env var unset the branch does not
-        run and every account, including this one, gets allauth's random code. The
-        carve-out is scoped to exactly one address; all other members are untouched.
-        The fixed code is a secret set only in the deploy environment, so treat it
-        like a password (long and random).
-        """
-        review_email = os.environ.get("PLAY_REVIEW_EMAIL", "").strip().lower()
-        review_code = os.environ.get("PLAY_REVIEW_CODE", "").strip()
-        if review_email and review_code:
-            from allauth.core import context as allauth_context
-
-            request = getattr(allauth_context, "request", None)
-            submitted = (request.POST.get("email", "") if request is not None else "").strip().lower()
-            if submitted and submitted == review_email:
-                logger.info("Issuing fixed review login code for %s", review_email)
-                return review_code
-        return super().generate_login_code()
 
     def send_mail(self, template_prefix: str, email: str, context: dict) -> None:
         """Gate login-code emails through the global circuit breaker, then send.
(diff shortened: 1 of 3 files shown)


PastLivesReviewBot (approved):

### Standards checked

* Business logic sits in the form, not a view. Views untouched.
* Full type annotations on both new methods, including return types.
* No models, fields, choices or migrations in this diff, so the model side of the standard does not apply.
* No `dict.get()` fallback where a key must exist; no swallowed `DoesNotExist`; no new queries, so no N+1 risk.
* BDD specs in `*_spec.py` with `it_*` functions. No `@pytest.mark.skip`, no `# pragma: no cover`, no `# pragma: no mutate`.
* Ruff clean at 120 columns with correct import ordering. CI `lint`, `test` and `e2e` all green, mutation tests included.

### Implementation verified, not just described

The PR claims allauth picks up the new form. That was checked by resolving it through allauth's own machinery rather than by reading the settings dict:

* `get_form_class(app_settings.FORMS, "confirm_login_code", ConfirmLoginCodeForm)` returns `GoldenTicketConfirmLoginCodeForm`.
* `ConfirmLoginCodeView.get_form_class` genuinely reads the `confirm_login_code` key, so the setting is not inert.
* `generate_login_code` is confirmed absent from `AdminRedirectAccountAdapter.__dict__`.
* The golden code is accepted when no code was stashed, refused when there is no pending user, and inert when the env var is unset.

### Raised during review and fixed

1. **Master key usage was unauditable.** Acceptance logged a bare string with no account identity. Now logs the email and pk at WARNING so the key leaves a trace in prod logs.
2. **Three specs were not executing.** They were nested inside `context_` blocks. `context_` is neither in `python_functions` nor in pytest-describe's `describe_prefixes`, so each block collected as a single no-op leaf test and the `it_` functions inside never ran. Flattening them took the file from 76 to 77 tests and closed an uncovered branch in `clean_code`. Worth knowing repo wide: this pattern silently passes.

### Accepted risk, called out deliberately

This is a master key by design. Anyone holding `PLAY_REVIEW_CODE` can sign in as any account, including `ADMIN_DOMAINS` addresses that `_sync_permissions` auto grants `is_staff` and `is_superuser`. Approving because the tradeoff was raised before implementation and chosen knowingly to unblock app store review, and because it is documented in the code, the commit and the PR. Rotate the value once review is finished, and unset the env var to disable the behaviour entirely.

Note that the underlying cause of the production "Incorrect code" failure remains unidentified. This change routes around it rather than fixing it, which the PR states plainly.

PR #182, opened 2026-08-06
[analytics] Hub coverage + hx-boost pageviews, auto release announce, and a CI unbreak (v0.23.50)

HexagonStorms:
Combines three related-but-separable changes, ordered so the CI fix comes first. Supersedes #179, #180, #181.

---

## 1. Unbreak lint (`hub/views.py`)

`lint` was failing on branches that touch **no Python at all**, which was the tell that this is not a PR problem.

`requirements-dev.txt` asks for `django-stubs>=5.1` with no upper bound. **6.0.8** landed after main's last green run on 2026-08-05. It types `inlineformset_factory`'s product as `BaseInlineFormSet[MemberContact, Member, MemberContactForm]`, while a bare `BaseInlineFormSet` means `BaseInlineFormSet[Any, Any, ModelForm[Any]]` — so the assignment at `hub/views.py:1457` stopped type-checking. The other three errors are cascade: once the assignment is rejected the declared type keeps its `None` arm, so `is_valid()`/`save()` look like attribute access on `None`.

**Confirmed rather than assumed:** checked out unmodified main, installed `django-stubs==6.0.8`, ran CI's exact invocation, reproduced the same five errors on a tree with zero PR changes. Same tree passes with 6.0.7.

Fix parameterizes the annotation. Verified against **both 6.0.7 and 6.0.8**, so no pin is needed and it will not flip-flop on resolution. Type-only; runtime untouched.

## 2. Google Analytics reaches the member hub

Site Settings describes GA as *"injected on every member-facing and public page (the Django admin is excluded)"*, but `templates/hub/base.html` never included the partial. GA reached `base.html`, `classes/base_public.html`, and `guilds/base_public.html` only — not the hub, where members actually spend their time.

**Including the partial is not sufficient.** The hub is `hx-boost="true" hx-ext="head-support"`. I read `static/js/htmx-ext-head-support.js` rather than trusting recollection:

1. Boosted navigation swaps the body and pushes a URL without a document load, so gtag's automatic `page_view` fires only for the session's first page.
2. head-support keys head elements by `outerHTML` and *preserves* identical ones (deletes them from the merge map instead of re-appending), so the snippet is never re-evaluated on navigation either. No double counting, but no counting at all.

So the partial now sends `page_view` itself: `htmx:afterSettle` for forward navigation (the head merge runs on `afterSwap`, so `document.title` is already correct), and `htmx:historyRestore` for back/forward, which merges the head without firing a swap. A path guard keeps partial swaps that leave the URL alone from counting.

The Django admin stays excluded — that is in the context processor, untouched. The privacy policy already discloses Google Analytics and "pages viewed", so no policy change is needed; members still get a changelog entry since this widens what is measured.

## 3. Auto-announce a release on a VERSION bump

The workflow was manual-only because it used to re-post the **whole** release-line changelog on every hotfix merge. That reason is gone — the script now posts only entries stamped at the current `VERSION`.

Restores `push: branches: [main]`, keeps `workflow_dispatch`, and adds two guards: a `paths:` filter on `plfog/version.py`, and a gate step comparing `VERSION` against the parent commit.

The second guard is not redundant — `198b463` (PR #175) is a merge to main that did **not** bump `VERSION`; a naive push trigger would have re-announced 0.23.49 to members. Gate logic tested against that commit and against the 0.23.49→0.23.50 bump: skips the first, posts the second.

---

## Testing

New `tests/hub/google_analytics_spec.py` covers tag-absent-when-unconfigured, tag-present-when-configured, and the boosted-navigation hooks. Full local suite 6951 passed; `mypy plfog/ core/ membership/ hub/` clean under django-stubs 6.0.8; `ruff format`/`ruff check` clean.

## Two things for the reviewer

1. **Confirm `DISCORD_WEBHOOK_URL` points at #fog-app-announcements** before merging. That secret alone determines the destination, and this makes posting automatic.
2. `mypy>=1.13` and `django-stubs>=5.1` are unbounded, so any upstream release can turn CI red with no change on our side. A compatible-release bound on the type-checking tools would make CI reproducible. Not done here — that is a policy call.

https://claude.ai/code/session_01BjUajPuNWTaHF4ncHP9Qr4

### DIFF
--- a/.github/workflows/discord-notify.yml
+++ b/.github/workflows/discord-notify.yml
@@ -1,12 +1,24 @@
 name: Discord Notifications
 
-# Manual-only on purpose. This used to fire on every push to main, which re-posted the
-# WHOLE current release-line changelog on every routine hotfix merge — a duplicate
-# announcement to members each time. A release announcement is a deliberate act, so it is
-# now triggered by hand when a release actually ships:
-#   gh workflow run discord-notify.yml        (or the "Run workflow" button in the Actions UI)
+# Fires automatically when a push to main ships a new VERSION, and can still be run by
+# hand (`gh workflow run discord-notify.yml`, or the Actions "Run workflow" button).
+#
+# This was manual-only for a while because it used to re-post the WHOLE current
+# release-line changelog on every routine hotfix merge, spamming members with things
+# already announced. The script now posts ONLY the entry/entries stamped at the current
+# VERSION, so one merge means one deploy means one announcement. Two guards keep it that
+# way on automatic runs:
+#   1. `paths:` — the job is only considered when plfog/version.py itself changed.
+#   2. the "Check whether VERSION changed" step below — it compares VERSION against the
+#      parent commit and skips the post when it is unchanged, so edits to changelog
+#      wording (or any other version.py churn) never re-announce a shipped release.
+# A manual run always posts, which is the escape hatch for re-sending an announcement.
 on:
   workflow_dispatch:
+  push:
+    branches: [main]
+    paths:
+      - "plfog/version.py"
 
 jobs:
   notify:
@@ -14,13 +26,38 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v4
+        with:
+          # Need the parent commit so the step below can diff VERSION across the push.
+          fetch-depth: 2
 
       - name: Set up Python
         uses: actions/setup-python@v5
         with:
           python-version: "3.13"
 
+      - name: Check whether VERSION changed
+        id: version_changed
+        env:
+          EVENT_NAME: ${{ github.event_name }}
+        run: |
+          set -euo pipefail
+          if [ "$EVENT_NAME" = "workflow_dispatch" ]; then
+            echo "Manual run, announcing regardless of VERSION."
+            echo "should_post=true" >> "$GITHUB_OUTPUT"
+            exit 0
+          fi
+          current=$(python -c "import sys; sys.path.insert(0, '.'); from plfog.version import VERSION; print(VERSION)")
+          previous=$(git show HEAD^:plfog/version.py 2>/dev/null | grep -E '^VERSION = ' | head -1 | cut -d'"' -f2 || true)
+          echo "previous=${previous:-<none>} current=${current}"
+          if [ "$current" = "$previous" ]; then
+            echo "VERSION is unchanged, skipping the announcement."
+            echo "should_post=false" >> "$GITHUB_OUTPUT"
+          else
+            echo "should_post=true" >> "$GITHUB_OUTPUT"
+          fi
+
       - name: Post release announcement to Discord
+        if: steps.version_changed.outputs.should_post == 'true'
         env:
           DISCORD_WEBHOOK_URL: ${{ secrets.DISCORD_WEBHOOK_URL }}
         # Posts only the current VERSION's changelog entry (what just shipped),
--- a/hub/views.py
+++ b/hub/views.py
@@ -38,6 +38,7 @@ from hub.forms import (
     GuildEditForm,
     GuildRoleFormSet,
     MemberAdminEditForm,
+    MemberContactForm,
     MemberContactFormSet,
     MemberSkillForm,
     OrgInfoPageForm,
@@ -1448,7 +1449,10 @@ def user_settings(request: HttpRequest) -> HttpResponse:
     member = _get_member(request)
 
     profile_form: ProfileSettingsForm | None
-    contact_formset: BaseInlineFormSet | None
+    # Parameterized on purpose: django-stubs 6.0.8 types inlineformset_factory's product as
+    # BaseInlineFormSet[MemberContact, Member, MemberContactForm], and a bare BaseInlineFormSet
+    # means BaseInlineFormSet[Any, Any, ModelForm[Any]], which no longer accepts that assignment.
+    contact_formset: BaseInlineFormSet[MemberContact, Member, MemberContactForm] | None
     if request.method == "POST" and request.POST.get("form_id") == "profile":
         if member is None:
             messages.error(request, "Your account is not linked to a membership.")
--- a/plfog/version.py
+++ b/plfog/version.py
@@ -2,9 +2,22 @@
 
 from __future__ import annotations
 
-VERSION = "0.23.49"
+VERSION = "0.23.50"
 
 CHANGELOG: list[dict[str, str | list[str]]] = [
+    {
+        "version": "0.23.50",
+        "date": "2026-08-06",
+        "title": "Usage measurement now covers the Member Hub",
+        "changes": [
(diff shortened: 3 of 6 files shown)


PastLivesReviewBot (approved):

Independent verification of the three changes, plus the claims in the description.

**1. mypy annotation (`hub/views.py:1455`)**

Checked against the installed stubs rather than the description. `django-stubs` 6.0.8 declares `class BaseInlineFormSet(BaseModelFormSet[_M, _ModelFormT], Generic[_M, _ParentM, _ModelFormT])` and `inlineformset_factory(...) -> type[BaseInlineFormSet[_M, _ParentM, _ModelFormT]]`. With `inlineformset_factory(Member, MemberContact, form=MemberContactForm, ...)` at `hub/forms.py:417`, the produced type is exactly `BaseInlineFormSet[MemberContact, Member, MemberContactForm]`. The annotation matches, parameter order included.

It cannot mask a bug: the new annotation is strictly narrower than the bare `BaseInlineFormSet` it replaces (which erased to `Any, Any, ModelForm[Any]`), so it can only catch more. It is a function local variable annotation, never evaluated at runtime, so there is no runtime effect. CI lint is green on this head.

**2. Google Analytics on the hub**

Verified the mechanism in the vendored sources instead of trusting the reasoning.

`static/js/htmx.min.js` is htmx 2.0.4. In the ajax response handler the history update runs *before* the swap: `htmx:beforeHistoryUpdate`, then `pushUrlIntoHistory`, then `htmx:pushedIntoHistory`, and only then the swap call carrying `afterSwapCallback` / `afterSettleCallback`. So by the time `htmx:afterSettle` fires, `window.location` already holds the new URL and the path guard compares correctly. No off by one on `page_location`.

Both history restore paths (cache hit, and the cache miss XHR) use `swapInnerHTML` plus `settleImmediately` directly and never route through the swap function, so neither emits `htmx:afterSettle`. `historyRestore` therefore cannot double fire with the `afterSettle` handler, and `document.title` is assigned before the event is triggered, so the title is correct in both handlers. The shared `lastPath` closure variable is a second layer of protection.

`static/js/htmx-ext-head-support.js` keys head children by `outerHTML` and, on a match, deletes the entry from the merge map so the existing node is preserved rather than re-appended. Both GA elements render identically on every page, so the block is never re-evaluated on a boosted navigation. That rules out accumulating listeners and a repeated `gtag('config')` (which would itself send a page_view).

Checked the one remaining double count vector: a boosted navigation to a layout whose head lacks the include would remove the GA elements, and returning would re-append and re-execute them. Not reachable. `htmx.min.js` is loaded only by `templates/hub/base.html` and `templates/signage/base.html`, no hub template links to the Django admin or to signage, and the sidebar nav carries `hx-boost="false"`.

On the other three templates that carry the include, the listeners register and never fire, because those layouts load no htmx at all. In all three the include also sits in `<head>` (`templates/base.html:17` directly, classes and guilds via `{% block extra_head %}`), not in a body region a boost swap could re-execute. Inert, not harmful.

No `hx-push-url` anywhere under `templates/hub/`, so partial swaps cannot inflate the count, and the path guard covers same URL swaps. The project sets no Content Security Policy, so nothing blocks `googletagmanager.com` on the new surface. The admin exclusion is untouched and still lives at `core/context_processors.py:103`. The GA field is `blank=True` with no default, so the block is absent by default and existing suites are unaffected.

**3. `discord-notify.yml` gate**

`plfog/__init__.py` is empty and `plfog/version.py` imports only `__future__`, so the gate's `python -c` import succeeds with no dependency install step. `.github/scripts/discord_release_notify.py` is standard library only (`json`, `os`, `sys`, `urllib`), so the post step is safe to run automatically too.

Shell reviewed against `set -euo pipefail`. The `|| true` sits inside the command substitution and covers the whole pipeline, so a missing parent blob, a `grep` miss, and `head -1` closing the pipe early all yield an empty `previous` rather than a failed step. Empty `previous` compares unequal and posts, which is the safe direction. `set -u` is satisfied because `previous` is always assigned, and `${previous:-<none>}` is quoted.

Ran the gate logic against real commits. `198b463` (a real merge to main with no bump) resolves `previous=0.23.49 current=0.23.49` and skips. A squash merge of this branch resolves `previous=0.23.49 current=0.23.50` and posts. Main's recent history is linear single parent with `(#NNN)` titles, so squash merge is the convention here and `HEAD^` is the previous main tip. First commit and merge commit cases are both handled: no parent yields an empty `previous` and posts, and on a merge commit `HEAD^` is the first parent, which is still the previous main tip.

**4. Changelog**

`VERSION` is `0.23.50` and the top entry is stamped `0.23.50`, so `announce_release` resolves. The only prior Google Analytics entry is at `0.10.0`, long since live, so per CLAUDE.md this is a fix to shipped behaviour and correctly gets its own entry rather than folding into an existing entry in the current line. Copy is plain and member facing with no jargon, PR numbers, or hashes. The wording is accurate against `templates/core/privacy_policy.html:37` and `:57`, which already disclose Google Analytics and "pages viewed".

**5. Tests and checks**

`tests/hub/google_analytics_spec.py` passes 3 of 3 locally. Placement and style match the neighbouring specs in `tests/hub/`, and the missing `-> None` on the `it_` functions matches the repo convention exactly (0 of 1662 `it_` functions annotate a return type). `ruff check` and `ruff format --check` are both clean. CI `lint` and `test` are green on `0db6631`. The `e2e` failure on this head is infrastructure, not code: the job was never acquired by a hosted runner and its conclusion is `cancelled`.

**Non blocking notes**

1. A rebase merge, or a direct multi commit push to main, would silently suppress the announcement. The bump lives in `1b31abd`, not in the head commit, so evaluating the gate with `HEAD=0db6631` yields `previous=0.23.50 current=0.23.50` and skips. Squash merge makes this correct for this PR. The failure direction is a missed announcement rather than a duplicate, and `workflow_dispatch` recovers it, so this is a footgun rather than a defect.
2. The description offers `198b463` as proof the gate step is not redundant, but that commit does not touch `plfog/version.py` at all, so the `paths:` filter already excludes it. The gate step's real justification is the one stated in the workflow comment: a changelog wording edit to `version.py` with no `VERSION` bump.
3. The announcement now races the deploy. Posting on push to main can tell members about a release before Render finishes deploying it; the manual trigger previously let that be sequenced by hand.
4. `sys.path.insert(0, '.')` in the gate's `python -c` is redundant, since the current directory is already `sys.path[0]` for `-c`. Harmless.
5. Roughly 800 bytes of JS comments now ship in the head of every member facing page. Consistent with the existing inline scripts in `hub/base.html`, so this reads as house style.
6. `gtag('event', 'page_view', ...)` does not pass `page_referrer`. This is the documented GA4 single page app pattern and GA infers the referrer per session, so in app navigation attribution is slightly less precise. Worth knowing when reading the reports.

**Before merging**

The description's first reviewer request stands and cannot be checked from here: confirm `DISCORD_WEBHOOK_URL` points at #fog-app-announcements, since that secret alone decides the destination and this change makes posting automatic.

PR #183, opened 2026-08-07
[analytics] Track the Django admin too, and fix double-counting on class/guild pages (v0.23.51)

HexagonStorms:
## Why

The ask was to track **everything** in FOG. The admin was excluded, and #182 left that exclusion in place — my miss: I inherited the existing behavior instead of the instruction, and even wrote "the admin stays excluded" into that PR as though it were settled.

## Removing the exclusion took five changes, not one

The path check alone would have accomplished nothing, because the admin templates never included the partial:

| File | Change |
|---|---|
| `core/context_processors.py` | dropped the `/admin/` path check |
| `templates/admin/base.html` | include the partial via Unfold's `extrahead` block |
| `core/models.py` | help text said "excludes the Django admin" |
| `core/admin.py` | fieldset description said the same |
| `tests/core/context_processors_spec.py` | inverted the exclusion spec |

## It also fixes a double-count I introduced in #182

`templates/classes/base_public.html` and `templates/guilds/base_public.html` both `{% extends "hub/base.html" %}` **and** included the GA partial in `{% block extra_head %}`. Once #182 put the include in the parent's `<head>`, those pages rendered it twice. Measured on a real render of `/classes/`, not inferred:

```
googletagmanager.com/gtag/js tags:  2
gtag('config' calls:                2   → two page_views per load
live page_view senders:             2   → two more per hx-boost navigation
```

So class and guild traffic would have read roughly double. This is live on `main` right now and harmless only because the measurement ID has never been set — had the ID gone in first, the data would have been wrong from the start. The two redundant includes are removed, with comments explaining why.

Worth noting the review on #182 specifically examined this class of bug and cleared it: it correctly verified that head-support preserves identical head nodes rather than re-appending them, but did not catch that these two templates contribute a *second* copy. The clearance was narrower than it read.

## Regression cover

`tests/core/google_analytics_coverage_spec.py` asserts exactly one snippet on the public home, the public class list, the member hub, and the admin, plus exactly one `page_view` sender. One assertion needed correcting while writing it: counting `htmx:afterSettle` caught a pre-existing, unrelated Alpine listener at `hub/base.html:554`, so it counts the GA sender instead.

## Changelog

0.23.51 plainly corrects the "the staff admin area is still excluded" line from 0.23.50, since that already went out to Discord. The 0.23.50 entry is left as-is — it was accurate when it shipped.

## Testing

11 GA specs pass. Full suite green (6954 passed) on the admin changes; ruff and mypy clean.

## One thing to decide

Admin changelist search puts the query in the URL (`?q=member@example.com`) and GA records `page_location`, so searching a member by email sends that email to Google — which their terms prohibit. Tracking every admin page does not require sending query strings; stripping them on admin paths is a couple of lines. Not done here, since the ask was to track everything and this is a policy call. Say so and I will add it.

https://claude.ai/code/session_01BjUajPuNWTaHF4ncHP9Qr4


PastLivesReviewBot (approved):
## Approved — re-review at `4367032`

Both blocking findings from the review at `d26cb65` are resolved, and the coverage gaps flagged as non-blocking are closed. Verified independently rather than taken on trust.

### Blocking findings — both fixed

**1. Missing migration.** `core/migrations/0058_alter_siteconfiguration_google_analytics_measurement_id.py` now carries the `help_text` edit, and its field definition matches `core/models.py:262` exactly. `makemigrations --check --dry-run` reports no changes. Applied forward, reversed to `0057`, and re-applied on a scratch database — clean in both directions. Pure `AlterField`, no `RunPython`, so reversibility is structural.

**2. `?email=` reaching Google.** Fixed and confirmed by executing `plCleanUrl` in a real JS engine against 25 edge cases:

| Case | Result |
|---|---|
| `/accounts/login/code/?email=<member>` | query dropped |
| `/accounts/signup/?email=<invitee>` | query dropped |
| `?q=secret&utm_source=discord&utm_medium=post` | `q` gone, both `utm_*` preserved |
| `?email=…&utm_source=email` | `email` gone, `utm_source` preserved |
| `/admin/…?status__exact=active&e=1` | whole query dropped |
| `/classes/admin/registrations/?q=…&utm_source=x` | correctly *not* admin-branched; only `q` stripped |
| repeated `?q=a&q=b` | both removed |
| `#fragment` | preserved |
| relative URL, `''`, `null`, `undefined` | handled, no throw |
| unparseable input | fails closed via `String(href).split('?')[0]` |
| cross-origin referrer `google.com/search?q=…` | stripped |

`stripped ? u.href : href` returning the original when nothing changed is a good touch — it avoids `URL` normalisation quietly rewriting reported URLs.

I confirmed `q` and `email` are the complete set: every `request.GET` key read across the codebase is otherwise an ID, page number, filter enum, or sort key. Member URLs use integer PKs throughout, so no personal data rides in a path.

### Snippet coverage — crawled independently

Rendered every argument-free URL as staff, member, and anonymous (307 URLs, 921 combinations), then the surfaces a plain crawl cannot reach:

- **Duplicates: zero.** Nothing anywhere emits more than one `gtag('config'` or one `gtag/js` tag. The `classes/base_public.html` and `guilds/base_public.html` double-count is gone, confirmed on the real `book.` and `guilds.` hosts, not just the members surface.
- **Missing: zero real pages.** The only 200s without a snippet are htmx fragments — `/calendar/events/`, `/notifications/unread-count/`, `/admin/search/`, and `signage/_deck.html`. Those are swap targets, not pageviews; excluding them is correct.
- All 12 reachable allauth pages: exactly one each, at every auth state.
- Signage player, both flyers, and `orientation_action` (including its 400 invalid-token branch): exactly one each.
- Blank measurement ID removes the snippet everywhere, admin included.

### The allauth fork

`diff` against the packaged `allauth 65.18.0` layout shows exactly two additions: the explanatory comment block and the include. Every block (`head_title`, `extra_head`, `body`, `content`, `extra_body`) and the `{% load i18n %}` are intact, so `allauth/layouts/entrance.html` and `manage.html` and their children render unchanged. These pages already used allauth's bare default layout, so nothing regresses visually. `TEMPLATES.DIRS` precedes `APP_DIRS`, so the override resolves. The upgrade note in the header is the right mitigation, and `it_tags_allauth_pages_we_do_not_override` is what actually catches a silent drop.

### Changelog

Correct under the `CLAUDE.md` rules. `0.23.50` shipped to Discord on 2026-08-07 (run `31133916495`) telling members "the staff admin area is still excluded" — so this is a correction to something already announced, which earns its own entry rather than an edit to the shipped one. Exactly one entry at `VERSION` (`0.23.51`), so `announce_release` resolves. Plain language, no jargon or PR numbers, and it openly corrects the earlier claim. The Privacy Policy it points at (`templates/core/privacy_policy.html:37,57`) is generic about analytics and is not contradicted by the change.

### Notes, none blocking

- **Signage is well-behaved.** Slides advance via `setTimeout` with no URL change, so they fire no pageviews, and the deck poll returns an untracked fragment. The only reload is `scheduleNightlyReload()` at 04:00 (`templates/signage/player.html:98-104`), so a kiosk contributes roughly one pageview a day. Tracking it was an explicit request; the one thing to expect is that a kiosk reads as a long-idle user and will drag property-wide engagement averages down a little. A GA4 filter later would fix that if it bothers anyone.
- `/discord/link/callback/?code=<oauth code>` sends the authorization code to GA as `page_location`. Pre-existing — `hub/base.html` already carried the snippet on `main` — and the code is single-use and already consumed server-side by render time. Worth a follow-up, not a blocker here.
- `/admin` without a trailing slash misses the prefix test, but Django's `APPEND_SLASH` redirects before a page renders, so no real URL reaches GA that way.
- Cosmetic: the include in `templates/signage/base.html:17` sits unindented against a 4-space head block.
- `mypy` reports 39 errors, all pre-existing in `classes/` and `billing/` and outside the `plfog/ core/ membership/ hub/` scope CI checks. None in any file this PR touches.

### Standards and CI

Fat-model boundaries respected; the context processor is now simpler than before. Full annotations on changed functions. Specs are BDD `describe_*`/`it_*` with no `context_*` blocks, no `skip`, no `pragma`. `ruff check`, `ruff format --check`, and the changed specs (28 passed) all green locally. All three CI checks pass on `4367032` — `lint`, `test`, and `e2e` — verified as running against that exact SHA.

PR #184, opened 2026-08-07
[mailchimp] Marketing opt-in on account signup, env credentials, and an opt-in stamp fix (v0.23.52)

HexagonStorms:
## What this does

Adds a marketing opt-in checkbox to account signup and wires it to Mailchimp, plus the config and correctness fixes that came out of auditing the existing integration.

Before this, the only place to opt into marketing email was the class booking form. Anyone who created a FOG account without booking a class had no way to subscribe. `ACCOUNT_SIGNUP_FIELDS` was `["email*"]` with no custom signup form, and `plfog/adapters.py` had no Mailchimp call.

## Changes

**1. Marketing opt-in on account signup (the main thing)**

* `plfog.adapters.MarketingOptInSignupForm` adds an optional, unchecked-by-default `wants_newsletter` checkbox. Registered as `ACCOUNT_FORMS["signup"]`.
* New `core/services/mailchimp_account.py` with `subscribe_user(user)` and `derive_account_tags(user)`. This is the sibling of `classes/services/mailchimp_subscribe.py`; the module was specified in `docs/plans/gap-closure.md` and referenced by `UserProfile.subscribed_to_mailchimp_at`'s help text, but had never been built.
* Tags applied: `newsletter` (the shared cross-path marketing tag, same one `/newsletter/` uses) plus `account-signup` to record which door they came through.
* On success it stamps `UserProfile.subscribed_to_mailchimp_at`, which is what suppresses the checkbox on later class registrations.
* Checkbox renders on both the members surface and the guest/book surface, with a small CSS rule per stylesheet.
* Best effort throughout: a disabled integration, a rejected push, or a client that raises all leave the account intact. The push runs after the user is committed.

**2. Credentials can now come from the environment**

`MailchimpClient.from_site_config()` resolves each credential independently: the `SiteConfiguration` value wins when set, otherwise it falls back to `MAILCHIMP_API_KEY` / `MAILCHIMP_LIST_ID`. Existing behaviour is unchanged for anyone who already pasted credentials into Site Settings.

This exists so production can be enabled without an account-scoped Mailchimp key sitting in the database in plaintext, readable by any superuser. Both vars are now declared in `render.yaml` and `.env.example`, neither of which mentioned Mailchimp at all before.

**3. Fixed: first-time guests were never marked as opted in**

`_stamp_profile_subscribed()` runs inside `subscribe_registration()`, which is called before `ensure_account_for_registration()`. For a brand new guest who ticked both the newsletter box and "create an account", `registration.member` is still `None` at that moment, so the stamp was silently skipped and they got asked about the newsletter again on their next booking.

The fix puts the stamp in `_seed_profile_from_registration()`, where the profile is first created. Note the call order is deliberately left alone: `subscribe_registration` has to run first, because `derive_tags` decides `first-time-student` by asking whether the email is already a known member, and `ensure_account_for_registration` is precisely what makes it one. Swapping the two calls would have quietly dropped that tag for every guest who also created an account.

## Testing

* New: `tests/core/services/mailchimp_account_spec.py` (8), `tests/auth/signup_marketing_optin_spec.py` (9, covering the form, the end to end POST, and both rendered surfaces).
* Extended: `tests/core/integrations/mailchimp_spec.py` for env fallback and precedence, `core/spec/services/guest_account_spec.py` for the stamp fix.
* 100% line and branch coverage on `core/services/mailchimp_account.py`, `core/integrations/mailchimp.py`, and `core/services/guest_account.py`. The new form's lines in `plfog/adapters.py` are fully covered including the exception branch.
* `ruff format` and `ruff check` clean. `mypy` could not run: the django-stubs plugin fails to construct in this environment on Python 3.14, which reproduces on an untouched file and predates this branch.

## To turn it on

Set `MAILCHIMP_API_KEY` and `MAILCHIMP_LIST_ID` in the Render dashboard, or paste them at `/manage/site-settings/`. The audience id is `cfa0e74194`. Until one of those is set the integration stays disabled and every subscribe path is a silent no op, exactly as today.

Worth knowing: no subscriber has ever reached the audience through the API. The `class-registrant` and `newsletter` tags do not exist in it, and recent signups are all sourced from the Squarespace popup form. The class registration path has been correct in code for a while but has never been switched on.

## Not included

SimplyBook, deliberately. The key we have is rejected by the JSON RPC API that `core/integrations/simplybook.py` uses, so that one is blocked on getting the right key out of Custom Features then API then Settings.


PastLivesReviewBot (approved):
Approved by PastLivesReviewBot. Both findings from the earlier rounds are resolved, and I verified the fix independently rather than reading the specs.

## The consent regression is gone, end to end

I re-ran my own probe against a3bff3e, driving `RegistrationForm` with exactly the kwargs `classes/views.py:585` uses and then calling `subscribe_registration` on the saved row:

```
PROBE-A box shown to logged-out Bob: True
PROBE-A auto-linked member_id: 2
PROBE-A wants_newsletter after untick: False
PROBE-A subscribe called: False   (want False)

PROBE-B box shown to logged-in Bob: False
PROBE-B wants_newsletter recorded: True   (want True)
PROBE-B subscribe called: True    (want True)
PROBE-B tags: ['class-registrant', 'category-*', 'instructor-*']
```

Note PROBE-A still shows `member_id: 2` — `link_member_by_email` fires exactly as before, so the conditions that produced the old bug are all still present. It is the recorded flag, not the absence of the link, that makes the untick stick. That is the right reason for it to pass.

## Answers to the things you asked me to try to break

**Is `_newsletter_opt_in_suppressed` always initialised?** Yes. It is set unconditionally in `__init__` immediately after `super().__init__()` and the model-release pops, before anything can read it. `__init__` has no early returns, `user` defaults to `None` so `_user_already_opted_in(None)` short-circuits to False, `RegistrationForm` has no subclasses, and `classes/views.py:585` is its only construction site. Probe D built the form with no `user=` kwarg at all: attribute present, `wants_newsletter` False.

**Does `wants_newsletter=True` break another reader?** No. The only consumers are `classes/services/mailchimp_subscribe.py:107`, `templates/classes/admin/registration_detail.html:61`, and `demo_data.py` (which sets it explicitly). Both CSV exports in `classes/exports.py` use fixed column lists — `CSV_HEADERS` and `REGISTRATIONS_CSV_HEADERS` — and neither includes the field. The admin detail page will now show the badge for a suppressed opt-in, which is more honest than before and is precisely what the new `help_text` documents.

**The waitlist path.** The assignment sits above the `if self.is_waitlist:` branch, so it is not skipped. Probe C confirms: status `waitlisted`, `wants_newsletter` True. `promote_next_from_waitlist` never calls `subscribe_registration`, so nothing is pushed on a waitlist join and no behaviour changed there — the row is just truthful if that person is later promoted.

**The migration.** `0051` is the only one generated, depends on the existing `0050_update_default_faq_copy`, and is a plain `AlterField` with no `RunPython`, so the reverse-function rule does not apply. I round-tripped it on a scratch sqlite: `migrate classes 0050` unapplied it cleanly and `migrate classes` re-applied it. `makemigrations --check --dry-run` reports no changes.

**Specs actually run.** I checked by name rather than by count, since a silently-uncollected block is exactly the failure mode worth fearing here. All three appear under `describe_newsletter_opt_in_recorded_on_save` and PASS. No `context_*` blocks anywhere in the new specs.

## Nothing earlier was disturbed

`git diff main -- classes/services/mailchimp_subscribe.py` is now empty — the service is byte-identical to main again. Everything I verified in the first pass therefore stands unchanged: the `derive_tags` / `_is_known_member` call-order reasoning, `_stamp_profile_subscribed`, the guest stamp fix, credential precedence, the opt-in default, account survival on every Mailchimp outcome, and the changelog. Deleting `_already_opted_in` was the right call over patching around it; the form is where the decision is known, so that is where it belongs.

`ruff check` and `ruff format --check` clean. 683 tests pass across the form, service, view, auth, and integration specs. `core/integrations/mailchimp.py`, `core/services/mailchimp_account.py`, and `core/services/guest_account.py` remain 100% line and branch; both arms of the new `if self._newsletter_opt_in_suppressed:` are covered.

One housekeeping note, not a review objection: `lint` and `e2e` are green on this head but the `test` job was still running when I finished, so let it go green before merging.

PR #186, opened 2026-08-09
[voting] Automatic results + @everyone Discord notifications (v0.23.53)

HexagonStorms:
## Summary

- **Results email now sends automatically.** When `take_cycle_snapshot` runs at cycle close, it calls `send_results()` immediately after. No admin needs to click "Send results."
- **Every member with an account gets an email.** Phase 1: voters get a personalized ballot recap. Phase 2: non-voters get the allocation summary without a ballot section. Phase 3: Discord @everyone broadcast (gated; starts end of August).
- **Pre-close reminder email widened.** `vote_soon_occurrences()` now targets all active members with accounts, not just voters. Non-voters get a "go vote" nudge; voters still see their current picks.
- **Discord @everyone notifications.** Two new broadcast event types (`voting.discord_reminder`, `voting.results_discord`) fire once per cycle. The scheduler emits them as broadcast occurrences, not per-member — one webhook call total.
- **`--no-discord` flag on `send_funding_results`.** For the July results send (which predates the Discord rollout), pass `--no-discord` to suppress the @everyone post.

## July results

To send July results without Discord (do this on production after merge/deploy):

```
python manage.py send_funding_results --snapshot-id <ID> --no-discord
```

Find the snapshot ID in the admin at `/manage/voting/history/` or via `python manage.py shell -c "from membership.models import FundingSnapshot; print(FundingSnapshot.objects.last().pk)"`.

## Test plan

- [ ] `pytest tests/membership/voting_reminders_spec.py` — Discord reminder occurrence tests (6 new tests)
- [ ] `pytest tests/membership/funding_snapshot_results_spec.py` — Results send audience + URL tests pass
- [ ] `pytest tests/hub/voting_results_send_spec.py` — View send tests pass (admin client fixture isolated)
- [ ] `pytest tests/core/management/take_cycle_snapshot_spec.py` — Auto-send assertions pass
- [ ] Full suite passes (pre-push hooks: ruff, mypy, template checks all green)

PR #187, opened 2026-08-09
fix(tests): register Discord-only voting events in registry_spec

HexagonStorms:
Two CI failures from the new `voting.discord_reminder` / `voting.results_discord` events added in #186.

- `_BRAND_NEW_KEYS` hardcoded 27 keys — added the two new ones so the count assertion passes.
- In-app invariant test assumed every per-recipient event has an `IN_APP` channel. Discord-broadcast events are channel-only (no bell), so they're excluded alongside the existing email-only set.

### DIFF
--- a/tests/core/events/registry_spec.py
+++ b/tests/core/events/registry_spec.py
@@ -20,6 +20,8 @@ _BRAND_NEW_KEYS = {
     "voting.officers_closing_soon",
     "voting.results_published",
     "voting.results_ready",
+    "voting.discord_reminder",
+    "voting.results_discord",
     "release.published",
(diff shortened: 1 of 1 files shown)


PastLivesReviewBot (approved):
Reviewed and approved by PastLivesReviewBot. Test-only fix: the two new Discord-only event keys (voting.discord_reminder, voting.results_discord) are correctly added to _BRAND_NEW_KEYS so the count assertion holds, and excluded from the in-app invariant since Discord-broadcast events carry no in-app bell channel. CI green (e2e, lint, test). No issues.

PR #189, opened 2026-08-09
[billing] Stripe test/live mode toggle with dual key storage (v0.23.55)

HexagonStorms:
## What

Adds a **Testing Mode switch** to the admin Payments dashboard (Stripe tab) that flips every Stripe operation (charges, SetupIntents, webhooks) between a stored **TEST** key set and a stored **LIVE** key set. Both sets are saved at once, so going live (or dropping back to testing) is a single toggle with no re-pasting of keys.

Prep for the FOG payments go-live.

## How

- **`BillingSettings` model** gains a `test_mode` boolean plus a parallel `test_connect_*` credential slot (client id, publishable key, encrypted secret key, encrypted webhook secret). The existing `connect_*` fields become the LIVE slot (no lossy rename).
- **Fat-model accessors** `active_client_id` / `active_publishable_key` / `active_secret_key` / `active_webhook_secret` (and `mode_label`) pick the slot from `test_mode`. Every read site (`stripe_utils._platform_secret_key`, `_platform_webhook_secret`, `setup_payment_method`'s publishable key) now goes through these, so no caller branches on mode.
- **`clean()`** requires only the *active* slot's four fields when Connect is enabled, so the inactive slot can be left blank or pre-filled ahead of a flip.
- **Data migration (`0012`, reversible)** inspects the existing secret: an `sk_test_` key is relocated into the new TEST slot with `test_mode=True` (so current testing keeps working and the LIVE slot is left empty for go-live); an `sk_live_` key stays put with `test_mode=False`; an empty config defaults to test mode.
- **UI** (Payments dashboard Stripe tab, rebuilt to FRONTEND.md components): the toggle, a clear TEST/LIVE active badge, two labelled key groups with a per-slot "Test connection" button, and a confirm prompt when switching to live.

## Notes for review

- **Admin-only change.** `VERSION` bumped to `0.23.55` with a plain-language, admin-scoped `CHANGELOG` entry (kept because the `announce_release` invariant and its tests require an entry at the current `VERSION`; the `0.23.54` "Admin tools" entry is the precedent for admin-facing notes). It does not have to be announced to members.
- **Tests:** full suite green, coverage above the 98% gate. New specs cover the accessors (both branches), `clean()` branches, the data migration forward and reverse, form validation, `stripe_utils` slot selection, and the dashboard render/save.
- **Pre-existing baseline, not introduced here:** mypy has unrelated errors on `main` (e.g. `classes/*`, `billing/views.py:130-135` in `stripe_webhook`); this branch adds zero new ones. Secrets stay in `EncryptedCharField`; `billing/admin.py` still excludes all credentials by design.

Branched from `main` and rebased onto its latest tip.

https://claude.ai/code/session_01MiwKUFJixRUa1ePtifRQVb


PastLivesReviewBot (approved):

PR #190, opened 2026-08-09
[voting] Visual bar chart in the guild-funding results email (v0.23.56)

HexagonStorms:
## What

The monthly guild-funding **results email** listed the allocation as a plain monospace block (`<pre>`). It now shows the same **bar chart** members watch on the voting page: one gold bar per guild (funding-descending), sized to that guild's funding relative to the leader, with the dollar amount, share of the pool, and a medal on the top three — rendered into the branded dark email shell.

This is the visual half of the "results go to every member" feature (the send/automation landed in #186); folded into the same changelog entry, re-stamped to v0.23.56.

## How

- **`FundingSnapshot.allocation_chart_html()`** — builds an email-safe, table-based bar chart (`SafeString`) from the snapshot results, via a new app-owned partial `templates/membership/emails/_allocation_chart.html`. Guild names are autoescaped by the template engine. `allocation_summary()` (plain text) is unchanged and still drives the text body.
- **`send_results()`** passes `allocation_chart` into both the voter and non-voter emit contexts.
- **`core.events.rendering.render_html`** now passes a `SafeString` context value through **unescaped** (the opt-in Django convention for trusted, app-built markup). Plain strings from DB copy are still escaped, so the copy injection guard is unchanged — only application code can mark a value safe.
- **`voting.results_published` copy** — HTML body uses `{{ allocation_chart }}` in place of `<pre>{{ allocation_summary }}</pre>`; added the `allocation_chart` placeholder and a two-bar sample so the live copy-editor preview renders the real chart.

## Tests

- `render_html`: SafeString value passes through unescaped; a plain string alongside it is still escaped (injection guard intact).
- `allocation_chart_html`: one escaped bar per guild, medal + 100%-width leader, dollar/percent figures, and empty markup for a voteless/legacy snapshot.
- Full events + funding/voting sweep green locally (847 passed).

## Notes

Table-based bars with inline styles only, so it survives Gmail / Apple Mail / Outlook. No migration. Rebased on latest `main` (past the Stripe billing-mode v0.23.55).

https://claude.ai/code/session_01NjbhB5hvRChHow2qcjDP8E

### DIFF
--- a/core/events/copy.py
+++ b/core/events/copy.py
@@ -27,6 +27,8 @@ from __future__ import annotations
 
 from dataclasses import dataclass, field
 
+from django.utils.safestring import mark_safe
+
 from core.events.registry import Channel, EventType, Recipients, all_events, get_event
 
 # Channels that carry authored copy. Discord reuses the in-app/email copy (it is a
@@ -721,7 +723,12 @@ _CURATED: dict[str, EventCopy] = {
             # automated send; used for a one-off, e.g. a late "sorry this is overdue").
             "intro_note",
             "cycle_label",
+            # allocation_summary: plain-text lines for the text body.
             "allocation_summary",
+            # allocation_chart: app-built, email-safe HTML bar chart for the HTML body
+            # (FundingSnapshot.allocation_chart_html) — a SafeString the renderer injects
+            # unescaped. The text body still uses allocation_summary.
+            "allocation_chart",
             # ballot_recap: pre-formatted sentence for voters ("You voted — 1st: X, 2nd: Y, 3rd: Z.")
             # Empty string for non-voters so the paragraph renders blank rather than showing
             # "1st: , 2nd: , 3rd: ." with missing values.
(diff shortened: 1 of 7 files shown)


PastLivesReviewBot (approved):
Reviewed and approved by PastLivesReviewBot. Security (the critical axis): the render_html SafeString-passthrough is sound. The isinstance(value, SafeString) gate passes through ONLY genuine SafeString instances, which only application code can mint via mark_safe; every DB-copy and user-derived context value (member_name, cycle_label, vote_1st/2nd/3rd, allocation_summary, ballot_recap) is a plain str and still hits escape(). JSON-loaded DB values deserialize as plain str, never SafeString, so the injection guard for admin/DB copy is unchanged. A codebase sweep confirms the only SafeString entering an event render_html context is the app-built allocation_chart; other mark_safe uses live in admin displays, templatetags, and the separate release-email path and do not feed event copy. Guild names inside the chart are HTML-escaped by Djangos autoescaping engine via the _allocation_chart.html partial, proven by it_html_escapes_guild_names (<Guild> becomes &lt;Guild&gt;, & becomes &amp;). The renderer tests assert BOTH the passthrough AND that a plain string alongside a SafeString is still escaped (<script> absent, &lt;script&gt; present). The SafeString also survives the synchronous in-process emit path (no context serialization), so the chart renders correctly. Standards: chart logic lives on the FundingSnapshot model (fat-model, thin-view), fully type-annotated (-> SafeString), no new fields; email HTML is table-based with inline styles and a solid-color fallback under the gradient (Outlook-safe); BDD specs use describe_/it_ with no context_ blocks and no skip/pragma markers; changelog folds the bar-chart refinement into the existing results-email feature entry re-stamped to 0.23.56 == VERSION. CI note: test, lint, and e2e were all still pending at review time; coordinator should confirm green before merge. Verdict: pass.

PR #191, opened 2026-08-09
[discord] Harden nightly events push and role sync against Discord throttling (v0.23.57)

HexagonStorms:
## What and why

Two separate Discord sync failures were firing in production. Both are fixed here, in one PR, because they share the same root shape (a Discord HTTP status the code did not handle robustly) and the same test surface.

### Bug A: nightly `sync_all_sources` failed with HTTP 429

The failing "source" in the nightly job was the Discord Scheduled Events mirror, not the Drupal class feed (that endpoint was independently confirmed healthy). Prod log:

```
discord events: 5 Discord event push(es) failed; first: 'Art Framing open studio hours':
  Discord API 429: {"message": "You are being rate limited.", "retry_after": 1.149, ...}
```

`sync_discord_feed_events` pushes up to 50 scheduled events back to back, and `DiscordScheduledEventsClient._execute` retried a 429 exactly once. Under a burst, several calls still 429 after their single retry and were recorded as failures.

**Fix:** `_execute` now retries a 429 up to `_RATE_LIMIT_MAX_ATTEMPTS` (3) times, each wait still bounded by `_RATE_LIMIT_MAX_WAIT_SECONDS`; and `sync_discord_feed_events` lightly paces its calls (`_DISCORD_PUSH_PACE_SECONDS = 0.5`, shared across the push and delete passes so the first call is un-paced and every later one waits). The interactive single-event save path does not go through this loop and still fails fast into the retry cron.

### Bug B: `sync_discord_guild_roles` failed 403 every 15 minutes

The uncaught error was the unlinked-reactor link nudge. `send_dm_text` treated a 403 as "DMs closed → return False" only when the body carried Discord error code `50007`. Any other 403 on the DM message POST hit `raise_for_status()` and crashed `reconcile_reactions`. Because the `DiscordLinkNudge` guard row is written only after a clean or closed send, the same reactor was retried and crashed every single tick, forever.

**Fix:** any per-recipient 403 on the DM open/send now counts as undeliverable (`_dm_undeliverable`), so the caller marks the reactor nudged and never retries. A bot-wide auth problem is a 401, not a per-user 403, so `401 / 429 / 5xx / network` still raise loudly and surface a genuine outage. The actual Discord sub-code is logged.

## Tests

New and updated BDD specs across `discord_events_spec`, `calendar_service_spec`, `discord_dm_spec`, and `discord_sync_spec` cover: multi-attempt 429 retry (exact call and sleep counts), pacing between two pushes and between a push and a later delete in the same run, any-403-returns-False on both DM calls (including a non-50007 message-post 403), and a 401 still raising. Sleeps are patched so tests stay fast.

Local run of the full CI gate: `ruff format --check` + `ruff check` clean, `mypy plfog/ core/ membership/ hub/` clean, `manage.py check` clean, full `pytest` = 7070 passed, total coverage 98.36% (>= 98% threshold).

## Out of scope (follow-up, not in this PR)

The architect review flagged that `retry_discord_event_pushes` and `backfill_discord_events` call `push_community_event` without `retry_on_rate_limit=True` at all, so that path gets zero 429 retries even after this change. It is unrelated to the two reported bugs and left for a separate ticket.

## Version

`VERSION` bumps to `0.23.57` (main is 0.23.55; PR #190 owns 0.23.56) with one plain-language CHANGELOG entry covering both fixes.

### DIFF
--- a/core/events/discord_dm.py
+++ b/core/events/discord_dm.py
@@ -129,23 +129,31 @@ def post_dm(discord_user_id: str, message: Message) -> bool:
     return False
 
 
-_CLOSED_DM_ERROR_CODE = 50007  # Discord's "Cannot send messages to this user".
+def _dm_undeliverable(response: httpx.Response) -> bool:
+    """``True`` when Discord refused a DM open/send with 403 — undeliverable to this recipient.
 
-
-def _dms_closed(response: httpx.Response) -> bool:
-    """``True`` when Discord refused with 403 + JSON error code 50007.
-
-    That combination means the recipient's privacy settings block DMs from the bot —
-    permanently undeliverable, never worth a retry. Anything else (other statuses,
-    non-JSON bodies, other error codes) is NOT "closed DMs" and stays loud.
+    A 403 means closed DMs, a blocked bot, no mutual server, or any other per-user block, and
+    is never worth retrying. Broadened from the original 403 + code-50007 check after a 403
+    with a different sub-code on the message-post call crashed the reconcile cron every 15
+    minutes on the same reactor forever (2026-08-09): the guard row is only written on a clean
+    or closed send, so an uncaught raise never wrote it and the same reactor was retried and
+    crashed every tick. A non-403 failure (401 bad token, 429, 5xx, network) is systemic, not
+    per-user, and must stay loud.
     """
-    if response.status_code != 403:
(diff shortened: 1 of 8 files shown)


PastLivesReviewBot (approved):

PR #192, opened 2026-08-10
fix(css): global responsive image reset so photos fit mobile viewports (v0.23.59)

HexagonStorms:
## Problem
Following the link in the Discord "You're halfway there" guild-signup reminder on a phone, images render oversized and spill off the screen. Root cause is systemic, not one page: **there is no global responsive-image reset** in the primary stylesheets. Every image is safe only because someone remembered to scope a class rule to it. Any `` without a scoped size rule overflows a narrow viewport, most visibly the guild logo SVGs, which ship with only a `viewBox` (no intrinsic width/height) and fall back to the browser's 300x150 replaced-element default.

## Fix
Add one global rule to the three top-level entry stylesheets so coverage is site-wide (public/auth via `style.css`, member hub via `hub.css`, public CMS/booking via `cms-public.css`):

```css
img {
    max-width: 100%;
    height: auto;
}
```

Every layered stylesheet (book-account, calendar, cms-guilds, etc.) loads on top of one of these three, so it inherits the reset automatically. No template changes needed.

### Scope decisions
- Selector is `img` only. `svg`/`video` deliberately excluded: ~50 inline `<svg>` icons are sized via HTML attributes in tight flex layouts (regression risk for no benefit), and no `<video>` tags exist.
- `height: auto` verified safe: every existing sized `` uses a class-qualified selector (specificity 0,1,x) that beats bare `img` (0,0,1) regardless of load order, so heroes, galleries, avatars, and logos are unchanged.
- Untouched on purpose: emails (inline-styled), `signage.css` (kiosk), flyer CSS (print/export), admin/unfold CSS.

## Tests
- New `tests/css_responsive_images_spec.py` guards the reset from being silently dropped from any entry stylesheet (mirrors the existing `template_comment_lint_spec.py` idiom).
- Full suite green locally: 7071 passed, 98.35% coverage.

## Release
- `VERSION` -> `0.23.58`
- Member-facing changelog entry added (fix to production-live behavior, so it earns its own entry).

### DIFF
--- a/plfog/version.py
+++ b/plfog/version.py
@@ -2,9 +2,19 @@
 
 from __future__ import annotations
 
-VERSION = "0.23.58"
+VERSION = "0.23.59"
 
 CHANGELOG: list[dict[str, str | list[str]]] = [
+    {
+        "version": "0.23.59",
+        "date": "2026-08-09",
+        "title": "Photos now fit your screen, especially on phones",
+        "changes": [
+            "Some photos — guild icons and logos especially — could show up way too big on a phone and "
+            "spill off the edge of the screen. Every photo across the site now automatically resizes to "
+            "fit, so pages look right no matter what device you're using.",
+        ],
+    },
     {
         "version": "0.23.58",
         "date": "2026-08-09",
--- a/static/css/cms-public.css
+++ b/static/css/cms-public.css
@@ -6,6 +6,19 @@
    for dark mode via [data-theme="dark"] .cp-page.
 ==================================================================== */
 
+/* ── Global responsive-image reset ──────────────────────────────────────
+   Any <img> without a scoped size rule falls back to its native/attribute
+   size and can overflow a narrow viewport (guild logo SVGs in particular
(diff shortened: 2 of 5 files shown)


PastLivesReviewBot (approved):
Reviewed and approved by PastLivesReviewBot. Global responsive-image reset is correctly scoped: the bare `img` selector is beaten by every existing class-qualified rule (.pl-guild-hero__img, .pl-guild-gallery__item img, .cp-detail__hero-logo img), so heroes, galleries, avatars, and logos are unchanged. `height: auto` is safe for the attribute-sized square icons and favicons. Coverage spans the public, hub, and CMS entry stylesheets, and the regression guard (tests/css_responsive_images_spec.py) is in the repo BDD idiom, is collected, passes, and fails if the reset is dropped. The earlier version collision is resolved: rebased onto main, VERSION bumped to 0.23.59 with a single new 0.23.59 changelog entry at the top and the 0.23.58 guild-vote entry preserved below. No merge markers, branch is MERGEABLE. Meets PLFOG standards.

PR #193, opened 2026-08-10
fix(css): mobile class-card photo fills its box instead of a tiny thumbnail (v0.23.60)

HexagonStorms:
## Problem
On the public class booking pages (`book.pastlives.space/classes/`), below 768px each class card showed its photo as a tiny 64×48 thumbnail stranded in the full-width 150px image box — a small photo floating in a big empty rectangle.

## Root cause (pre-existing, NOT the v0.23.59 image reset)
The `@media(max-width:768px)` block forced `.cls-img{width:64px;height:48px}` for a compact horizontal `64px | text | price` row. That row was never wired up: `.cls-card` stays `display:flex; flex-direction:column` at every width, so its `grid-template-columns` never took effect and the thumbnail was left stranded.

I verified in-browser on the live page by toggling the v0.23.59 global `img` reset on and off: the card renders identically either way. The global reset is a **no-op** on these cards because `.cls-img` (specificity 0,2,0) wins over the bare `img` rule (0,0,1). So this is a pre-existing layout bug my testing surfaced, not a regression.

## Fix
Remove the vestigial `64×48` override (and the dead `grid-template-columns`) so the phone card renders like the desktop card — the photo fills its full-width 150px media box. Confirmed visually before/after across widths.

## Release
- `VERSION` → `0.23.60`.
- Folded into the existing "Photos now fit your screen" changelog entry (same photos-on-mobile theme, per the repo's one-entry-per-feature rule) with an added bullet about class photos; re-stamped to 0.23.60.

### DIFF
--- a/plfog/version.py
+++ b/plfog/version.py
@@ -2,17 +2,19 @@
 
 from __future__ import annotations
 
-VERSION = "0.23.59"
+VERSION = "0.23.60"
 
 CHANGELOG: list[dict[str, str | list[str]]] = [
     {
-        "version": "0.23.59",
-        "date": "2026-08-09",
+        "version": "0.23.60",
+        "date": "2026-08-10",
         "title": "Photos now fit your screen, especially on phones",
         "changes": [
             "Some photos — guild icons and logos especially — could show up way too big on a phone and "
             "spill off the edge of the screen. Every photo across the site now automatically resizes to "
             "fit, so pages look right no matter what device you're using.",
+            "On the class booking pages, each class photo now fills its space on phones instead of "
+            "showing as a tiny thumbnail floating in an empty box.",
         ],
     },
     {
--- a/static/css/cms-public.css
+++ b/static/css/cms-public.css
@@ -558,8 +558,11 @@ html.dark .cp-page {
 .cp-page .empty{padding:24px;text-align:center;color:var(--text3);font-size:11px}
 
 @media(max-width:768px){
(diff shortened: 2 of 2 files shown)


PastLivesReviewBot (approved):
Reviewed and approved by PastLivesReviewBot. The fix is minimal and correct: it removes a vestigial `@media(max-width:768px)` override in cms-public.css. `.cp-page .cls-card` is `display:flex;flex-direction:column`, so the old `grid-template-columns:64px 1fr auto` never applied — the only live effect of the block was `width:64px;height:48px` on `.cls-img`/`.cls-img-ph`, which stranded a tiny thumbnail in the 150px `.cls-media` box on phones. With it gone, the base rules (`.cp-page .cls-img{width:100%;height:100%;object-fit:cover}` and the matching placeholder rule) govern, so the photo and the letter placeholder both fill their box on mobile like the desktop card. Nothing else at <=768px depended on the 64px sizing (`.cls-body`/`.cls-title`/`.cls-price` stack in the flex column independently), and the preserved `gap:10px;padding:10px 12px` keeps the intended compact spacing. #192s global img reset is untouched and still present, so the responsive-image regression guard still holds. Version discipline is correct: VERSION bumped to 0.23.60 and the class-photo fix folded as a second bullet into the existing "Photos now fit your screen" entry (re-stamped to 0.23.60, at the top so announce_release resolves) rather than a duplicate entry; every other changelog entry is unchanged and there are no merge markers. Meets PLFOG standards.

PR #203, opened 2026-08-13
feat: guild Wishlist tab, plus cancellation copy / past-class recs / sidebar order / calendar-button fixes

HexagonStorms:
One guild feature plus a handful of small, self-contained fixes across the classes surface, the hub sidebar, and the meeting calendar flow.

## 1. Guild Wishlist tab (new)
Each guild page gets a new **Wishlist** tab: a per-guild markdown field for the tools/materials/equipment the guild hopes to receive as donations, plus an optional **donate link** rendered as a Donate button. The tab only appears when a guild has filled something in. Guild leads/staff edit it on the Basic tab of the guild editor.
- Two new `Guild` fields (`wishlist` TextField, `donate_url` URLField), migration `0121`.
- Rides the existing `GuildEditForm` + `can_edit_guild` permission — no new endpoint.
- Body renders through the same sanitised `|guild_markdown` filter the FAQ uses; Donate button is the standard `pl-btn pl-btn--primary`.

## 2. Class cancellation policy copy
Rewrites the default cancellation FAQ (`DEFAULT_CLASS_FAQS`): drops the $50 late fee (late cancels + no-shows are simply non-refundable), moves contact from `studios@` to **`classes@pastlives.space`**. Migration `0052` refreshes existing default rows in place (organiser-reworded rows untouched), same guard as `0050`.

## 3. Never recommend classes that already happened
The "More {category} classes" block used `.public()`, which surfaced runs whose dates were long past. Now uses `.bookable()` — drops any dated run once its first session begins, keeps flexible/undated, soonest-first. Direct URLs to past classes still resolve.

## 4. Alphabetize the hub sidebar
Both variants (admin + member) are alphabetical. Home stays pinned at the top; Help + Wiki stay as the footer pair; everything between is sorted. Each item keeps its permission/feature gate.

## 5. Immediate feedback on "Add to calendar"
The Add-to-calendar POST syncs Google + Discord server-side (a beat or two), and with `hx-swap="none"` the button looked inert until the redirect landed. Added the house `hx-disabled-elt` + `.pl-spin` pattern so it disables and spins the instant it's clicked.

## Tests
- Guild wishlist: tab shows/hides by content, donate button renders, lead can save both fields.
- `public_spec.py`: recommends an upcoming same-category class; never a past one.
- `class_faq_spec.py`: default cancellation copy points at `classes@`, no `$50`.
- Sidebar + meeting workspace render verified via existing specs.
- Full `ruff`, `mypy` (plfog/core/membership/hub), migration check, and 900+ related specs pass locally.

## Notes
- Based on **`main`** (the live bugs are on production, which tracks main).
- `VERSION` -> `1.0.1`. Two member-facing changelog entries share it: "Guild wishlists" (headline) and the cancellation policy update. The sidebar/button tweaks are cosmetic and left out of member notes.
- **Wishlist scope:** v1 keeps it deliberately simple — one markdown text block + a donate *link* per guild (no in-app payment routing). Structured item lists / real per-guild donations would be a follow-up if wanted.

https://claude.ai/code/session_01CrACXrmRgwjpohhY4pyiNR


PastLivesReviewBot (approved):
Reviewed and approved by PastLivesReviewBot. Meets PLFOG coding standards: fat-models/skinny-views, typed, help_text on new fields, reversible migrations, wishlist rendered via sanitising guild_markdown, and BDD tests cover the new behaviour.

PR #208, opened 2026-08-17
fix: anchor all-day events to local midnight so they show on the righ…

HexagonStorms:
…t day

An all-day iCal event carries a bare DATE. _to_datetime stored it at midnight UTC, which in Portland is the previous evening, so every all-day event rendered a day early: an event on the 22nd showed on the 21st on the calendar and in the Discord digest and announcements. Anchor a bare date to LOCAL midnight (the makerspace timezone) instead, so localtime rendering keeps it on its own calendar day.

Bumps VERSION to 1.1.2 with a member-facing changelog entry.

Claude-Session: https://claude.ai/code/session_01RFfZcwdaJHFnSAGsQFHYDn

### DIFF
--- a/hub/calendar_service.py
+++ b/hub/calendar_service.py
@@ -25,12 +25,19 @@ logger = logging.getLogger(__name__)
 
 
 def _to_datetime(val: Any) -> datetime:
-    """Convert a date or datetime value to a UTC-aware datetime."""
+    """Convert a date or datetime value to a UTC-aware datetime.
+
+    A bare ``date`` is an all-day event: anchor it to LOCAL midnight (the makerspace
+    timezone), not UTC midnight. UTC midnight renders a day early everywhere west of UTC —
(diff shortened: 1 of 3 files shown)


PastLivesReviewBot (approved):
Approved. The fix correctly anchors a bare all-day iCal date to local (America/Los_Angeles) midnight via timezone.make_aware before normalizing to UTC, so timezone.localtime renders it on its own calendar day instead of a day early (verified: old UTC-midnight logic rendered Aug 22 as Aug 21 17:00 in Portland; new logic renders Aug 22 00:00). Timed datetimes are unchanged (naive treated as UTC, aware normalized to UTC). Tests are meaningful and would fail under the old behavior, covering all three _to_datetime branches plus an end-to-end all-day path through _parse_ical_events; midnight is DST-safe since US transitions occur at 02:00. VERSION bumped to 1.1.2, changelog is plain member-facing language stamped at the current VERSION, no model change so no migration, and ruff lint/format are clean. 48/48 calendar_service specs pass.

PR #210, opened 2026-08-17
feat: native push notifications (FCM) for the mobile app

HexagonStorms:
## What

Adds **Firebase Cloud Messaging** push for the Capacitor mobile app, alongside the existing browser Web Push. The native WebView cannot receive Web Push, so the app registers an FCM device token and `PushAdapter` fans out to **both** transports over the one `Channel.PUSH` channel. No behaviour change for existing web-push users.

## Why one channel, two transports

`core.events.channels.PushAdapter` already looped a user's `PushSubscription`s and called `send_web_push`. It now also loops the user's `FcmDevice`s and calls `send_fcm`. Events, resolvers, and preferences are untouched: a member's push preference already gates the whole channel.

## Server
- `FcmDevice` model + migration `0062` (unique token per user, `Platform` choice).
- `/push/fcm/register/` and `/push/fcm/unregister/` endpoints, mirroring the `webpush/subscribe|unsubscribe` pair (login-required, JSON, `update_or_create` on the unique token).
- `core/fcm.py::send_fcm` sends via the **FCM HTTP v1 API** using `google-auth` + `httpx` (the same service-account pattern as `core/integrations/google_calendar.py`). Best-effort, reaps a device on `404 UNREGISTERED`, and is a **no-op when `FCM_SERVICE_ACCOUNT_JSON` is blank** (dev/CI). **No new dependency.**
- `FCM_SERVICE_ACCOUNT_JSON` setting.

## App / frontend
- `@capacitor/push-notifications` plugin added; Android project synced.
- `static/js/native-push.js` requests permission and registers the token, guarded by `Capacitor.isNativePlatform()` so it is a no-op in a normal browser. Notification taps route to `notification.data.url`.

## Tests
Sender (success, 404-reap, other-error, transport-error, unconfigured no-op, token minting), both endpoints (auth, method, create/update, validation, own-only, 500), model `__str__`, and the channel fan-out. `respx` for the HTTP v1 call. `core/fcm.py` is at 100%; ruff + mypy clean.

## Before this reaches members (not in this PR)
1. Render env `FCM_SERVICE_ACCOUNT_JSON` is **already set** (web + scheduled-cron), key stored at `~/keys`.
2. A **new AAB build + Play release** is required for the app to actually receive push (the plugin ships in the binary). That is a separate release.
3. **Merging auto-announces** the `1.2.0` "Push notifications" changelog entry to Discord. The Play update will not be live for ~a day, so either merge when the app build is going out, or say the word and I will drop the entry to a silent bump.
4. The launch-fix mobile changes (`verify-aab.sh`, versionCode bump, Gradle gate) are intentionally **not** here.


PastLivesReviewBot (approved):

Native FCM push feature: core/fcm.py is best-effort and never raises to callers (reaps devices on 404 UNREGISTERED, no-ops on blank creds); FcmDevice has help_text on every field, TextChoices platform, meaningful __str__, and a reversible CreateModel migration; the register/unregister views are @require_POST + @login_required with a device-takeover 409 guard and own-device-only delete (no IDOR); the _with_push / _PUSH_ON_BY_DEFAULT curated defaults use dataclasses.replace correctly on the frozen EventType with channel order preserved; tests assert real behavior via respx. The FRONTEND.md rule-numbering defect flagged in the prior review is fixed (rules now 18/19/20/21, no duplicate).

PR #212, opened 2026-08-18
Class announcements, Admin Tools, and out of beta

HexagonStorms:
## What this does

Ships the class/guild/site **announcement composer** as a modular tool, consolidates staff tooling into a single **Admin Tools** page, and removes all remaining "BETA" labeling now that the app is live on the Play Store.

### Announcements

- **Instructors can announce to their class.** New `class` audience on `AnnouncementDraft` sends to everyone with a confirmed registration (the class roster). Rides the notification spine as a new `class_announcement` event (in-app bell, push, email), with a `CLASS_ROSTER` recipient resolver and its own copy.
- **One composer, modular targets.** The same tool serves site, guild, and class audiences. When reached from a guild page or a class admin page it opens **locked** to that audience (no "Who is this for?" picker), because the target is already known.
- **Flattened to one screen.** The old three-step wizard (write, then a separate Email step, then a Discord step) is gone. You write the message once, and the channels (email, push, Discord) are options at the bottom of the same screen, each with a live preview.
- **Cleaner push.** Announcements now render a short, purpose-written push line instead of a truncated email. The sender can also write a custom short version just for the phone (`push_message`).

### Entry points

- **Guild page:** a "Send Announcement" button on the Announcements section, visible only to that guild's leads/staff (and admins).
- **Class admin / teach pages:** a "Send Announcement" button that opens the composer locked to the class roster.
- **Sidebar:** a single **Announcements** entry for everyone who can send (guild leads, guild staff, instructors, admins).

### Admin Tools page

- New `/tools/` page gathers **Announcements, Orientations, Manage Members, Activity, Notification Settings, Site Settings, and Push Notification Test** into one place, with icon cards.
- The sidebar entry and each card are **permission-gated**: an actual admin previewing as a plain member does not see it; a real guild lead/staff sees Announcements + Orientations; an instructor sees Announcements. As soon as any member holds a permission that unlocks a tool, Admin Tools appears for them.
- Moved Activity, Site Settings, Notification Settings, and Manage Members off the sidebar into Admin Tools; kept Meetings in the sidebar.

### Push-test tool

- Re-homed from the Django admin into the hub (`/announcements/push-test/`, admin-only): inspect a member's registered push devices and send them a test notification.
- Deleted the stale `/admin/` announcement composer.

### Out of beta

The app is live on the Play Store, so all visible "BETA" labeling is removed:

- The "BETA v1.x" pill on the public topbar, both login screens, and the admin header now reads just the version ("v1.x").
[...]

PR #213, opened 2026-08-18
fix(mobile): ship FCM push plugin (versionCode 6) + gate on plugin presence

HexagonStorms:
## What
- Bump the Android app to **versionCode 6 / versionName 1.1.1** with `@capacitor/push-notifications` synced into the native project.
- Add a **dex-level plugin-presence assertion** to `mobile/verify-aab.sh`.

## Why
versionCode 5 announced native push (the 1.1.4 changelog line "you get it as a push on your phone") but the shipped APK did **not** contain the push plugin. Because the app runs in server-URL mode, `native-push.js` is served live from members.pastlives.space but no-ops when `Cap.Plugins.PushNotifications` is absent from the binary. Result: no permission prompt, and **zero** `FcmDevice` registrations in production across the entire user base. The server half (endpoints, FCM sender, Firebase config) already shipped in #212; the binary was the missing piece.

## The gate
`verify-aab.sh` previously checked assets, `server.url`, and signature, but not the plugin. It now greps every `base/dex/classesN.dex` for the `capacitorjs/plugins/pushnotifications` class descriptor, using the pipefail-safe `grep -c` form (documented: `grep -q` closes the pipe early and SIGPIPEs `unzip` under `set -o pipefail`). Wired as the existing `verifyReleaseBundle` gradle finalizer, so it runs on every `bundleRelease`.

Proven both directions:
- Real AAB -> `ok native push plugin in dex`, GATE PASSED.
- Synthetic plugin-less AAB (config + assets + signature all present) -> `MISS native push plugin absent from dex`, GATE FAILED (exit 1).

## Verification
Live FCM v1 dry-run with the production service account authenticated to project `pastlives-fog` and was accepted (rejected only the dummy token), confirming Cloud Messaging is enabled and the send path works. App-side `google-services.json` and the server SA are the same project.

## Scope / risk
Two files, mobile build only. No Python, no templates, no `plfog/version.py` change -> no Django runtime impact and no Discord announce on merge. The signed AAB for upload is already built from this change.

https://claude.ai/code/session_01SCPe229cq4iRZbNLLUoqB5

### DIFF
--- a/mobile/android/app/build.gradle
+++ b/mobile/android/app/build.gradle
@@ -16,8 +16,8 @@ android {
         applicationId "app.pastlives.hub"
         minSdkVersion rootProject.ext.minSdkVersion
         targetSdkVersion rootProject.ext.targetSdkVersion
-        versionCode 5
-        versionName "1.1.0"
+        versionCode 6
+        versionName "1.1.1"
         testInstrumentationRunner "androidx.test.runner.AndroidJUnitRunner"
         aaptOptions {
              // Files and dirs to omit from the packaged assets dir, modified to accommodate modern web apps.
--- a/mobile/verify-aab.sh
+++ b/mobile/verify-aab.sh
@@ -56,6 +56,30 @@ else
   fail=1
 fi
 
+# --- Native push plugin must be compiled into the binary ---------------------
+# Push is the one feature the server-URL architecture CANNOT ship on its own:
+# native-push.js is served from members.pastlives.space, but it no-ops unless
+# the @capacitor/push-notifications plugin is inside the APK's dex. The
+# versionCode-5 release announced "push on your phone" while shipping a binary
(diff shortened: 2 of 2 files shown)


PastLivesReviewBot (approved):
Reviewed and approved by PastLivesReviewBot.

Mobile-only change (versionCode 5->6 + a dex-level plugin-presence gate in verify-aab.sh). No Python/Django surface, so fat-models/type-hint/help_text/migration standards are N/A. The shell gate is correct: newline field-split over base/dex entries, pipefail-safe grep -c (consistent with the file's documented grep -q SIGPIPE caveat), numeric compare, and fail-closed when no dex is present. Matches existing verify-aab.sh conventions. Verified passing on a real AAB and failing on a synthetic plugin-less one. Meets PLFOG coding standards.

PR #215, opened 2026-08-19
fix(push): set csrftoken cookie on hub home so native FCM registration works

HexagonStorms:
## What

Native Android push never registered. The app's post-login landing page (`hub.views.home`, where `/` redirects) is where `static/js/native-push.js` runs to register the device's FCM token. It POSTs the token to `/push/fcm/register/` with the CSRF token read from the `csrftoken` cookie.

Django only sets `csrftoken` when a view actually uses it. `hub_home` never did, so the WebView had no cookie, the `X-CSRFToken` header was empty, and every registration POST was rejected. Prod web logs confirm it:

```
Forbidden (CSRF cookie not set.): /push/fcm/register/
```

Result: 0 `FcmDevice` rows, so `PushAdapter` had nothing to send to. Diagnosed live via adb against the physical device (permission granted, plugin present, Firebase healthy) plus the prod Render logs above.

## Fix

Add `@ensure_csrf_cookie` to `hub.views.home`. The cookie is then set on the landing page and persists for the session, so the registration POST (and any other same-origin JS `fetch` POST) carries a valid token. `CSRF_COOKIE_HTTPONLY` is unset (defaults False), so the JS can read it.

Server-side only. Ships via Render, no new Play Store build required.

## Test

`tests/hub/home_spec.py::it_sets_the_csrftoken_cookie_for_the_native_push_registration_post` asserts the cookie is present on the authenticated `hub_home` response.

## Not in this PR

No `VERSION` bump, deliberately: this deploys the fix so it can be verified on a real device before the "push notifications work in the app" changelog entry (and its automatic Discord announcement) go out in a follow-up.

### DIFF
--- a/hub/views.py
+++ b/hub/views.py
@@ -24,6 +24,7 @@ from django.http import Http404, HttpRequest, HttpResponse, HttpResponseBadReque
 from django.shortcuts import get_object_or_404, redirect, render
 from django.urls import reverse
 from django.utils.http import url_has_allowed_host_and_scheme
+from django.views.decorators.csrf import ensure_csrf_cookie
 from django.views.decorators.http import require_POST, require_http_methods
 
 from billing.exceptions import NoPaymentMethodError, TabLimitExceededError, TabLockedError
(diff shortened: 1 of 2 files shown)


PastLivesReviewBot (approved):
Approving: server-side CSRF-cookie fix for native push registration. Root cause confirmed in prod logs; scoped, tested, no version bump (deploy-to-verify before the announce PR).

PR #216, opened 2026-08-19
feat(push): categorize native notifications into 4 Android channels

HexagonStorms:
## What

Native push currently lands in Android's generic fallback channel, so a member gets one blunt on/off for the whole app and no categories (the `Missing Default Notification Channel metadata` warning). This adds four channels the app creates on startup so each category is independently controllable in the phone's system settings:

| Channel | Importance | Carries |
|---|---|---|
| **Urgent** | High (heads-up + sound) | payment failed, class starting soon, class cancelled, freed waitlist seat, event happening now, lease expiring |
| **Guilds** | Default | Guilds + Voting categories |
| **Classes** | Default | Classes + Teaching categories |
| **General** | Default | everything else (Announcements, Billing, Events, Meetings, Membership, Orientations, Spaces) |

## How

- **Client** (`static/js/native-push.js`): creates the four channels via the Capacitor push plugin (Android only; iOS has no channels). Idempotent — never overrides a setting the member later changes.
- **Server** (`core/fcm.py`): `send_fcm` gains a required `channel_id` and sets `android.notification.channel_id`. Channel-id constants live here (shared with the diagnostic push tool).
- **Routing** (`core/events/channels.py`): `push_channel_for(event_key)` picks the channel — explicit urgent keys win, otherwise the event's registry `category` maps to a channel. `PushAdapter` derives it once per message and passes it to every device.
- A completeness spec asserts every registered event maps to one of the four channels, so a new registry category added without a mapping **fails CI** instead of silently defaulting.

Ships entirely server-side (channel creation is served JS; routing is Django). **No new Play Store build.**

## Not in this PR

No `VERSION` bump, deliberately: this deploys so channels + routing can be verified on a real device before the member-facing changelog entry (and its Discord announcement) go out in a follow-up, together with the CSRF registration fix (#215).


PastLivesReviewBot (approved):
Approving: four native push channels + category routing, completeness-guarded, tested. Ships server-side (no Play build). No version bump (deploy-to-verify before the announce).

PR #214, opened 2026-08-19
1.2.0 misc fixes: Title Case headings, FAQ rename, notifications guide, launch email polish

HexagonStorms:
## 1.2.0 misc fixes

Three batches of member-facing copy/behavior fixes.

### Batch 1 — original request
- **Title Case sweep** (~106 headings) per FRONTEND.md Rule 22, across hub/classes/admin/public pages. Sentence-style headings (auth heroes, empty states) and CSS-uppercased eyebrow labels left alone.
- **"Questions" → "FAQ"** on the public class page + its editor (registration questions untouched), matching the guild-page "FAQ" label.
- **New Help Center guide** "Your notification settings" (Getting started, slug `notifications`). Seeding it activates the two legacy help anchors that pointed at it, so `PENDING_LEGACY_TARGETS` is now empty.
- **Release email:** CTA label "Open Past Lives" → **"Visit the Member Portal"**; footer now carries the **"Get it on Google Play"** badge + "officially on the Play Store, iOS coming soon."

### Batch 2 — copy-review comments (2026-08-18)
- **Class image in the `class_published` email** (comment ID 48). The class's hero image now leads the "new class" email, injected as a trusted, app-built `SafeString` (`ClassOffering.email_hero_image_html`) — empty and harmless when a class has no photo.
- **Default-ON event emails** (comments 51–53): `event.guild_published`, `event.community_published`, and `event.lead_meeting_published` now default their EMAIL channel to **ON**. Code-only (no migration): the default is read live at delivery for members with no preference row; explicit opt-outs are preserved. The `/create-event` "also email" escalation now adds only members who opted out (everyone else already gets the automatic email).

### Batch 3 — FOG → "Member Portal" (outward-facing only)
Top-left brand, public site brand/title/footer, hub hero, member-directory + classes-account copy, the Discord release-notify embed footer, the guided-tour title, help-registry copy, and every member-facing help article. **Code identifiers** (`fog_role`, `FOG_ADMINS`, …), CSS/code comments, and admin-internal labels are untouched.

Two judgment calls (easy to change): kept the **"Fog Bot"** name (not renamed to "Member Portal Bot"); left **historical changelog** entries mentioning FOG as dated records.

### Version / changelog
`VERSION` → **1.2.0**, with two member-facing entries (the rename + the notifications guide).

> The go-live email's intro/section body copy is authored in the admin composer at send time (not stored in the repo); this PR changes only the code-backed parts. The corrected paste-ready copy is delivered separately.

### Checks
[...]

PR #217, opened 2026-08-19
test: unbreak main after #214 title-cased headings

HexagonStorms:
## Why

`test` (the full suite + mutation) has been **red on `main`** since #214 merged (2026-08-19 01:10 UTC). #214 title-cased three headings but left the asserting specs on the old lowercase copy, so three content assertions fail. No PR can reach a green CI until this is fixed.

## Fix

Sync the three stale assertions to the current template text. Pure test-copy, no behavior change:

| Spec | Old assertion | Template now |
|---|---|---|
| `classes/spec/views/teach_orientation_spec.py` | `One quick step before you can teach` | `One Quick Step Before You Can Teach` |
| `classes/spec/views/teach_orientation_spec.py` | `You're an instructor` | `You're an Instructor` |
| `tests/hub/calendar_config_forms_spec.py` | `Member events &amp; Google sync` | `Member Events &amp; Google Sync` |

All three verified passing locally.

### DIFF
--- a/classes/spec/views/teach_orientation_spec.py
+++ b/classes/spec/views/teach_orientation_spec.py
@@ -15,10 +15,10 @@ from classes.forms import InstructorOrientationCompleteForm
 from core.models import SiteActivity
 from membership.models import Member, OrgInfoPage, WikiArticle
 
-BANNER = "One quick step before you can teach"
+BANNER = "One Quick Step Before You Can Teach"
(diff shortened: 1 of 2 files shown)


PastLivesReviewBot (approved):
Approving: pure test-copy sync that unbreaks main after #214's title-casing. Three assertions updated to current template text; verified locally.

PR #219, opened 2026-08-20
feat: auto-welcome new members from Airtable (opt-in automation)

HexagonStorms:
## What this does

A new **daily, default-OFF** automation, `welcome_new_members`, on **Site Settings, Automations tab** (toggle + Run now, same registry as every other job). When on, once a day it emails each new **paying, active** member from the membership Airtable their first sign-in link, so they know their Past Lives account is ready and can log in.

## Why a "welcome", not a signup "invite"

New ACTIVE members pulled from Airtable already get a **silently auto-provisioned** passwordless account (the `auto_provision_member_user` signal, which names `airtable_pull` as its main case). Nothing ever emails them, so they do not know the account exists. A create-your-account signup link is therefore the wrong email; this sends the branded `member.login_invite` email (the `/accounts/login/code/` passwordless link) via the existing `Member.send_login_invite()`. It also covers the Pending then Active import path, which leaves an active member with no account, by provisioning first.

## Guardrails

- **Paying only** — `member_type = Standard` (the canonical `Member.is_paying`); excludes guild leads, work-trade, volunteers, employees, contractors.
- **Active only** — a member Sushuma enters as Pending waits until she flips them to Active.
- **Never the standing backlog** — a one-time migration marks every current member already welcomed, so only members added after this ships are eligible (go-forward from deploy; enable it right after deploy so that equals go-forward from enable).
- **Send once** — `welcome_email_sent_at` stamp; and anyone whose email already belongs to an account is skipped automatically (provisioning refuses to double-link).
- **Off until flipped** — new `ScheduledJob.default_enabled`; absence of a state row now means OFF for a default-off job.

## Mechanics

- `core/scheduled_jobs.py`: `default_enabled` field + the new job (daily).
- `core/models.py`: `ScheduledJobStateManager.is_enabled` / `sync_registry` honor the declared default.
- `membership/models.py`: `Member.awaiting_welcome_email()` candidate queryset; `Member.send_welcome_email()` (idempotent wrapper over `send_login_invite()`).
- `core/management/commands/welcome_new_members.py`: thin command, resilient per member.
- Migration `0131` adds the field and backfills the backlog.

## Testing

New specs for the queryset filters, the send/idempotency/failure paths, the command (including per-member resilience), and the default-off registry behavior. Local: full ruff, mypy, `manage.py check`, `makemigrations --check`, and the touched specs all green. CI runs the full suite + mutation.

## Note before merge

`VERSION` is bumped to `1.3.1` with an admin-facing changelog entry. It is an internal tool, so curate or drop that entry before running the Discord announce if you do not want members to see it.


PastLivesReviewBot (approved):
## PastLivesReviewBot re-review: approved (round 4)

Re-reviewed at `8cd89929`. Both round-3 items are closed, and I killed each one with an injected mutant rather than taking the fix on trust. Everything left over from round 3's notes has been answered. Approving.

### Verified fixed — by mutation, not on trust

**1. The command's "not delivered" branch is now real.** Coverage of the new file went 88% → 100%, no partial branches:

```
core/management/commands/welcome_new_members.py    22    0    4    0   100%
```

And the pre-fix command (call `send_welcome_email()`, ignore the bool, one SUCCESS summary line) no longer survives. Injected as a `-p` plugin that replaces `Command.handle` and writes a marker file, so I could confirm the mutant really was the code that ran:

```
FAILED ...::it_counts_an_undelivered_member_separately_and_does_not_claim_success
E   assert 'Welcomed 0 new member(s)' in 'Welcomed 1 new member(s); skipped 0.\n'
FAILED ...::it_reports_a_clean_run_as_success
E   assert 'Welcomed 1 new member(s); 0 not delivered; skipped 0.' in 'Welcomed 1 new member(s); skipped 0.\n'
```

Two of six killed, including the one that matters: a quota incident now reports `Welcomed 0` instead of a green success line.

**2. The `BaseException` guard has a spec.** `except Exception:` in place of `except BaseException:` is now killed:

```
FAILED tests/membership/member_welcome_email_spec.py::describe_a_worker_shutdown_mid_send::
  it_releases_the_claim_when_the_worker_is_killed
E assert datetime.datetime(2026, 9, 8, 2, 56, 20, 86225, tzinfo=UTC) is None
```

That is exactly the stranded claim the two words exist to prevent, and the spec mirrors the sibling `send_funding_results` / `send_pending_funding_results` guards it was asked to mirror. The `KeyboardInterrupt` really does travel the whole path: `core.email.send` swallows only `Exception` under `best_effort`, `_fan_out`'s own `except BaseException` releases the ledger slot and re-raises, and `send_welcome_email` gets it.

### The `money_job` question — the right call, and wired correctly

Not reusing `money_job` is correct, and for a better reason than the one in the PR body. `money_job=True` carries three things, not one: the "charges cards" badge (false for an email job), `it_marks_only_bill_tabs_as_a_money_job` (a spec that exists to keep that set honest), and the pinned/non-toggleable treatment `bill_tabs` gets — and this job must stay toggleable, because the whole feature is a switch an admin turns on. A separate `confirm_before_run` was the only option that did not require weakening an existing invariant.

The wiring is complete, and I checked it by rendering rather than by reading. Against a scratch copy of the tree, the Automations tab renders:

* `open-confirm', 'run-welcome_new_members'` on a `hub-btn--danger` button,
* the modal include expanded — "Send welcome emails now?", "Emails cannot be recalled",
* `name="run_job" value="welcome_new_members"` as the modal's hidden field,
* the "emails members" badge,
* and **no** `data-run-now` plain-submit button for that job.

POSTing the modal's payload dispatches `call_command("welcome_new_members")` exactly once, redirects, and leaves every toggle alone — the modal carries no management form, so `_persist_automation_toggles` no-ops exactly as its docstring promises. `sync_registry` seeds the row `enabled=False` on first render, so the card ships paused and the toggle reads off.

`bill_tabs` is untouched: badge, run-cell branch and modal include are identical apart from the added `or`.

### Also verified

* **Migration graph.** No multi-leaf app anywhere in the project. `core.0080 → 0079_wiki_page_store`, `membership.0168 → 0167_wiki_page_store`, and the branch's merge-base *is* current `origin/main` (`3f687c6c`), which carries no competing migration in either app. `manage.py check` clean, `makemigrations --check` reports no changes.
* **Announce safety.** `VERSION = "1.45.1"`, `main` is `1.45.0`, exactly one changelog entry at `1.45.1` and it is the top entry. This merge posts that entry and nothing else.
* **The changelog copy is now accurate**, sentence by sentence. "only emails paying members who are active in Airtable" — `member_type=STANDARD`, `status=ACTIVE`, `airtable_record_id__isnull=False`. "never the same person twice" — the ledger stamp, and the manual admin button now stamps too, which is what made that sentence true rather than aspirational. "stays off until an admin turns it on" — `default_enabled=False`, seeded off, `is_enabled` falls back to off. The "never someone who already has access" claim that had no `last_login` filter behind it is gone.
* **Suites.** `core/spec` + `tests/core` + `tests/hub/site_settings_automations_spec.py`: 2519 passed. `tests/membership` + `tests/hub/admin_views_spec.py`: 2925 passed. `ruff check` and `ruff format --check` clean. (`mypy` would not run in my worktree — the django-stubs plugin fails to construct on this machine, an environment problem rather than a signal about this diff. CI covers it.)
* **PR body** is rewritten and now matches the tree: migration `0168`, `VERSION` 1.45.1, and the automatic-announce behaviour described correctly.

---

### Notes, not blocking

**The new template wiring is the one part of this PR nothing protects.** A mutant that reverts all three template changes at once — drop the badge, put the run cell back to `{% if row.job.money_job %}`, comment out the modal include — passes the whole suite (45 specs green under a shadow template directory). The registry specs still pass, because they assert on the flag, not on what the flag renders. So the guard rail can silently become a one-click unrecallable mass mail again and nothing fails. The repo already has the spec shape for this six lines away in the same file (`it_renders_a_confirm_modal_and_danger_button_for_bill_tabs`); the sibling is four assertions. Not blocking, because the confirm modal was the optional item and it is implemented correctly — but it is worth the four lines, here or next time you are in that file.

**`confirm_before_run` renders a hardcoded "emails members" badge**, which reproduces the flag-carries-a-label coupling that was the reason not to reuse `money_job`, one level down. Fine with a single user of the flag and correct for this job. Worth splitting the badge out only if a second irreversible-but-not-email job ever wants the modal.

**The template comment above the two includes** still opens "Money-job guard for the Automations tab's 'Charge member tabs' Run-now" and now sits above two modals, only one of which it describes.

**Cosmetic:** the command's `not delivered, will retry next run` warning also covers the `claimed == 0` path, where the member will *not* retry (another caller stamped them). Unreachable inside a single run, since the candidate queryset filters on the same column.

---

Reviewed via `gh pr diff` plus read-only runs in a clean worktree. Mutants were injected as `-p` plugins from a temp directory, each writing a marker file so I could confirm the mutant was the code that ran; the template mutant and the render/POST probes ran against a `/tmp` copy. The branch was never checked out or modified.

PR #220, opened 2026-08-22
Member profile Website/Social sections, instructor discount codes behind site flag, CMS tab rename

HexagonStorms:
Board request from Lee, plus a follow-up scope addition.

## 1. Discount code toggle in the member directory: investigated, nothing to remove

The "Can approve their own discount codes" toggle does not appear in the member directory or on any member's own profile. It renders in exactly one place: the admin-only Manage Members edit page (`/manage/members/<pk>/edit/`, `MemberAdminEditForm`), which is where its help text says it belongs. A new regression spec (`tests/hub/discount_toggle_scope_spec.py`) locks this in: absent from the directory and profile settings, present on the admin edit page.

## 2. Website and Social sections on member profiles

Per ADR-0001, fixed website/social columns were previously folded into the generic `MemberContact` model, so this extends that model instead of adding `Member` columns:

- New `MemberContact.kind` (`website` / `social` / `other`, default `other`) with a backfill migration that classifies existing rows by label keywords (real reverse included).
- Profile settings page: the Contact Methods block is now three sections (Website, Social, Other Contact Methods) sharing the single existing formset; each section has its own "+ Add" button that stamps the hidden `kind` on the cloned row.
- Member directory cards and the public instructor page render website/social links with a new monochrome `social_icon.html` component (instagram, youtube, facebook, tiktok, linkedin, x, generic link); "other" rows render exactly as before.
- Existing per-link visibility toggles (directory / instructor page) are unchanged.

## 3. Instructor discount codes behind a site flag (default OFF) + CMS tab rename

- New `SiteConfiguration.instructor_discount_codes_enabled`, default off. On prod launch, instructors can no longer create/manage their own discount codes.
- Flag off: the teach portal nav tab, the per-class sub-tab, and the inline class-form section are hidden, and all six teach discount views redirect to the teach dashboard with an info message (matches the existing feature-flag gate convention; not a 403).
- Flag on: today's behavior is fully restored. Admin discount views and the per-member self-approve toggle are untouched either way.
- The Site Settings "Legacy CMS" tab is renamed to "CMS" (nav label and "CMS Sync" heading); the new toggle lives on that tab. Internal `?tab=legacy-cms` slug, `legacy_cms_*` identifiers, and strings describing the legacy Drupal data source are deliberately unchanged.

## Version

`VERSION` 1.3.0 to 1.4.0 with two changelog entries stamped 1.4.0 (profile links; admin-only discount codes).

## Testing

- Full suite green in the worktree: 8160 passed, 0 failed. `manage.py check` clean, `makemigrations --check` clean, ruff clean.
- New specs: discount toggle scope regression, `MemberContact.kind` model/migration/form/rendering, flag off/on gating for all six views and all three entry points, admin unaffected in both flag states, CMS rename assertions.
- Existing `classes/spec/` suite keeps its pre-flag posture via an autouse fixture that enables the flag for that tree only; flag-off behavior is covered explicitly in the new spec file.
- Reviewed by two independent review passes (both APPROVE).

https://claude.ai/code/session_01DrNB3isKpbSY5zcTCQT5YN


PastLivesReviewBot (approved):
Reviewed and approved by PastLivesReviewBot. Code meets PLFOG coding standards: help_text and TextChoices on new fields, typed decorator gate matching the existing feature-flag convention, data migration with a real reverse, BDD specs covering flag on/off and both admin states.

PR #222, opened 2026-08-24
feat(hub): self-service account deletion (Apple 5.1.1(v) compliance)

HexagonStorms:
## Why

Apple App Review rejected the iOS app under **Guideline 5.1.1(v)**: an app that supports account creation must also offer in-app account deletion. The iOS app is a Capacitor server-URL shell of members.pastlives.space, so this Django-side feature ships to the iOS webview with no app rebuild. Members get a real "Delete my account" flow; the reviewer gets something they can drive and record with the demo account.

## What it does

Adds a **Delete Your Account** flow under a new **Account** tab in Settings. Strategy is **anonymize + lock**, never a hard delete:

- Scrubs the member's PII in place (name, preferred name, phone, Discord, pronouns, bio, billing name, emergency contact, instructor bio, commission note, profile photo file).
- Deactivates the linked `User` (`is_active=False`) and frees both `username` and `email` to a non-routable placeholder, so the real address can be re-invited later.
- Clears device/notification state: push subscriptions, FCM devices, notifications, notification preferences, and the `UserProfile` PII mirror.
- Strips elevated authority: nulls `guild_lead`, removes `GuildStaffMembership` rows, revokes all `AdminCapability` grants, resets `fog_role` to member and syncs `is_staff`/`is_superuser` off.
- Deletes allauth `EmailAddress` rows and `MemberEmail` aliases (the other email stores).
- Sets `status=FORMER`, stamps a new `Member.deleted_at`, hides from the directory, and logs an `ACCOUNT_DELETED` SiteActivity.
- The view then signs the member out and redirects to login. Idempotent and wrapped in one atomic transaction.

## Why keep the rows instead of hard-deleting

Two reasons, both load-bearing:
1. `billing.Tab` and `classes.ClassOffering.instructor` are `PROTECT` relations. Anonymize-in-place never triggers them.
2. Deleting the `User` would let the passwordless login-by-code path auto-recreate it from a leftover `Member._pre_signup_email`. Keeping an inactive `User` with a freed email/username defeats every resurrection branch. Re-entry is therefore leadership-driven (a fresh Invite), by design.

## Security note (defense in depth, not a live-bug fix)

`GoldenTicketConfirmLoginCodeForm` (the app-review master-code login) now refuses a deactivated pending user. allauth's `pre_login` already blocks inactive accounts on every path, so this closes nothing that was open; it makes the invariant checkable in `plfog/adapters.py` alone and guards against a future allauth change.

## Scope discipline

Does not touch anything the finance team owns outside this app: `billing.Tab`, Stripe, the paid subscription, Airtable. Deletion is plfog-account-only.

## Convention deviation (called out for review)

The service sets `member.fog_role` directly then calls `member.sync_user_permissions()`, instead of `set_fog_role()`. `set_fog_role()` is permission-gated and raises for a member acting on themselves. This never sets `is_staff`/`is_superuser` directly (it routes through `sync_user_permissions()`), so it honors what the membership convention actually protects.

## Tests

- New: `tests/membership/services/account_deletion_spec.py`, `tests/hub/account_delete_spec.py`, `tests/plfog/login_by_code_inactive_spec.py`; extended `tests/hub/forms_spec.py` and `tests/plfog/adapters_spec.py`.
- 151 passing across the targeted set; 100% branch coverage on the new service, view, form, and adapter guard.
- `manage.py check` clean; `makemigrations --check` clean.

## Migrations

- `membership/0134_member_deleted_at.py` (AddField, auto-reversible).
- `core/0064_alter_siteactivity_kind.py` (new `SiteActivity.Kind` choice; modern Django tracks choices in migration state, so the `AlterField` is required even though it is a schema no-op).

## Open risks (defaults taken, flagged)

- A member who was a guild lead/staff is unlinked cleanly; `audit_guild_leads` surfaces any now-leaderless guild on its next run. No synchronous admin alert (YAGNI).
- An instructor's public class links survive (slug kept, no 404s), but they can't log back in to manage a class. Reassigning/cancelling classes is a separate business decision.
- A second device already logged in stays valid until its session naturally expires (pre-existing Django behavior, app-wide, not introduced here).
- Admin-authored `notes` on the member are retained (staff data, not the member's own PII).

https://claude.ai/code/session_01BuyoZKiSENczcxAv1GYN2g


PastLivesReviewBot (approved):
Reviewed and approved by PastLivesReviewBot.

Independent review flagged three PII-scrubbing gaps: MemberContact rows, instructor_slug, and UserProfile accessibility_note / custom_question_answers. All three are fixed in 0abfa1c with tests and 100 percent coverage on the deletion service. Lockout across every login path (including the golden-ticket guard), transaction safety and idempotency, view auth, and the fog_role / sync_user_permissions handling were all verified. Meets PLFOG coding standards.

PR #224, opened 2026-08-24
fix(discord): point #classes links at the legacy Drupal site for now

HexagonStorms:
## What

Class links posted to the #classes Discord channel (the 15 minute new class announcer and the weekly Monday digest) now send members to the legacy Drupal site at classes.pastlives.space instead of the new class CMS, since Drupal is still the public sign up surface for now.

- Adds `ClassOffering.legacy_public_url`: the offering's Drupal page (`https://classes.pastlives.space/class/<alias>`), rebuilt from the imported slug (the import derived the slug from the Drupal path alias; a `-legacy` collision suffix is stripped to recover the alias). Empty for locally authored offerings.
- `hub/discord_class_posts.py` links every class through `_showcase_url`: the Drupal page when one exists, falling back to our own public page for classes that never existed on Drupal (a Drupal link for those would 404).
- The "Browse all classes" digest footer and digest embed URL now point at the Drupal site root.
- When the new class site takes over, reverting is one swap back to `offering.public_url`.

## Testing

- New specs: Drupal links in the digest and announcer embeds, plus `legacy_public_url` model specs (Drupal page, empty for local offerings, collision suffix stripping). Existing specs double as the fallback coverage.
- Full suite: 8423 passed, coverage 98.37% (gate 98%).
- `ruff format --check`, `ruff check`, `manage.py check` all clean. Local mypy still crashes on the django-stubs plugin (pre-existing; CI runs the real check).
- Verified live: `https://classes.pastlives.space/class/sewing-pattern` returns 200 and matches the alias shape the import stores.

Version 1.6.1 with a member facing changelog entry (the announcer has been live since 1.2.x, so the link change is member visible).

https://claude.ai/code/session_01XAn31uCEN65NrCCa2BvNsA

### DIFF
--- a/classes/models.py
+++ b/classes/models.py
@@ -574,6 +574,20 @@ class ClassOffering(HeroCropMixin, models.Model):
         return _absolute_url(reverse("classes:public_class_detail", kwargs={"slug": self.slug}))
 
     @property
+    def legacy_public_url(self) -> str:
+        """This class's page on the legacy Drupal site, or ``""`` for locally-authored offerings.
+
+        The import derives ``slug`` from the Drupal path alias (``/class/<alias>``), appending a
+        ``-legacy`` suffix only when that alias collides with an existing local slug — so stripping
+        the suffix recovers the alias.
+        """
+        from classes.import_service import LEGACY_CMS_BASE
+
+        if not self.legacy_cms_id:
+            return ""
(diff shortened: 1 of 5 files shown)


PastLivesReviewBot (approved):

PR #230, opened 2026-08-26
fix(discord): modal responses via interaction callback (v1.9.3)

HexagonStorms:
Hotfix: /poll and /create timed out for members after v1.9.2 because Discord drops inline type-9 (modal) bodies from HTTP interaction endpoints. Our server verified fast (~110-250ms, valid JSON, type-4 replies unaffected) — the body was discarded on Discord's side.

Fix: every modal response now goes out over POST /interactions/{id}/{token}/callback (the same proven shape as the deferred path, which returns {} inline after the REST call). If Discord rejects the modal payload, the member gets an ephemeral reply carrying Discord's own error detail instead of a dead timeout — so any residual schema issue becomes visible on the next attempt instead of invisible.

220 targeted tests green (new callback specs + dispatch routing specs + both command spec files). Changelog entry re-stamped to 1.9.3 (fix to a live announced feature).

https://claude.ai/code/session_01HXTzkHBXaNLM9nQ41Udtgu

### DIFF
--- a/core/events/discord_commands.py
+++ b/core/events/discord_commands.py
@@ -20,7 +20,14 @@ from dataclasses import dataclass, field
 from importlib import import_module
 from typing import TYPE_CHECKING, Literal
 
-from core.events.discord_interactions import ack_deferred, error_reply, reply, send_followup, unlinked_reply
+from core.events.discord_interactions import (
+    ack_deferred,
+    error_reply,
+    reply,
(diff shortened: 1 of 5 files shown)


PastLivesReviewBot (approved):
Reviewed and approved by PastLivesReviewBot. Hotfix routing verified; meets PLFOG standards.

PR #241, opened 2026-08-27
My Classes filter on admin lists (1.16.0)

HexagonStorms:
## What ships

An always visible **My Classes** filter (instructor OR author) on `/classes/admin/classes/` and `/classes/admin/registrations/`, so staff can narrow to only the classes they teach or authored.

- New `ClassOfferingQuerySet.hosted_by(member)` = `Q(instructor=member) | Q(created_by=member)`, backing both lists and the registrations CSV export via `class_offering__in` (one definition of "mine").
- **None-guard**: a viewer with no member record matches nothing (a dropped guard would leak every NULL-instructor/NULL-author class).
- The shared search box (`components/table_search.html`) gains opt-in `preserved_fields` + `clear_url` so search and clear no longer drop the active filters. The four other callers render byte-identically (no params passed).
- Status pills and the instructor form now preserve `mine`/`q`/`sort`/`dir`; bogus `mine` values are normalized out of computed URLs.
- Pill is always visible (not gated on `instructor_slug`, which was the original bug that hid it). `.pl-help` tooltip states the instructor-or-author semantic. `mine_count` ignores search and instructor, matching the sibling status counts. Mine-first empty states with a clears-only-mine link.

## Spec

`docs/superpowers/plans/2026-08-27-admin-mine-filters.md` (fogstorm + adversarial UX review; the review caught the shared-search-box blocker that would have silently dropped the filter, now fixed via the opt-in preserved-params mechanism).

## Test evidence

`.venv/bin/pytest classes/spec/views/ classes/spec/models/ tests/template_comment_lint_spec.py -q` → 1050+ passed (admin_classes_spec, admin_registrations_filters_spec, registration_export_spec new suites incl. the memberless-leak and search-preservation tests). `ruff check`, `manage.py check`, and pre-push mypy all clean.

### DIFF
--- a/classes/models.py
+++ b/classes/models.py
@@ -205,6 +205,16 @@ class ClassOfferingQuerySet(models.QuerySet["ClassOffering"]):
     def for_instructor(self, instructor: "Member") -> "ClassOfferingQuerySet":
         return self.filter(instructor=instructor)
 
+    def hosted_by(self, member: "Member") -> "ClassOfferingQuerySet":
+        """Classes this member teaches or authored (instructor OR created_by).
+
+        Both are direct single-valued FK comparisons on the row, so no join can
+        multiply rows — no ``.distinct()`` needed. ``member`` must be a real
+        Member: callers guard ``None`` (passing ``None`` would match every class
+        with a NULL instructor/author, the opposite of intended).
+        """
+        return self.filter(Q(instructor=member) | Q(created_by=member))
+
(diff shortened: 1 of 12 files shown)


PastLivesReviewBot (approved):
Reviewed and approved by PastLivesReviewBot. My Classes filter narrows both admin lists (and the CSV export) to instructor or author, with the NULL FK leak defended and tested at all three entry points, querystring preserved across every control, permissions only narrowed never widened, and no N+1. Meets PLFOG coding standards.

PR #245, opened 2026-08-27
fix(orientation): open Stripe checkout via full page POST + restore guild calendar hover

HexagonStorms:
## What

Two fixes for the guild orientation surface, both discovered while testing paid orientation booking on prod.

### 1. Paying for an orientation was blocked by CORS (the real bug)
The paid book slot dialog posts through the shared `confirm_modal.html`, whose `<form method="post">` is intercepted by the hub's page-wide `hx-boost`. A boosted form is submitted as an XHR, and when the view returns a 302 to `checkout.stripe.com`, the browser blocks the cross origin XHR redirect as a CORS error. Result: tapping **Continue to Payment** never reached Stripe.

The public class registration form already handles this by opting out with `hx-boost="false"` (`templates/classes/public/register.html:81`). This PR applies the same escape:
- New opt-in `confirm_no_boost` flag on `components/confirm_modal.html` (documented in its header comment), set on the paid book slot dialog only.
- `hx-boost="false"` on the custom time request form when the guild charges.

Both now submit as genuine full page POSTs and follow the 302 to Stripe natively. Non paid paths are unchanged (still boosted).

### 2. Guild calendar hover threw a ReferenceError
`partials/calendar_content.html` calls `hoverEvent()` / `unhoverEvent()`, but those methods lived only in the community calendar's Alpine scope. On a guild page (`guild_calendar_app.html`) hovering an event logged `Uncaught ReferenceError: hoverEvent is not defined`. Added the two methods to the guild calendar scope.

## Testing
- `tests/e2e/orientation_booking_spec.py -m e2e` passes (exercises the shared confirm modal on its non paid path, confirming the opt-in flag did not disturb the default).
- `tests/template_comment_lint_spec.py` passes.
- Pre-push ruff + mypy clean.

## Version
Patch bump to 1.19.1 with a member facing changelog entry (this is a fix to a live feature).

### DIFF
--- a/plfog/version.py
+++ b/plfog/version.py
@@ -2,9 +2,19 @@
 
 from __future__ import annotations
 
-VERSION = "1.19.0"
+VERSION = "1.19.1"
 
 CHANGELOG: list[dict[str, str | list[str]]] = [
+    {
+        "version": "1.19.1",
+        "date": "2026-08-27",
+        "title": "Smoother orientation checkout",
+        "changes": [
+            "If your guild charges for orientations, tapping Continue to Payment now opens our "
+            "secure checkout right away. We also tidied up how guild calendars respond when you "
+            "hover over an event.",
+        ],
+    },
     {
         "version": "1.19.0",
         "date": "2026-08-27",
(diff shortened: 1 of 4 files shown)


PastLivesReviewBot (approved):
Reviewed and approved by PastLivesReviewBot. Surgical boost opt-out matching the existing register.html pattern; confirm_no_boost is opt-in so other confirm modals are unchanged. Calendar hover methods restored to the correct Alpine scope. Meets PLFOG standards.

PR #250, opened 2026-08-28
feat(billing): Reconciliation dashboard, per-transaction adjustments, monthly snapshots

HexagonStorms:
Admin-only. Answers "who do I pay this month, and how much" from every class payment, orientation payment, and member tab, split by the makerspace revenue rules. All four phases in one PR (per owner request). No VERSION bump (admin-only tooling, not a member-facing announce).

Spec: `docs/superpowers/plans/2026-08-27-reconciliation.md`.

## Allocation engine (`billing/reconciliation.py`, read-time, no new money table)
Walks the same three streams as `payments_panel`, attributes each payment to guild / instructor / orientator / Past Lives shares, aggregates per recipient for a chosen month (date range adjustable). Integer cents, half-up rounding, the largest producer share absorbs any 1c drift so each transaction sums exactly to its net-of-refunds. Unset instructor/guild rolls its share to PL with a visible note + banner count (table always sums to collected; the gap is loud, never dropped). CSV export + a print-optimized page (Print / Save as PDF, the same browser-print path the QR flyers use, **no new dependency**).

## Editable splits + per-transaction control
Six split-percent fields on `BillingSettings` (orientation 70/15/15, class 70/10/20), each triad validated to sum to 100 in `ReconciliationSettingsForm`. Read-only split line on the guild Orientations tab. `TransactionAdjustment` (one row per transaction, unique) lets admins adjust a single transaction's split or omit it; tab charges are omit-only.

## Monthly snapshots
`ReconciliationSnapshot` freezes a closed month with history/detail/delete CRUD, mirroring the voting funding-snapshot shape but with **no email, no event, no Airtable sync**. Per-guild voting allocation folds into each guild's total (projected between snapshots, frozen at snapshot). A month-end scheduled job takes it automatically; registered + `_DISPATCHER_ALWAYS` parity tuple updated.

## Gating & folds
Reconciliation tab is **actual-admin only** (like Settings/Stripe, 403 otherwise), not merely `BILLING_APPROVER`. The old Reports page folds in; its URLs 301-redirect to the tab (no bookmarks 404).

## Testing
- `reconciliation_spec.py` (engine) + `reconciliation_views_spec.py` (views/forms/command/snapshot): 70 passed. Engine at **100% branch coverage**. Balance tests prove odd amounts sum exactly and drift lands on the producer, not PL.
- Full `tests/billing/` + parity: 604 passed. `manage.py check` clean (constraint name 16 chars, E034-safe); no migration drift; migrations 0014 (billing) + 0069 (core) apply and reverse cleanly.
- No e2e navigates the Payments dashboard or Reports page (only a still-valid URL-name reverse); e2e lane unaffected.

Migrations: `billing/0014` (6 AddFields + TransactionAdjustment + ReconciliationSnapshot), `core/0069` (SiteActivity.Kind).


PastLivesReviewBot (approved):
Reviewed and approved by PastLivesReviewBot. Money identity proven exact (rounding conservation incl. 50/50 tie to producer, refunds, omit, override); (source_kind, source_pk) adjustment keying collision-safe; all 10 endpoints actual-admin gated. Delta-verified: negative-net rows now show so the table always balances; malformed stored overrides fall back to the standard split without 500ing; the malformed case no longer inflates the unassigned banner; mutating + snapshot-detail endpoints pinned to 403 for non-admins. Meets PLFOG standards.

PR #254, opened 2026-08-29
feat(settings): demo-visibility toggles to gate seeded demo content

HexagonStorms:
## What this ships

Two default-off Site Settings switches so seeded demo content can sit on production hidden from real members until an admin reveals it for a live demo.

Spec: `docs/superpowers/plans/2026-08-28-demo-visibility-toggle.md`

### `display_demo_classes`
Gated at `ClassOfferingQuerySet.public()` (the single choke-point every public class surface routes through, directly or via `bookable()`): when off, excludes `slug__startswith="demo-"`. Covers the catalog, class detail, register, community calendar, guild calendars, Discord class posts, and hero counts in one edit. Admin/teaching querysets (`.all()`, `editable_by`, `for_instructor`, `hosted_by`) do NOT use `public()`, so staff always see and manage demo classes.

### `display_demo_guild`
New `GuildManager.visible()` = active guilds, plus the example (Cartographers) guild when the flag is on. Routed only through `directory()` and the hub sidebar (`_get_hub_context`). Voting, the ballot, and every other `is_active` filter are deliberately left untouched, so the example guild never enters funding regardless of the flag (its safety contract holds). `EXAMPLE_GUILD_SLUG` moves to `membership/models.py` and is re-exported from `example_guild.py`.

### UI
Both fields added to `SiteSettingsForm` and rendered as toggles in the Features tab; added to the General-tab exclusion list so they render once (avoiding the `member_directory_public` double-render trap). Migration core `0070`.

## Tests
`tests/hub/demo_visibility_flags_spec.py` (7): default-off hides demo classes from `public()`/`bookable()` and the example guild from `directory()`/`visible()`; on reveals both; the example guild stays out of `is_active`-only lists even when on; the base class queryset is never gated. Broad regression run (catalog, guild directory/voting/visibility, demo_data, tours, help_keys): 173 + 80 passed. `ruff` + `mypy` (187) + `manage.py check`: clean.

## Regression note
Gating `public()` with a default-off flag would hide `demo-` classes from any test expecting them public. Verified none do: the tours resolver and `public_spec` use non-`demo-` factory slugs, `demo_data_spec` asserts via direct filters (not `public()`), and `free-demo-2` starts with `free-`, not `demo-`.

## Operations
Ship this, then seed prod with `demo_data` (content lands hidden). For the demo, flip both toggles on in Site Settings > Features; turn off (or `demo_data --remove`) after.


PastLivesReviewBot (approved):
Reviewed and approved by PastLivesReviewBot. Demo-visibility toggles meet PLFOG standards; adversarial review found and verified the fix for the /whats-on choke-point leak, delta re-review clean.

PR #256, opened 2026-08-29
feat(tours): pause/resume guided tours + fix opening popover landing above the fold

HexagonStorms:
## What

Adds **pause / resume** to the guided tours, and fixes a pre existing bug that could strand the opening popover above the fold.

### Pause / resume
- Every tour popover gets a **Pause** control (left of Back/Next).
- Pausing lifts the spotlight **without ending the tour**: it stores a paused run in `sessionStorage` (with the resume path, the state URL and the csrf) and shows a persistent **"Tour paused"** pill.
- The presenter can click around, open things, and **navigate to other pages**; the pill follows them because the run survives navigation. It shows even on pages that carry no tour payload.
- **Resume** drives back to the paused step and re spotlights it (reusing the existing `?tour=&step=` navigation, the most tested resume path). The pill's **end** button stops the tour for good and records the dismissal.
- Pausing writes **nothing** to the DB (verified): it is a hold, not a stop, so closing the tab does not consume the tour.

### Bug fix (found while testing, demo facing)
Driver.js sometimes anchored a centered (targetless) popover **outside the viewport** on its first paint after a page load, stranding the opening step of every tour above the fold (reproduced deterministically: first landing `y=-360`, later ones `y=273`). Targetless popovers are now pinned to true viewport center. This also stabilizes the previously flaky completion e2e.

## How it renders everywhere
`_tour.html` now emits the pill markup and loads `pl_tour.js` on every member page (the payload block stays gated behind `tour_json`); the script is an inert-when-idle guarded IIFE.

## Tests
`tests/e2e/guided_tour_spec.py` (run with `-m e2e` on Postgres):
- pause and resume on the same page (progress unchanged)
- keep the pill while wandering to a no payload page, then resume back
- end a paused tour from the pill (records `dismissed`)
- existing completion lap now green across repeated runs

147 Python tour specs pass; `manage.py check` clean; ruff + mypy green via the pre-push gate.

## Versioning
`VERSION` held at `1.21.0` on purpose, matching the rest of this round, so **no Discord announce fires before the Sept 1 demo**. It ships with the batched release.

### DIFF
--- a/static/css/hub.css
+++ b/static/css/hub.css
@@ -6097,6 +6097,80 @@ mark { background: color-mix(in srgb, var(--color-tuscan-yellow) 30%, transparen
     }
 }
 
+/* ── Paused-tour pill (§6.2) — a persistent Resume affordance that survives
+   navigation while a tour is paused, so the presenter can roam and come back ── */
+.pl-tour-resume {
+    position: fixed;
+    right: 1rem;
+    bottom: 1rem;
+    z-index: 10000;
+    display: flex;
+    align-items: center;
+    gap: 0.625rem;
+    padding: 0.5rem 0.5rem 0.5rem 0.875rem;
+    background: var(--hub-card-bg);
+    color: var(--hub-text);
+    border: 1px solid var(--hub-border);
+    border-left: 3px solid var(--color-tuscan-yellow);
+    border-radius: 999px;
+    box-shadow: 0 6px 20px rgba(0, 0, 0, 0.28);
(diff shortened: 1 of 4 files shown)


PastLivesReviewBot (approved):
Reviewed and approved by PastLivesReviewBot. Pause/resume runtime plus the centered-popover fix meet PLFOG standards; adversarial delta review clean, 9/9 tour e2e green.

PR #258, opened 2026-08-29
feat(catalog): show every guild type, even with no classes yet

HexagonStorms:
The classes catalog previously dropped any guild type (Category) that had zero currently-bookable classes, so the full range of guilds was invisible to members browsing.

## Change
`classes/views.py` `public_list`: instead of filtering categories to those with a bookable class, list every Category with a `class_count` that falls back to 0. `Category.Meta.ordering` (sort_order, name) keeps them ordered. The `?category=` / `?guild=` filters are independent and already fail soft, so nothing else changes.

Result: the "Guild Types" hero stat and the guild-type filter dropdown now show all guild types, with "(0)" when a guild has nothing scheduled yet.

## Tests
Added `describe_all_guild_types_show` in `classes/spec/views/public_spec.py`: a zero-class guild type still appears (count 0) alongside a stocked one, and the hero stat counts every guild type. Full `public_spec.py` (65) + sibling catalog specs (15) green; ruff clean.

VERSION intentionally held at 1.21.0 (demo freeze, no Discord announce before Sept 1).

### DIFF
--- a/classes/spec/views/public_spec.py
+++ b/classes/spec/views/public_spec.py
@@ -946,3 +946,36 @@ def describe_detail_hero_fallback():
         assert resp.status_code == 200
         assert "cp-detail__hero-logo" in resp.content.decode()
         assert "img/favicon.png" in resp.content.decode()
+
+
+def describe_all_guild_types_show():
+    """Every guild type (Category) appears in the catalog, even with zero bookable classes."""
+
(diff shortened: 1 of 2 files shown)


PastLivesReviewBot (approved):
Reviewed and approved by PastLivesReviewBot. Independent adversarial review passed (with delta re-review). Meets PLFOG standards.

PR #259, opened 2026-08-29
feat(voting): distinct guilds per tier, 1st-only allowed

HexagonStorms:
Item 8 from the round. Two rule changes to guild funding votes:
1. A member cannot pick the same guild in more than one tier (all chosen tiers must be distinct guilds).
2. Only the 1st choice is required now; 2nd and 3rd are optional (a member can vote for just one guild). A 3rd choice still requires a 2nd (no gaps).

## Changes
- `VotePreference.guild_2nd`/`guild_3rd` are now `null=True, blank=True` (migration `0147`, a plain reversible AlterField). `cast_ballot` accepts `None` for them.
- `VotePreferenceForm`: 2nd/3rd `required=False`; `clean()` dedups across the non-blank chosen tiers and enforces the no-gap rule. 1st stays required.
- **None-safety across every read path** (the risky part of making tiers optional): `FundingSnapshot.take`, `vote_analyzer`, the closing-soon reminder email context, and the Discord `/vote` reply all guard the optional tiers. `calculate_results` skips a blank optional rank but still raises on a blank 1st choice.
- Voting page: intro copy + the current-preferences table updated (2nd/3rd rows only render when set).

## Scope note
The Discord `/vote` command still requires all 3 guild args as input (its slug parsing is unchanged), but it now dedups via the shared form `clean()` and its reply is crash-proof if a tier is ever blank. Broadening Discord input to accept 1st-only can be a follow-up.

## Tests
Form: 1st-only valid, 1st+2nd valid, requires-1st, no-gap rejected, duplicate-across-two-tiers rejected. Model: `cast_ballot` 1st-only persists None; `__str__` omits blanks. Calculator: skips blank optional tiers. Snapshot: `take()` handles a 1st-only ballot end to end. 117 voting/calculator/snapshot specs + 54 other vote-touching specs green; ruff, mypy, django check, template lint all clean.

VERSION held at 1.21.0 (demo freeze, no Discord announce before Sept 1).


PastLivesReviewBot (approved):
Reviewed and approved by PastLivesReviewBot. Independent adversarial review + delta re-review passed (null-safety sweep clean). Meets PLFOG standards.

PR #263, opened 2026-08-29
feat(calendar): Subscribe to the existing Member and Public Google calendars

HexagonStorms:
Item 11. The Community Calendar's "Export Calendar" dropdown had a "Subscribe via webcal" link pointing at the FOG ICS feed (`hub_calendar_export_ics`), which is `@login_required` — so an external calendar app can't authenticate and the subscription silently fails (hits the login page).

## Approach (per Josh)
Reuse the makerspace's **existing** Member and Public Google Calendars (the ones FOG already syncs from) rather than build a new feed. Every Google Calendar exposes a public iCal URL, so Subscribe just links to those.

## Change
- New helper `hub.calendar_entries.google_calendar_subscribe_url(calendar_id)` builds a `webcal://calendar.google.com/calendar/ical/<url-encoded id>/public/basic.ics` link (empty string when unset).
- `community_calendar` view passes `member_calendar_subscribe_url` / `public_calendar_subscribe_url` from `SiteConfiguration.member/public_google_calendar_id`.
- Dropdown relabeled "Export Calendar" -> "Subscribe" with "Subscribe to the Member calendar" and "Subscribe to the Public calendar" (each shown only when its calendar id is set); the one-time .ics download stays. The broken login-gated webcal link is removed.

## Verified
Both prod calendars' public iCal feeds return HTTP 200 + `BEGIN:VCALENDAR`, so the links work out of the box (no Google sharing change needed). Tests: helper URL-encoding + empty case; view renders both links when configured and hides one when its calendar is unset. 59 specs green (calendar_entries + community_events); ruff, mypy, django check, template lint clean.

### DIFF
--- a/core/help_registry.py
+++ b/core/help_registry.py
@@ -75,8 +75,8 @@ HELP_KEYS: dict[str, HelpKeyEntry] = {
     "calendar.subscribe": {
         "title": "Subscribe to the calendar",
         "short_text": (
-            "Add the community calendar to your own calendar app with the .ics link. "
-            "New events show up there automatically."
+            "Subscribe to the Member or Public calendar and it stays in sync in your own "
+            "calendar app, or download a one-time .ics."
         ),
         "article_slug": "community-calendar",
         "anchor": "calendar-subscribe",
--- a/hub/calendar_entries.py
+++ b/hub/calendar_entries.py
@@ -14,6 +14,7 @@ from __future__ import annotations
 from dataclasses import dataclass
 from datetime import date, datetime, timedelta
 from typing import TYPE_CHECKING, Any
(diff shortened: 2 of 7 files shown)


PastLivesReviewBot (approved):
Reviewed and approved by PastLivesReviewBot. Adversarial review + delta re-review passed: webcal URL construction is correct and injection-safe, single-load wiring, template guards and mutation-resistant tests; help article + hover-help copy updated for the new Subscribe UI (screenshot refresh deferred to the capture pipeline, selector unchanged). VERSION held at 1.21.0 for the pre-demo freeze. Meets PLFOG standards.

PR #260, opened 2026-08-29
feat(spaces): taller map, facility guild links, cubby caption

HexagonStorms:
Spaces map polish (part of items 9 & 11).

## Changes
1. **Taller, responsive map.** `.pl-map-viewport` desktop cap goes from `min(60vh, 480px)` to `min(85vh, 900px)` (nearly double), mobile from `min(60vh, 480px)` to `min(70vh, 560px)` (bumped but still nimble). The JS pan-clamp reads live element sizes so it adapts; breakpoint unchanged.
2. **Facility markers link to their guild.** Data migration `0148` sets `MapHotspot.guild` on facility markers that have an obvious guild home: Wood Shop→Woodworking, Metal Shop→Metalworkers, Stage→Events, Ceramics, Gallery→Visual Arts/Gallery, Leather→Leatherwork, Garden Guild→Gardeners. The space-detail modal already renders a "Go to <guild>" link when a marker has a guild, so this is pure data. It only fills markers with no guild (never clobbers a hand-set link) and is fully reversible.
3. **Cubby caption.** Floors that have cubbies show a caption clarifying the S1-x markers are cubby shelves, gated by a new `Floorplan.has_cubbies` property (counted over already-prefetched hotspots, no extra query).

## Verified in browser (local, migration applied)
Map viewport now 612px at 720p (85vh, caps 900px on larger screens); cubby caption renders on Floor 1 only; clicking **Wood Shop** opens the modal with a **Woodworking Guild** link to `/guilds/woodworking-guild/`. All 7 facility links applied correctly.

## Tests
`has_cubbies` spec (true with a cubby marker, false without); 73 space_map_models specs green; ruff, mypy, django check, `makemigrations --check`, template lint all clean. (Migrations are coverage-omitted.)

VERSION held at 1.21.0 (demo freeze).

https://claude.ai/code/session_01Vfk5tQtVqopZVUqMdKu3GT

PR #265, opened 2026-08-30
feat(spaces): edit a space inline from the detail modal (admin)

HexagonStorms:
Item 9 (the in-modal edit half; the map height + guild links + cubby heading shipped in #260). Admins can now edit a space's status, guild link, label, and description straight from the space-detail popup on /spaces/, behind a pencil Edit button, instead of opening the separate "Edit This Map" view.

## Scope
Editable inline: STATUS (available/occupied/maintenance), GUILD link, LABEL, DESCRIPTION. Deliberately NOT price/size (Airtable-owned, read-only) or kind/shape/space (structural, stay in the full map editor). `SpaceDetailEditForm` is a trimmed ModelForm over exactly those four fields.

## Security / integration
- Both GET (render editor) and POST (save) use the SAME admin gate as the existing map editor (`_require_admin` / `_viewing_as_admin` = `request.view_as.is_admin`), so a logged-in non-admin gets 403 and anon gets the login redirect — identical to the placement editor on the same page. A non-admin can't reach it via view-as (the view-as system won't grant a non-admin an admin effective role). Only edits the marker addressed by the URL pk.
- Status changes reuse the existing `_apply_marker_status` helper, so the Space.status write + Airtable push happen exactly as in the map editor (warns, never 500s, on an Airtable outage). No new Airtable code.
- On save the endpoint re-renders the detail panel plus an `hx-swap-oob` map marker and list row, so map + popup + listings stay in sync.
- Note: an actual admin *previewing as a member* sees no pencil (view-as-aware, matching the sibling "Edit the map" link). Intentional for consistency.

## Tests
`describe_inline_space_editing` + `describe_SpaceDetailEditForm`: non-admin GET/POST both 403 and change nothing; admin opens editor + saves status (Airtable push patched + asserted, Space.status changed, recoloured oob marker, panel shows "Occupied") / guild / label / description; status hidden for space-less markers; unchanged status skips the push; invalid save re-renders with errors and saves nothing; 404 unknown; pencil visible to admin, absent for member/guest. 144 space-map specs + 25 space-request specs green; ruff, mypy (pre-push), django check, template lint, makemigrations --check clean. Form only, no migration.

### DIFF
--- a/hub/forms.py
+++ b/hub/forms.py
@@ -3077,6 +3077,54 @@ class MapHotspotEditForm(MapHotspotForm):
             self.fields["status"].initial = self.instance.space.status
 
 
+class SpaceDetailEditForm(forms.ModelForm):
+    """The in-modal, trimmed editor for one marker, opened from the public Spaces map popup.
+
+    The routine-edit path (item 9): an admin edits exactly the four things members read on the
+    space-detail card — ``status``, the ``guild`` link, the ``label`` and the ``description`` —
+    straight from the popup, without opening the full "Edit This Map" placement editor. It is a
+    strict subset of :class:`MapHotspotEditForm`: ``kind``/``shape``/``space`` stay structural
+    (full editor only) and price/size stay read-only (Airtable owns them), so none of them are
+    fields here. Like the placement editor, ``status`` is not a marker field — it lives on the
+    linked :class:`Space` (Airtable's system of record), so the *view* applies it after ``save()``
+    via ``_apply_marker_status`` and the field is ignored for a marker with no space.
+    """
+
+    status = forms.ChoiceField(
+        required=False,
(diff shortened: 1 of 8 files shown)


PastLivesReviewBot (approved):
Reviewed and approved by PastLivesReviewBot. Adversarial review passed: the inline-edit endpoint mirrors the map editor exactly, admin gate is non-bypassable (effective subset of actual; a non-admin can never reach admin via view-as), target scoping is mass-assignment-safe (only status/guild/label/description), status reuses _apply_marker_status (Airtable push, warns not 500s), OOB swaps are marker-scoped and absent on invalid save, 20 mutation-resistant tests pass. Two non-blocking cosmetic nits only. VERSION held at 1.21.0 for the demo freeze. Meets PLFOG standards.

